PR #23, opened 2023-07-21
[WIP] Mob moss work on packaging flarum

albertchae:
- mob start [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob start [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]
- mob next [ci-skip] [ci skip] [skip ci]


pkgs/flarum/php-packages.nix
```diff
@@ -0,0 +1,1358 @@
...
+        };
+      };
+    };
+  };
+  devPackages = {};
+in
+composerEnv.buildPackage {
+  inherit packages devPackages noDev;
+  name = "flarum-flarum";
+  src = composerEnv.filterSrc ./.;
```
albertchae, line 1351: This line seems to be written assuming that this `nix` expression will be evaluated in the project repo, but because it's in ngipkgs, we only end up with `composer.lock` and the `vendor` directory

pkgs/flarum/default.nix.bak
```diff
@@ -0,0 +1,35 @@
+{ stdenv, fetchFromGitHub, php, phpPackages, ... }:
```
albertchae, line 1: ignore this file for now, this is a mix of a `default.nix` we were handwriting plus trying to copy in some of the autogenerated nix code from `composer2nix`

pkgs/flarum/php-packages.nix
albertchae: We came across https://discourse.nixos.org/t/source-combinators/13331/8 and https://discourse.nixos.org/t/easy-source-filtering-with-file-sets/29117 but neither seem to be usable yet

pkgs/flarum/composer-env.nix
```diff
@@ -0,0 +1,244 @@
+# This file originates from composer2nix
+
+{ stdenv, lib, writeTextFile, fetchurl, php, unzip, phpPackages }:
+
+let
+  inherit (phpPackages) composer;
+
+  filterSrc = src:
+    builtins.filterSource (path: type: type != "directory" || (baseNameOf path != ".git" && baseNameOf path != ".git" && baseNameOf path != ".svn")) src;
```
fricklerhandwerk, line 9: This code is generic, and while usually I'd say that stuff should be part of `composer2nix` and just be `import`ed, you can actually trick it into doing what you want with something along these lines:

```suggestion
    let
      flarum-src = builtins.fetchTarball ...;
      # hardcode flarum source here. not great
      filtered-src = builtins.filterSource (path: type: type != "directory" || (baseNameOf path != ".git" && baseNameOf path != ".git" && baseNameOf path != ".svn")) flarum-src;
    # combine what with what's actually in this source tree
    in pkgs.symlinkJoin [ src filtered-src ];
```

Obviously this is fugly, but it doesn't make any sense to touch the generated sources file. I'm sure you'll find a more reasonable workaround.

No idea if this will actually work with symlinks.

pkgs/flarum/php-packages.nix
fricklerhandwerk: Yeah, this is what I feared. `composer2nix` makes a few assumptions here that we can't work around without ripping open the whole thing. @infinisil is still working on file set combinators, but we may get enough of it into Nixpkgs for you to use.

PR #26, opened 2023-07-27
remove relic comment

matusf:


mightyiam:
Hey, @cleeyv . Could you confirm that this is indeed just a relic?

jfly:
> The core of the workaround he implemented is still present in the flake

For the record, we simply couldn't tell what was the "fugly hack" the comment was referring to. Maybe a picture will make our confusion clearer?

image

That, plus the fact that there's no other mention of `nixos-container` in this file, and this was all introduced in a [WIP commit](https://github.com/ngi-nix/ngipkgs/commit/52c19e19c1b88866027bdee307d9a3c4bf81baff) made us suspect that this was just some comment for some line of code that got removed, but the comment got left behind.

Now I realize [there *is* a mention of nixos-container in the readme](https://github.com/ngi-nix/ngipkgs#add-and-deploy-a-service), so maybe that's relevant? We didn't know about that when we deleted this comment.

cleeyv:
Thank you for this clarification, it all makes more sense now. Yes, the comment was referring to everything below it, which as the time (https://github.com/ngi-nix/ngipkgs/blob/52c19e19c1b88866027bdee307d9a3c4bf81baff/flake.nix), was *only* the attrset merge of the nixosConfigurations block, and it was only necessary to structure it that way as a workaround for the nixos-container problem, hence the comment.

Since then, I have also added hydraJobs and checks to that section, and now you've gone and given it a nice checkOutputs label and it looks much less fugly than it did back at the beginning! This has all happened in a pretty evolutionary way, without so much intentional design guiding it. This makes me think that future refactoring could clean up vestigial structural elements that aren't actually useful anymore, but that doesn't need to happen right away.

Right now I am working on adding more comments on the parts of the flake I understand well, and I welcome you all to do the same.

PR #30, opened 2023-07-31
Add `.mob` to `.gitignore`

lorenzleutgeb:
During Summer of Nix 2023, multiple mobs will work on various packages in this repository. They might all want to configure `mob` differently using a configuration file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 result
 .tmp
+
+# Configuration file for `mob` CLI (see <https://mob.sh>)
+# with contents varying per mob.
+.mob


mightyiam:
I disagree with this in principal.

Also, not all participants use a `.mob` file. In my mob none do. We use the environment variable, instead.

I would have each participant carry the burden of them happening to use some tool by running
```
echo "/.mob" >> .git/info/exclude
```

lorenzleutgeb:
People that do not use a `.mob` file, will likely not care about this additional entry in `.gitignore`. They might only get bitten if they change their mind, create the file and attempt to add it. So there's no additional cost for those that don't use `.mob` (like you).

mightyiam:
I guess I'm just too damn principled.

fricklerhandwerk:
I agree with @mightyiam and don't think it's too principled. One should not include workflow-specific information in version history. This should be part of your local git config.

lorenzleutgeb:
We all.agree that none of us want to check in `.mob`. Whether we use it or not.

Your argument applied to `.tmp`, which is already in `.gitignore`: It should not be there because we do not want anyone creating `.tmp` files leaking into versioned history. This should be ignored individually.

Usually I would just use gitignore.io to generate a list with hundreds of entries for `.gitignore` so it's really odd for me that this is so controversial...

mightyiam:
Purism, @lorenzleutgeb. Purism...

The notion of hundreds of lines in a `.gitignore` file troubles me. It's a different attitude.

I'd do just about anything to refrain from surprises in software.

A file being ignored because it's in some large list counts as a surprise for me.

Issue #3, opened 2023-07-11
Librecast


fricklerhandwerk:
The Librecast project contributes to decentralising the Internet by enabling multicast. It builds transitional protocols and software to extend the reach of multicast and enable easy deployment by software developers.

- [NLnet project](https://nlnet.nl/project/LibreCastLiveStudio/)
- [librecast.net](https://librecast.net/librecast.html)
- source code:
  - https://codeberg.org/librecast
  - https://codeberg.org/librecast/lcrq
  - https://codeberg.org/librecast/librecast
- More information: https://github.com/ngi-nix/ngi/issues/122

There exists a Debian and Guix package. Developers and packagers are on IRC with a Matrix bridge, and also available via email for questions and guidance.

albertchae:
@ngi-nix/moss is claiming this

albertchae:
We have packaged `lcrq`, `librecast` and `lcsync`, which were the 3 targets identified by @cleeyv in https://github.com/ngi-nix/ngi/issues/122#issuecomment-1591866588. We are reaching out to librecast maintainers on https://todo.sr.ht/~librecast to identify there are any other suitable packaging targets.

albertchae:
We are communicating with librecast maintainers on https://todo.sr.ht/~librecast/librecast/89 as well as IRC

albertchae:
We upstreamed `lcrq`, `librecast` and `lcsync` to `nixpkgs` in https://github.com/NixOS/nixpkgs/pull/248437

albertchae:
`lcrq` has been updated and `librecast` 0.7 will be out of RC soon, so we will update in nixpkgs accordingly after https://github.com/ngi-nix/ngipkgs/pull/36#issuecomment-1682025241

cc: @onepict

albertchae:
This was merged a few weeks ago, but we bumped `librecast` to 0.7 in https://github.com/NixOS/nixpkgs/pull/250862

albertchae:
I am closing this issue because I think we accomplished the goals of the issue under the scope of Summer of Nix.

There are 3 packages (`lcrq`, `librecast`, `lcsync`) listed on https://librecast.net/code.html
<img width="839" alt="image" src="https://github.com/ngi-nix/ngipkgs/assets/217050/8257adfe-eb77-4c82-b590-bc7d1c652d84">

And these are the 3 we both packaged into this repo `ngipkgs` and upstreamed into `nixpkgs`.

[...]

PR #35, opened 2023-08-05
librecast: package lcrq

jleightcap:
Towards #3.

Package https://codeberg.org/librecast/lcrq, direct dependency of https://codeberg.org/librecast/librecast.


albertchae:
We referenced the Guix packaging here https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/networking.scm?h=v1.4.0#n250

Issue #40, opened 2023-08-23
Repackage liberaforms without mach-nix


cleeyv:
Issue #39 identifies how the use of mach-nix for liberaforms (which uses an IDF) depended on an old version of nixpkgs as a workaround to a bug a with mach-nix https://github.com/DavHau/mach-nix/issues/549 which made it incompatible with newer versions of nixpkgs. That is documented here: https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/liberaforms/env.nix#L9-L16 it was from back in June when we set up ngipkgs with liberaforms as the initial test service.

Since then, I see that in early July a new PR was merged for mach-nix https://github.com/DavHau/mach-nix/pull/563 that apparently fixes the incompatibility with newer nixpkgs. Assuming this fix works, then in theory this issue could be resolved by sourcing the import of mach-nix for liberaforms from its master branch (to include the fix) and then removing the old nixpkgs workaround. I tried this here: https://github.com/ngi-nix/ngipkgs/tree/remove-old-nixpkgs but ran into the following error (which was also problematic in earlier iterations of packaging liberaforms):
```
Some requirements could not be resolved.
Top level requirements: 
  alembic==1.7.5  attrs==21.4.0  Babel==2.9.1  beautifulsoup4==4.10.0  bleach==4.1.0  cachelib==0.5.0  certifi==2021.10.8  cffi==1.15.0  charset-normalizer==2.0.9  click==8.0.3  cryptography==36.0.0  dnspython==2.1.0  email-validator==1.1.3  feedgen==0.9.0  F>
Providers:
  {'_default': ['wheel', 'sdist', 'nixpkgs'],
   'gdal': ['nixpkgs'],
   'pip': ['nixpkgs', 'sdist'],
   'setupmeta': ['wheel'],
   'setuptools': ['nixpkgs'],
   'wheel': ['nixpkgs', 'sdist']}
Mach-nix version: master
Python: 3.10.12
Cause: Requirements conflict: cryptography (<SpecifierSet('==36.0.0')>,)
The requirements which caused the error:
  cryptography (<SpecifierSet('==36.0.0')>,)

The given requirements might contain package versions which are not yet part of the dependency DB
currently used. The DB can be updated by specifying 'pypiDataRev' when importing mach-nix.
For examples see: https://github.com/DavHau/mach-nix/blob/master/examples.md
```
[...]

Issue #39, opened 2023-08-23
`nix flake check` with `no-url-literals` fails


lorenzleutgeb:
## Background

URL literals are deprecated:

 * https://github.com/NixOS/rfcs/pull/45
 * [Nix Manual](https://nixos.org/manual/nix/stable/contributing/experimental-features.html?highlight=deprecated#no-url-literals)

There was at least [one treewide commit in nixpkgs](https://github.com/NixOS/nixpkgs/commit/84cf00f98031e93f389f1eb93c4a7374a33cc0a9) removing URL literals three years ago, which [rewrote a URL literal in the definition of the `prettytable` (Python) package](https://github.com/NixOS/nixpkgs/blame/b469261cfc790bd92aae22e02e83db15102754b2/pkgs/development/python-modules/prettytable/default.nix#L39).

I think that ngipkgs should evaluate with `no-url-literals`.

## Problem

Today, I tried:

```shell
% git name-rev HEAD
HEAD main
% git rev-parse HEAD
ab5adeaed3c36a91236f99b7ccd41dfaf1cf48ea
% nix flake --allow-import-from-derivation --extra-experimental-features 'no-url-literals' check
error:
       … while checking flake output 'hydraJobs'

         at /nix/store/yfs4vb1xvhx3qlclr36z4mqi1r1qdhvp-source/flake.nix:88:9:

           87|         # See <https://hydra.ngi0.nixos.org/jobset/ngipkgs/main>.
           88|         hydraJobs.packages.${linuxSystem} = self.packages.${linuxSystem};
             |         ^
           89|

       … while checking the Hydra jobset 'hydraJobs'

         at /nix/store/yfs4vb1xvhx3qlclr36z4mqi1r1qdhvp-source/flake.nix:88:9:

           87|         # See <https://hydra.ngi0.nixos.org/jobset/ngipkgs/main>.
           88|         hydraJobs.packages.${linuxSystem} = self.packages.${linuxSystem};
             |         ^
           89|

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: URL literals are disabled

       at /nix/store/87jb2lym32p02bg284nhfxqw3vk37lwi-source/pkgs/development/python-modules/prettytable/default.nix:24:16:

           23|     description = "Simple Python library for easily displaying tabular data in a visually appealing ASCII table format";
           24|     homepage = http://code.google.com/p/prettytable/;
             |                ^
           25|     license = licenses.bsd3;
```

[...]

Issue #24, opened 2023-07-26
CI builds for darwin


cleeyv:
Last week I finished initial implementation of CI evaluations and builds of ngipkgs on each push and PR using github actions: #18. However, it was implemented to only build on linux (all the monorepo dev work thus far had been done on NixOS). After sharing an update on my progress, I received the following feedback from @fricklerhandwerk:
> Albert's mob asked about darwin builds in the CI
> Is there a particular reason you only have x86_64-linux?
> Apparently most of them are on darwin

I replied at the time with my partial understanding at the time of the problems with darwin builds of ngipkgs. Since then I have worked on it from various angles and have a much clearer understanding of the problems involved and potential solutions. 

### Background Info

One foundational piece of nix knowledge that is at the core of understanding the problem is the difference between [evaluation](https://nix.dev/tutorials/first-steps/nix-language#how-to-run-the-examples) (which generates a [derivation](https://nix.dev/tutorials/first-steps/nix-language#derivations)) and [realisation](https://zero-to-nix.com/concepts/realisation) (aka building), which is the carrying out of the build instructions in the derivation to produce an output in the nix store.

When using flakes, every derivation (package, test, dev shell, etc) needs to have a [system attribute specified](https://zero-to-nix.com/concepts/system-specificity) (defining the CPU architecture and the OS, such as `x86_64-linux` or `aarch64-darwin`) and all the dependencies of the derivation need to be available for the matching system. A derivation will have an evaluation-time error if one of its dependencies isn't available for its system. 

Since there are so many potential systems that nix can run on, for multi-platform software manually defining system attributes in a flake would require a lot of boilerplate code. Hence the widespread use, including in the ngipkgs flake, of [flake-utils](https://github.com/numtide/flake-utils/) which allows for assigning system attributes for a list of `defaultSystems`, which includes the following: `["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"]`.

[...]

PR #41, opened 2023-08-24
treewide: Remove IFDs

lorenzleutgeb:
Relates to:
 - https://github.com/ngi-nix/ngipkgs/issues/24
 - https://github.com/ngi-nix/ngipkgs/issues/39
 - https://github.com/ngi-nix/ngipkgs/issues/40


pkgs/kikit/solidpython/default.nix
```diff
@@ -62,13 +41,18 @@ in
       hash = "sha256-3fJta2a5c8hV9FPwKn5pj01aBtsCGSRCz3vvxR/5n0Q=";
     };
 
+    patches = [
+      # Unpin PrettyTable, see <https://github.com/SolidCode/SolidPython/issues/207>.
+      ./prettytable.patch
+    ];
```
jfly, line 47: We learned recently that nixpkgs actually has a more idiomatic way of relaxing dependencies. See https://nixos.org/manual/nixpkgs/stable/#using-pythonrelaxdepshook, and us using it here: https://github.com/NixOS/nixpkgs/pull/249464/commits/38a7f42312ad59e7604dc1ed01ebc1acaf59849c.

jfly (approved):
The kikit stuff LGTU (@ngi-nix/fungus)! Thanks for dealing with it. We actually had to remove IFD when we submitted this stuff to try to upstream it to nixpkgs in https://github.com/NixOS/nixpkgs/pull/249464/, sorry for the duplicated work!

pkgs/kikit/solidpython/default.nix
lorenzleutgeb: Nice, thanks! This might come in handy also for packaging pretalx. I'll update this PR ASAP.

lorenzleutgeb: Fixed in 35b6a22.

cleeyv (commented):
This generally looks good to me! The only thing I see missing is to replace the use of liberaforms as an example in the README for adding and deploying a service: https://github.com/ngi-nix/ngipkgs/blob/main/README.md?plain=1#L43-L49

I think pretalx would be an excellent alternative to liberaforms in this section. ~~It doesn't need to be exactly the same things covered, you can just copy what you've already put in your README for pretalx, so it could be changed to "Add and test a service" and we can include more about deployment later.~~

I'm working now on making this edit to the README using the pretalx example, hopefully a commit coming soon.

cleeyv (approved):

mightyiam:
Nooice

PR #50, opened 2023-09-15
flake: Add GitHub Action to update flake.lock

lorenzleutgeb:


cleeyv:
Sorry for the delayed reply on this. Looks good! I haven't used this action before so I'm curious to see how it is going to work in practice. Based on how it goes we can always make adjustments later if needed. 

The main detail I anticipate needing to adjust is the frequency of the update PRs. Every three days seems alright during summer of nix while there are many people working on the repo regularly but that could be greatly reduced for the rest of the year. 

I'll go ahead and merge this now!

PR #57, opened 2023-09-21
pretalx: fix plugins and enable tests for pretalx

andres-nav:
Fixed plugins: pretalx-venueless and pretalx-public-voting


augustebaum:
Addresses #2

PR #60, opened 2023-09-25
repo: Introduce `pkgs/by-name`

### LINKED ISSUE #51: `pkgs/by-name`
Inspired by https://github.com/NixOS/rfcs/pull/140 I'd like to propose: In order to minimize the size of `/all-packages.nix` let's agree on a convention for registering packages simply by their location in the tree.

My suggestion would be to use `/pkgs/by-name/${pname}/package.nix`. As of today, the following packages could be moved:

 * euclid3
 * flarum
 * gnunet-messenger-cli
 * kikit
 * lcrq
 * lcsync
 * libgnunetchat
 * librecast
 * pcbnew-transition
 * pybars3
 * pymeta3
 * rosenpass
 * rosenpass-tools

I expect that most packages in this repo would fit `pkgs/by-name`. Further, I do not expect that shards (like in nixpkgs) will be required anytime soon. If that point is ever reached, they can easily be introduced automatically.

I expect the glue code required to produce a package from such a structure to be very similar to https://github.com/tweag/nixpkgs/blob/77d50b03e4388f22e1f36a2621a9287a12a138be/pkgs/top-level/by-name-overlay.nix and I'd be happy to implement it.

Such a convention might speed up things like https://github.com/ngi-nix/ngipkgs/issues/12 slightly.

lorenzleutgeb:
Resolves #51


fricklerhandwerk (approved):

alejandrosame (approved):

alejandrosame:
Merging, it looks good and usable as is!

Issue #66, opened 2023-10-05
Migrate `arpa2`


alejandrosame:
https://github.com/ngi-nix/arpa2

alejandrosame:
@ngi-nix/fungus is working on this

mightyiam:
Me and @augustebaum are working on this. But we can use some help. There's some dependency resolution problem. An ancient library called comerr or similar. If anyone would be kind enough to join us for a session and help out with this problem, that would be fantastic.

@fricklerhandwerk @lorenzleutgeb

mightyiam:
That problem has been solved by now. Thanks to @tfc who really seems to know what he's doing.

PR #79, opened 2023-10-25
arpa2 

mightyiam:
- Add arpa2cm
- Add arpa2common
- Add steamworks
- Add steamworks-pulleyback
- Update flake
- Propagate arpa2cm from arpa2commmon
- Add quick-mem
- Update flake
- Add arpa2cm back to quick-mem
- Add lillydap
- Add quick-der
- Update flake
- Add leaf
- Update flake
- Add quick-sasl
- Add tlspool
- Update flake
- Add tlspool-gui
- Update flake
- Add hydra job to the flake
- Add all the packages to the flake's check
- Make quick-sasl build
- Add freeDiameter derivation
- Add kip derivation
- Update flake
- Update kip dependencies
- Add flake-compat to the inputs
- Add flake-compat
- Put steamworks-pullyback back into the overlay
- Update phases to use cmake
- Add aarch64-linux to the supported systems
- Nevermind, hydra doesn't like aarch64-linux
- Change python37 to python3
- Refactor repo
- Add steamworks-pulleyback to the checks
- Add cmake flag for extension install dir
- Add kip to the checks
- Change build directory name
- Format flake.nix with nixpkgs-fmt
- Make naming more consistant
- Move the outputs attribute set to the top
- Create helper function `mkArpa2Derivation`
- Make some packages use the new `mkApra2Derivation`
- Set `freeDiamater` version to `1.5.0`
- Convert all `callPackage` calls to new convection
- Fix `steamworks-pulleyback` `CMAKE_PREFIX_PATH`
- Convert more packages to `mkArpa2Derivation`
- Simplify the `mkArpa2Derivation` a bit
- Remove unnecessary `configurePhase`
- Add documentation
- Add `meta` attributes
- Add `README.md`
- Set `cmakeBuildType` to `Debug`
- Fix typo
- Remove unnessecary `postUnpack`
- Add `lib` as input
- arpa2: fmt
- arpa2: pkgconfig -> pkg-config
- arpa2: quick-der -> quickder
- arpa2: quick-mem -> quickmem
- arpa2: quick-sasl -> quicksasl
- arpa2: use root nixpkgs and rm some packages
- arpa2: non-flake inputs -> fixed-output drvs
- arpa2: pull package sources into package files
- arpa2: remove empty arpa2-python-packages overlay
- arpa2: remove unused poetry2nix input and overlay
- arpa2: inline the mkArpa2Derivation helper
- flake: dissociates args of explicit & by-name pkgs
- arpa2: migrate into root flake


flake.nix
```diff
@@ -44,15 +44,20 @@
...
-        args = {
+        pkgsByName = import ./pkgs/by-name {
           inherit (pkgs) lib;
           inherit callPackage;
         };
-        result = (import ./pkgs/by-name args) // (import ./pkgs args);
+        explicitPkgs = import ./pkgs {
+          inherit (pkgs) lib;
+          inherit callPackage;
+          libsForQt5 = pkgs.lib.makeScope pkgs.libsForQt5.newScope (final: allPackages);
```
lorenzleutgeb, line 56: What's the motivation for adding this scope? Just because it was used in the original arpa2 repository?

mightyiam: This seemed straightforward to us. Also, we're not certain we know of an alternative.

lorenzleutgeb: The alternative is to not introduce a new scope, and put it in the same scope as all other packages so far. Does this break anything? What is a good argument for introducing the scope versus just using the scope we already have?

Even if a new scope was introduced, I'd prefer a solution that does not involve listing the scopes in `flake.nix`, but rather `pkgs/default.nix` or similar. Would that be possible?

augustebaum: If we wanted to create the scope in `pkgs/default.nix` we would have to pass `pkgs` into it. We noticed that `pkgs` is currently not passed there; we're not sure why, so we arbitrarily decided to only pass the new `libsForQt5`. Would you have us pass `pkgs` into `pkgs/default.nix` and create the scope there?

augustebaum:
Closes #66

flake.nix
lorenzleutgeb: Yes.

mightyiam: The reasoning behind the new scope is as follows. tlspool-gui depends on tlspool. Also, it seems that typically in nixpkgs, qt apps are called from that scope. So creating a new scope from that scope and with the addition of our flake's packages would provide that package with an idiomatic signature and the dependency it requires.

Having said that, we changed our minds and force-pushed that commit. There's no new scope now.

PR #88, opened 2023-11-01
flake: do not check packages marked as broken

mightyiam:
We possibly plan to actually introduce a package marked broken.
So this is a precursor change.

We manually tested this by marking a single package with `meta.broken = true` and indeed `nix flake show` listed it under packages yet not under hydra jobs or checks.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -100,6 +100,9 @@
         linuxSystem = "x86_64-linux";
         pkgs = importNixpkgs linuxSystem [self.overlays.default];
         treefmtEval = loadTreefmt pkgs;
+        nonBrokenPkgs =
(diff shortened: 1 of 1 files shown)


lorenzleutgeb:
Yeah, I had this in mind, too, see this commit from 2023-08-24 which I never PRed: https://github.com/ngi-nix/ngipkgs/commit/a0d4d942a95186688ec70f909c0cc31884d65ce6

Do we agree that a new issue should be created whenever a package is marked broken to discuss how to handle the breakage?

mightyiam:
> Do we agree that a new issue should be created whenever a package is marked broken to discuss how to handle the breakage?

Sounds right.

PR #91, opened 2023-11-14
flake: dissociates args of explicit & by-name pkgs

matusf:


mightyiam (approved):
`pkgsByName` and `explicitPkgs` only _happen_ to take the same args currently. By definition, `pkgsByName` will never need any other args because they are always merely `callPackage`d. Whereas `explicitPkgs` may take additional args as needed. And, regardless, this is a little extraction worth making.

lorenzleutgeb (approved):

PR #95, opened 2023-11-20
add MIT license

fricklerhandwerk:
another option would have been MPL 2.0, which is generally preferable in terms of the NGI program: to ensure that public money keeps translating into public code.

the decision for MIT is purely practical: to ease migrating the code into Nixpkgs, the place where the software project supported through NGI will most likely be found and put to use in a broader context.

there was no feedback from Nixpkgs maintainers on the topic: https://discourse.nixos.org/t/per-file-license-notes-in-nixpkgs/32390

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 ngi-nix
+Copyright (c) 2023 NGIpkgs contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 ngi-nix
```
lorenzleutgeb, line 3:
```suggestion
Copyright (c) 2023 the contributors to NGIpkgs
```

I'd prefer something that's a little more specific than just 'ngi-nix'. Even though I understand we have to be sufficiently vague.

lorenzleutgeb (changes requested):
Choice of license LGTM, I just want to encourage giving another thought to how the copyright holders should be described.

LICENSE
fricklerhandwerk: I just dropped the GitHub template, thanks for the fixup!

```suggestion
Copyright (c) 2023 NGIpkgs contributors
```

lorenzleutgeb (approved):

RonnyLam:
TBH, I am not in favor of the MIT license. While it is the most easy one, it is also the most permissive one. However, since nixpkgs also has the MIT licence, I see no problem in this one. The packages themselves still have their own respected license.

RonnyLam (approved):

PR #100, opened 2023-11-28
flake: nixosTests refactor

mightyiam:
Does it make it more readable? Not sure. Maybe. How do you feel about it?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -34,13 +34,15 @@
         in
           mapAttrs (name: _: let
             mkTestModule = import "${dir}/${name}";
-          in
-            pkgs.nixosTest (mkTestModule {
+
+            testModule = mkTestModule {
               inherit pkgs;
               inherit (pkgs) lib;
               modules = extendedModules;
               configurations = importNixosConfigurations;
-            })) (readDir dir);
+            };
+          in
+            pkgs.nixosTest testModule) (readDir dir);
         callPackage = pkgs.newScope (
           allPackages // {inherit callPackage nixosTests;}
         );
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -29,21 +29,25 @@
...
+                mkNetwork = import "${dir}/${name}";
+                network = mkNetwork {
                   inherit pkgs;
                   inherit (pkgs) lib;
                   modules = extendedModules;
                   configurations = importNixosConfigurations;
-                })) (readDir dir)
+                };
+              in
+                pkgs.nixosTest network;
```
lorenzleutgeb, line 48: I am not sure that the argument passed to `nixosTest` should be called a "network". It does contain `nodes` that will end up networked, but it also contains the test script itself, and some other metadata (test name) and configuration (for interactive testing). The [NixOS manual says]: "We refer to the whole test above as a test module, whereas the values in [`nodes.<name>`](https://nixos.org/manual/nixos/stable/#test-opt-nodes) are NixOS modules themselves."

So, I'd go for `testModule` instead.
```suggestion
                pkgs.nixosTest testModule;
```

lorenzleutgeb (changes requested):

lorenzleutgeb:
Does make it more readable. However, not sure about the name "network".

mightyiam:
The name network I picked up from the nixpkgs manual. I found this function and the parameter is named network. Sorry, I'm on phone. @lorenzleutgeb .

flake.nix
mightyiam: I agree.

PR #101, opened 2023-11-28
flake: nixpkgs.nixosTests not available to local pkgs

mightyiam:


mightyiam:
@lorenzleutgeb I'm having a challenge wrapping my mind around this. What did I just remove here, and why does it seem to `nix flake check` okay without it?

lorenzleutgeb:
I think the difference is (abstract):

```diff
- importPackages = pkgs: let nixosTests = pkgs.nixosTests // ( /* ... */ );
+ importPackages = pkgs: let nixosTests =                    ( /* ... */ );
```

That is, you removed the application of `//` to merge `pkgs.nixosTests` with the expression that you kept.

I think that `pkgs.nixosTests` should be available, because we will want to re-export packages that were merged into Nixpkgs, like Rosenpass, which have tests in Nixpkgs, and we might want to reference them. I am not entirely sure about this. However, if you think that just having `pkgs.nixosTests` in there because we *might* need it, is bad, then we can also just remove it.

I thought that it might be required to access files that are common to multiple tests, but it isn't, since we would import those files, not use `pkgs.nixosTests` to access their contents.

mightyiam:
Okay. Thank you for the explanation, @lorenzleutgeb. For the sake of simplifying our flake, I'll merge this.

PR #105, opened 2023-11-29
flake: linuxSystem -> system

mightyiam:


mightyiam:
In this context, this is a simpler name.

PR #111, opened 2023-11-29
flake: deep merge the outputs attrsets

mightyiam:


mightyiam:
This isn't currently necessary, because we have no merge overrides, but it may be and it's no harm.

PR #114, opened 2023-11-29
ci: nix flake show comment in prs

mightyiam:


.github/workflows/pull_request.yaml
```diff
@@ -15,4 +15,17 @@ jobs:
       - uses: DeterminateSystems/magic-nix-cache-action@main
         with:
           upstream-cache: https://ngi.cachix.org/
+      - id: flake-show
+        name: Obtain `nix flake show` output
+        run: |
+          sudo apt install colorized-logs
+          echo 'output<<EOF' >> $GITHUB_OUTPUT
+          nix flake show --all-systems | ansi2txt >> $GITHUB_OUTPUT
+          echo 'EOF' >> $GITHUB_OUTPUT
```
lorenzleutgeb, line 24:
```suggestion
          echo 'output <<EOF' >> $GITHUB_OUTPUT
          nix flake show --all-systems | nix run nixpkgs#ansifilter >> $GITHUB_OUTPUT
          echo 'EOF' >> $GITHUB_OUTPUT
```

Not sure whether the reference to Nixpkgs will work like that...

lorenzleutgeb (changes requested):
I'd like to avoid installing packages via `apt`, because we're into Nix.

lorenzleutgeb:
What'd be even cooler, IMO, would be to show the differences.

What works quite nicely on my machine:
```shell
$ export REF=refs/heads/main
$ delta \
  <(nix flake show ".?ref=$REF&rev=$(git rev-parse $REF)" | nix run nixpkgs#ansifilter) \
  <(nix flake show | nix run nixpkgs#ansifilter)
```

So, for GitHub Actions we'd probably need:

```shell
diff -u \
  <(nix flake show ".?ref=$GITHUB_REF&rev=$(git rev-parse $GITHUB_REF)" | nix run nixpkgs#ansifilter) \
  <(nix flake show | nix run nixpkgs#ansifilter)
```

And then use

``````
```patch
${{ steps.flake-show.outputs.output }}
```
``````

mightyiam:
That would be nice. Perhaps a full-context diff?

lorenzleutgeb (approved):

mightyiam:
Thank you, @lorenzleutgeb .

PR #115, opened 2023-11-29
ci: Show diff of `nix flake show`

lorenzleutgeb:


mightyiam:
I love the result, but the implementation, for me, it seems to have gotten a bit "out of hand". This much bash... would you mind rewriting this in Rust, perhaps?

lorenzleutgeb:
Note that there are two automated comments on this PR. [The first one](https://github.com/ngi-nix/ngipkgs/pull/115#issuecomment-1832305919) shows how removal of a package would look like. [The second one](https://github.com/ngi-nix/ngipkgs/pull/115#issuecomment-1838966975) shows how a PR that does not change `nix flake show` would look like.

PR #122, opened 2023-11-30
style: consistent bindings into builtins and lib

mightyiam:
based on #121


mightyiam:
So, I went on a bit of a rampage.

mightyiam:
Thank you, @lorenzleutgeb

PR #133, opened 2023-12-20
bump nixpkgs

alejandrosame:
I needed to bump nixpkgs as a prerequisite of #99.

However, updating the input introduced some regressions, so I decided to bump and fix on a separate PR.

PR #134, opened 2023-12-21
ci: Don't comment from forks

lorenzleutgeb:
See #133

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -35,6 +35,7 @@ jobs:
             echo -e '```\n</details>\nEOF' >> "$GITHUB_OUTPUT"
           fi
       - uses: thollander/actions-comment-pull-request@v2
+        if: ${{ !github.event.pull_request.head.repo.fork }}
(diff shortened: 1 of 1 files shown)


alejandrosame (approved):
Seems simple enough. However I wonder how other automations like stalebot do it then.

fricklerhandwerk:
Wouldn't that mean that PRs from forks won't get CI checks any more? @lorenzleutgeb the [documentation you linked](https://github.com/thollander/actions-comment-pull-request#permissions) in the [other thread](https://github.com/ngi-nix/ngipkgs/pull/133#issuecomment-1866143754) says one could instead trigger on `pull_request_target` events, which "will execute the action from the target branch and not from the origin PR". What does that even mean? The [GitHub documentation about it](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) seems not helpful.

lorenzleutgeb:
Closing in favor of #176

PR #138, opened 2023-12-22
twister: init at 0.9.3

matusf:


pkgs/by-name/twister/package.nix
```diff
@@ -0,0 +1,48 @@
+{
+  autoreconfHook,
+  curl,
+  fetchgit,
+  gnunet,
+  jansson,
+  lib,
+  libgcrypt,
+  libmicrohttpd,
+  libnfc,
```
alejandrosame, line 10: This looks like an unused input.

Other than that, LGTM.

alejandrosame (approved):

pkgs/by-name/twister/package.nix
matusf: Thanks! Fixed.

PR #156, opened 2024-01-29
flarum: nixos test imports dummy.nix

### LINKED ISSUE #125: warning: system.stateVersion is not set
> warning: system.stateVersion is not set, defaulting to 24.05. Read why this matters on https://nixos.org/manual/nixos/stable/options.html#opt-system.stateVersion.

mightyiam:
closes #125

### DIFF
--- a/configs/all-configurations.nix
+++ b/configs/all-configurations.nix
@@ -3,7 +3,10 @@
   # Refer to <https://github.com/ngi-nix/ngipkgs/issues/40>.
   #liberaforms-container = import ./liberaforms/container.nix;
   flarum = {
-    imports = [./flarum];
+    imports = [
+      ./flarum
+      ./dummy.nix
+    ];
   };
   pretalx-postgresql = {
     imports = [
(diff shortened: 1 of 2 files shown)


configs/all-configurations.nix
```diff
@@ -3,7 +3,10 @@
   # Refer to <https://github.com/ngi-nix/ngipkgs/issues/40>.
   #liberaforms-container = import ./liberaforms/container.nix;
   flarum = {
-    imports = [./flarum];
+    imports = [
+      ./flarum
+      ./dummy.nix
```
fricklerhandwerk, line 8: It's not evident without context why that is necessary. But supposedly this could be done in the outer layer? `dummy.nix` doesn't seem like a helpful name, and maybe if it's really small that config addition can be declared inline at the point where it's applied to each config."

mightyiam: How about now, @fricklerhandwerk ?

fricklerhandwerk: Makes more sense to me!

fricklerhandwerk (approved):

Issue #15, opened 2023-07-13
Atomic Data


fricklerhandwerk:
Atomic Data is a modular specification for sharing, modifying and modeling graph data. It combines the ease of use of JSON, the connectivity of RDF (linked data) and the reliability of type-safety.

- [NLnet project](https://nlnet.nl/project/AtomicTables/)
    - source code:
        - https://github.com/atomicdata-dev/atomic-server
        - https://github.com/atomicdata-dev/atomic-data-browser
    
This project is actively developed, and the lead is interested in collaboration. We will figure out the best way to approach this, and it may involve some sort of onboarding if an upstream Nix setup is desired.

joepio:
Hi! Awesome that you want to help out :)

I'll first make sure both Atomic-Server and Atomic-Data-Browser codebases are combined into one monorepo.

fricklerhandwerk:
We can start with packaging the Rust backend, and then continue with the Typescript frontend when it's merged. Then testing both components in integration as well as building a module to run the service on NixOS would be possible.

joepio:
It's merged! It's now one monorepo. Check the `dockerfile` for how it's currently built. Let me know if I can help in any way! 

Thank you :)

albertchae:
@ngi-nix/moss is claiming this. We will ask you about any questions we have about the project as we learn more about it @joepio !

albertchae:
Brainstorming a plan of action (not necessarily in order):
- get a running instance of atomic data locally outside of nix – this was very helpful for debugging flarum issues
- potentially packaging pnpm with nix is difficult, @Chickensoupwithrice had some experience with this
- look into packaging rust separately and first (as suggested in https://github.com/ngi-nix/ngipkgs/issues/15#issuecomment-1638377333), hopefully it is decoupled enough 
  - Looking through the repo, we think packaging https://github.com/atomicdata-dev/atomic-server/tree/develop/lib / https://crates.io/crates/atomic_lib is the first step

albertchae:
Summarizing where we are right now:
https://github.com/atomicdata-dev/atomic-server has 5 components
- `lib` is a Rust library so will be compiled from source by components that need it, so it doesn't need to be packaged separately
- `server` and `cli` have been packaged in https://github.com/ngi-nix/ngipkgs/pull/62
- We are currently working on `browser` (and `desktop` depends on this via tauri).

[...]

PR #164, opened 2024-03-06
init: atomic-browser at 0.37.0

albertchae:
The next step of https://github.com/ngi-nix/ngipkgs/issues/15

- Use pnpm2nix-nzbr which supports pnpm version 6 lockfiles.
- Because atomic-browser is using pnpm workspaces, `installInPlace` and `distDir="."`

A follow up PR will create a module for this

Co-authored-by: Anish Lakhwara <anish+git@lakhwara.com>

### DIFF
--- a/pkgs/by-name/atomic-browser/package.nix
+++ b/pkgs/by-name/atomic-browser/package.nix
@@ -1,0 +1,36 @@
+{
+  fetchFromGitHub,
+  lib,
+  mkPnpmPackage,
+}: let
+  inherit
+    (lib)
+    licenses
+    maintainers
+    ;
+in
+mkPnpmPackage rec {
+  pname = "atomic-browser";
+  version = "v0.37.0";
+
+  monorepoSrc = fetchFromGitHub {
+    owner = "atomicdata-dev";
+    repo = "atomic-server";
+    rev = "v0.37.0";
+    hash = "sha256-+Lk2MvkTj+B+G6cNbWAbPrN5ECiyMJ4HSiiLzBLd74g=";
+  };
+
+  src = "${monorepoSrc}/browser";
+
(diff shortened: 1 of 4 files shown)


pkgs/by-name/atomic-browser/package.nix
albertchae, line 1: Tried to make this file follow the format of all the others since it appears they were standardized at some point

pkgs/by-name/atomic-browser/package.nix
```diff
@@ -0,0 +1,36 @@
...
+mkPnpmPackage rec {
+  pname = "atomic-browser";
+  version = "v0.37.0";
+
+  monorepoSrc = fetchFromGitHub {
+    owner = "atomicdata-dev";
+    repo = "atomic-server";
+    rev = "v0.37.0";
+    hash = "sha256-+Lk2MvkTj+B+G6cNbWAbPrN5ECiyMJ4HSiiLzBLd74g=";
+  };
```
albertchae, line 21: there's one other package with a monorepo here, should we follow the same pattern @Chickensoupwithrice 

https://github.com/ngi-nix/ngipkgs/blob/2aaa8b564709560bcb2add226627dd7e49562b31/pkgs/by-name/mcaptcha/package.nix#L12-L17

Chickensoupwithrice: Seems good enough to follow the convention

Chickensoupwithrice (approved):
Build works for me too.

albertchae:
This is failing CI because the builder we introduced (https://github.com/nzbr/pnpm2nix-nzbr/) is using IFD. We are exploring alternatives

fricklerhandwerk:
Discussed with @albertchae and @Chickensoupwithrice

Closing in favor of https://github.com/ngi-nix/ngipkgs/pull/170

PR #169, opened 2024-04-10
Static Site

### LINKED ISSUE #165: Static site that lists all available projects
Render a static site that lists all NLnet projects that are somehow accessible through this repo, including packages (derivations), the executables they produce, NixOS modules, sample configurations, and some metadata such as the project website.

The general approach we had discussed with @lorenzleutgeb would be
1. Equip derivations and modules with a metadata field that corresponds to the project key on [nlnet.nl/project](https://nlnet.nl/project/). Related: https://github.com/ngi-nix/ngipkgs/issues/74
2. Write a transformation in the Nix language that will convert all information in the monorepo to a suitable data structure, and then to an HTML document.
3. That HTML document can be checked into a branch and served via GitHub pages as a first step.

Anything fancy, such as skins, dynamic search, etc. can be added later. What's important is that there is a human-readable overview of what exists.

lorenzleutgeb:
Fixes #165

Comments:
 - Code to generate a (quite large) list/dump of options is in `flake.nix`, all details regarding the static site is in `overview/*`.
 - Since there's currently no straightforward way to query the module system about which options are defined in a given module, I had to add `nixos.modules.<name>.options` to the schema for `projects/*/default.nix` which is a regex that describes the names of the options that the module at `nixos.modules.<name>.path` defines. I don't particularly like this, but it's the best solution I can think of at the moment. I did briefly consult with the module system maintainers on Matrix.
 - Uses Pandoc to generate the HTML.
 - Uses the v.Nu HTML checker to ensure that our generated HTML is sane.
 - Also pipes through configurations with syntax highlighting (powered by Pandoc or whatever it depends on for syntax highlighting).

### DIFF
--- a/projects/Kbin/default.nix
+++ b/projects/Kbin/default.nix
@@ -5,7 +5,10 @@
 } @ args: {
   packages = {inherit (pkgs) kbin kbin-frontend kbin-backend;};
   nixos = {
-    modules.service = ./service.nix;
+    modules.service = {
+      path = ./service.nix;
+      options = "services\\.kbin\\..*";
+    };
     configurations = {
       base = {
         path = ./configuration.nix;
--- a/projects/default.nix
+++ b/projects/default.nix
@@ -40,6 +40,6 @@ in
       };
     in
       if isMarkedBroken project
-      then trace "Project '${name}' marked as broken (for system '${pkgs.system or "undefined"}'). Skipping." {}
+      then trace "Skipping project '${name}' which is marked as broken for system '${pkgs.system or "undefined"}'." {}
       else hydrate project
   ) (filterAttrs filter (readDir baseDirectory))
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -1,0 +1,39 @@
+name: overview
+
+on:
+  push:
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@main
+        with:
+          extra-conf: |
+            experimental-features = no-url-literals
+      - uses: cachix/cachix-action@v10
+        with:
+          name: ngi
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+      - run: nix ${{ runner.debug && '--debug --print-build-logs' }} build .#overview
+      - name: Upload Pages artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: result
+
+  deploy:
+    runs-on: ubuntu-22.04
+    needs: build
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
+
--- a/flake.nix
+++ b/flake.nix
@@ -61,7 +61,15 @@
       import ./projects {inherit lib pkgs sources;};
 
     # Functions to ease access of imported projects, by "picking" certain paths.
-    pick = {
+    pick = rec {
+      packages = mapAttrByPath ["packages"] {};
+      modulePaths = x:
+        mapAttrs (_: v:
+          concatMapAttrs (n: v:
+            if v ? path
+            then {${n} = v.path;}
+            else {})
+          v) (modules x);
       modules = mapAttrByPath ["nixos" "modules"] {};
       tests = mapAttrByPath ["nixos" "tests"] {};
       configurations = x: mapAttrs (_: v: mapAttrs (_: v: v.path) v) (mapAttrByPath ["nixos" "configurations"] {} x);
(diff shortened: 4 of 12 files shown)


projects/Flarum/project.nix
```diff
@@ -0,0 +1,7 @@
+{pkgs, ...}: {
+  packages =
+    if pkgs == {}
+    then builtins.throw "wut?"
+    else {inherit (pkgs) flarum;};
```
fricklerhandwerk, line 5:
```suggestion
  packages = {inherit (pkgs) flarum;};
```

Probably no need to throw?

Also that could as well be `default.nix` because it's the entry point to this piece of code.

projects/default.nix
```diff
@@ -0,0 +1,42 @@
...
+
+  projectDirectories = lib.filterAttrs (_: directory:
+    if pathExists (directory + "/project.nix")
+    then true
+    else trace "No project.nix found in ${directory}, skipping." false)
+  allProjectDirectories;
+in
+  mapAttrs (
+    _: directory: let
+      imported = import (directory + "/project.nix") args;
```
fricklerhandwerk, line 37: (not tested)

```suggestion
  mapAttrs (
    _: directory: let
      imported = import directory args;
```

Can't we just essentially say that each project dir must have a `default.nix` and everything will fall apart if any of them doesn't?

projects/mCaptcha/project.nix
```diff
@@ -0,0 +1,10 @@
+{pkgs, ...} @ args: {
+  packages = {inherit (pkgs) mcaptcha mcaptcha-cache;};
+  nixos = {
+    modules.service = ./service.nix;
+    tests = {
+      create-locally = import ./tests/create-locally.nix args;
+      bring-your-own-services = import ./tests/bring-your-own-services.nix args;
```
fricklerhandwerk, line 7:
```suggestion
{pkgs, sources }: {
  packages = {inherit (pkgs) mcaptcha mcaptcha-cache;};
  nixos = {
    modules.service = ./service.nix;
    tests = {
      create-locally = import ./tests/create-locally.nix { sources };
      bring-your-own-services = import ./tests/bring-your-own-services.nix { sources };
```

Explicit > implicit? IDK, maybe not abstracting over `args` will make it scale worse, but it feels from the current POV like we should say what exactly we're passing on. You'd essentially have to look at 3 files otherwise (top-level, this, the test file) to understand what's happening.

It could also be something like  `mapAttrs' (name: dir: ${stripSuffix name} = import dir { sources }; ) (readDir ./tests)`

projects/Flarum/project.nix
lorenzleutgeb: Yes, that's me debugging. Sorry.

projects/default.nix
lorenzleutgeb: Yes, though "falling apart" is not very nice for developing. For example we have the case of LiberaForms-E2EE which currently is unmaintained (#40). In such cases one might want to keep the directory and somehow have an option to skip it.

Of course, there's the alternatives of just deleting that stuff, or marking it as broken.

fricklerhandwerk: Let's just mark as broken.

projects/mCaptcha/project.nix
lorenzleutgeb: I changed to have explicit arguments, but still using `args`.

projects/default.nix
lorenzleutgeb: I added handling for `(import ./projects/<name>).broken` in `./projects/default.nix`.

lorenzleutgeb:
Depends on #185 to land first.

projects/Kbin/default.nix
```diff
@@ -5,7 +5,10 @@
 } @ args: {
   packages = {inherit (pkgs) kbin kbin-frontend kbin-backend;};
   nixos = {
-    modules.service = ./service.nix;
+    modules.service = {
+      path = ./service.nix;
+      options = "services\\.kbin\\..*";
+    };
```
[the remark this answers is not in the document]
lorenzleutgeb: That proposal works out for NixOS modules, yes. But not for, say, home-manager. We could do that by inverting things...

```nix
{
  modules = {
    nixos.services.kbin = ./service.nix;
    homeManager.programs.kbin-client-tool = ./hm.nix;
  };
}
```

But that's also not nice. Because we already have `nixos.{tests,configurations}`. So I suggest that we would add `homeManager` at the NixOS level once we actually have non-NixOS modules:

```
{
  nixos.modules.services.kbin = ./services.nix;
  homeManager.modules.programs.kbin-client-tool = ./hm.nix;
}
```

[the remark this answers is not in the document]
lorenzleutgeb: Wait... We also have to flatten the module names for flake outputs. So these names would get longer, e.g. `nixosModules."KBin/services/kbin"` but we could use the option path itself only, e.g. `nixosModules."services.kbin"`.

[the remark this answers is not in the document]
fricklerhandwerk: >  I suggest that we would add homeManager at the NixOS level once we actually have non-NixOS modules

That's what I had in mind, yes.

PR #173, opened 2024-04-11
meta-press, firefox-meta-press: init

wegank:
Ported from https://github.com/ngi-nix/meta-press. Tested on aarch64-linux.


wegank:
`Error: Resource not accessible by integration`

PR #178, opened 2024-04-12
Update Inputs

lorenzleutgeb:
We haven't updated inputs in a while, and there's some breakage. I fixed Kikit, but I believe that at least steamworks, Pretalx and wireguard-rs remain broken.

@fricklerhandwerk said @wegank could help me fix things. Once you start working, please leave a comment to tell me which package you are fixing on. I'll do the same so that we don't end up doing unnecessary work.

(I moved this from #123 to here, to avoid interference with the bot and have less noise.)


wegank:
steamworks: #179

wegank:
wireguard-rs: #180

wegank:
Ah, should I close them and push to this branch instead? Since these fixes are compatible with old revs, I thought sending them as separate PRs would make this one easier to review.

lorenzleutgeb:
I see. Yeah the question whether they are compatible with `main` was what added to my confusion :) Sorry about that. Merged. Thanks again!

wegank:
Looks like the PR is ready for review?

Issue #76, opened 2023-10-24
Re-Export from nixpkgs


lorenzleutgeb:
- [Rosenpass](https://github.com/NixOS/nixpkgs/pull/254813)
 - [Librecast](https://github.com/NixOS/nixpkgs/pull/248437)
 - [Flarum](https://github.com/NixOS/nixpkgs/pull/311608/)

cipharius:
* [arcan](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ar/arcan/package.nix)

Specifically A12 protocol libraries and arcan-net server/client binary, which is [part of NlNet project](https://nlnet.nl/project/Arcan-A12/). The package builds the whole arcan project though.

fricklerhandwerk:
Finished for the projects listed in the top comment.

PR #187, opened 2024-04-17
CONTRIBUTING.md: refactor

wegank:


CONTRIBUTING.md
```diff
@@ -1,47 +1,104 @@
...
+     nix build .#some-package.passthru.tests.some-test.driverInteractive
+     ./result/bin/nixos-test-driver # Start a shell
+     # Once in the spawned shell, start a VM that will execute the tests
+     start_all() # Run the VM
+     ```
+
+   - Format the Nix expressions with [Alejandra](https://github.com/kamadorueda/alejandra).
+
+     ```ShellSession
+     alejandra projects/some-module
```
lorenzleutgeb, line 89: I think we can abstract to `nix fmt`. This will even be valid in case we decide to use something different from Alejandra, and we're suggesting new-cli commands anyway, e.g. `nix build`.

```suggestion
   - Format the Nix expressions with [`nix fmt`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-fmt.html).

     ```ShellSession
     nix fmt projects/some-module
```

CONTRIBUTING.md
```diff
@@ -1,47 +1,104 @@
...
+- Package is not a [good candidate for Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#quick-start-to-adding-a-package).
+- Package is a good candidate for Nixpkgs, but no one is willing to be a maintainer :cry:.
+
+In any case, it is encouraged to create a pull request to Nixpkgs, then to this repository, with a comment linking to the pull request to Nixpkgs in the description and the Nix expressions.
+
+---
+
+Now that this is out of the way. To create pull requests to NGIpkgs:
+
+1. Set up a local version of NGIpkgs. If you don't have write access to NGIpkgs, create a [fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo#forking-a-repository) of the repository and clone it.
```
lorenzleutgeb, line 26: I'd suggest to directly link to [github.com/ngi-nix/ngipkgs/fork](https://github.com/ngi-nix/ngipkgs/fork), but of course there's value in *additionally* linking to the docs.

CONTRIBUTING.md
```diff
@@ -1,47 +1,104 @@
...
+
+   ```ShellSession
+   git checkout -b some-branch
+   ```
+
+1. To add a package, start by creating a `package.nix` file in the package directory `pkgs/by-name/some-package`, where `some-package` will be the package attribute name.
+
+   ```ShellSession
+   mkdir -p pkgs/by-name/some-package
+   some-editor pkgs/by-name/some-package/package.nix
```
lorenzleutgeb, line 43:
```suggestion
   $EDITOR pkgs/by-name/some-package/package.nix
```

That environment variable is understood (as a fallback) by e.g. [Git](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables), [GitHub CLI](https://cli.github.com/manual/gh_help_environment), [systemd](https://www.mankier.com/1/systemctl#Environment-%24SYSTEMD_EDITOR), [Mutt](http://www.mutt.org/doc/manual/) and relatively well known.

CONTRIBUTING.md
```diff
@@ -1,47 +1,104 @@
...
+
+     ```ShellSession
+     git add pkgs/by-name/some-package
+     nix build .#some-package
+     ```
+
+   - Format the Nix expressions with [Alejandra](https://github.com/kamadorueda/alejandra).
+
+     ```ShellSession
+     alejandra pkgs/by-name/some-package
```
lorenzleutgeb, line 58: See comment regarding `nix fmt` in line 89.

CONTRIBUTING.md
```diff
@@ -1,47 +1,104 @@
...
+     ```
+
+   To see how it’s done, an existing example is [libgnunetchat](https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/by-name/libgnunetchat/package.nix).
+
+1. To add a module, start by creating a `default.nix` file in the module directory `projects/some-module`, where `some-module/service` will be the module name.
```
lorenzleutgeb, line 63: I am not sure I understand this sentence correctly, but if I do, then it's on the verge of being outdated see #189.

CONTRIBUTING.md
```diff
@@ -1,47 +1,104 @@
...
+   ```shellSession
+   mkdir -p projects/some-module
+   some-editor projects/some-module/default.nix
+   ```
+
+   Make sure to:
+
+   - Add the module options in `service.nix`.
+
+   - Add the module tests in `test.nix`, or under a test directory, and expose them through `some-package.passthru.tests.some-test`, where `some-package` is a package attribute name.
```
lorenzleutgeb, line 74: Here we should mention that the way to access the NixOS test derivation in the package derivation definition is via `nixosTests.${projectName}`, see e.g.(

https://github.com/ngi-nix/ngipkgs/blob/a88709835a182babe2c7fbcf4db5424ebb735351/pkgs/by-name/mcaptcha/package.nix#L200

lorenzleutgeb (changes requested):
Consider writing one sentence per line. Not a blocker for me, but I wanted to mention it. See:
 - https://nick.groenen.me/notes/one-sentence-per-line/
 - https://hieuphay.com/one-sentence-per-line/
 - https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line
 - https://sive.rs/1s

CONTRIBUTING.md
fricklerhandwerk: > Note that there's a special case, e.g.
> 
> https://github.com/ngi-nix/ngipkgs/blob/66dac53297cfc1273fe2f40baa7e8ed08f7f04c7/projects/Rosenpass/default.nix#L8
>
> where the path of the module is set to null. The semantics is that there are some options that are relevant to the project, but these are not provided by a module in NGIpkgs.

I don't understand what that means. Where are those options provided then?

lorenzleutgeb: In a NixOS module that has been upstreamed to Nixpkgs. I am assuming that the same way that we want to re-export packages we want to re-export modules. We have #76 (which lacks a description) about this. The idea being that we do not want to capture strictly what's in NGIpkgs but also what was contributed to Nixpkgs.

fricklerhandwerk: Oh, but why wouldn't we point that to the source location then? E.g.

```
nixos.modules.rosenpass = "${sources.nixpkgs}/nixos/modules/services/rosenpass.nix;
```

lorenzleutgeb: Possible in principle, but we have to avoid loading it twice. If the module system figures this out since the path will be the same, it's easy and we have nothing to do. Otherwise we would have to figure out ourselves and that might be complicated. I can test that.

fricklerhandwerk: It will figure out itself if the Nixpkgs instance for the module system evaluation is the same as for obtaining the module expression. We will want that documented very clearly because it may not be evident and fail in inscrutable ways if the condition is violated.

PR #189, opened 2024-04-17
treewide, overview: Get option names by convention

lorenzleutgeb:
Follows up on https://github.com/ngi-nix/ngipkgs/pull/169#discussion_r1568901284

### DIFF
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -25,15 +25,14 @@
     concatLines
     flattenAttrsDot
     flip
+    hasPrefix
     mapAttrsToList
     optionalString
     ;
 
   empty = xs: assert isList xs; xs == [];
   heading = i: text: "<h${toString i}>${text}</h${toString i}>";
-  isPrefix = prefix: candidate: prefix == substring 0 (stringLength prefix) candidate;
-  concatStringsDot = concatStringsSep ".";
-  dottedLoc = option: concatStringsDot option.loc;
+  dottedLoc = option: concatStringsSep "." option.loc;
 
   lastModified = let
     sub = start: len: substring start len self.lastModifiedDate;
@@ -49,7 +48,7 @@
       spec = attrNames (flattenAttrsDot (project.nixos.modules or {}));
     in
       filter
-      (option: any ((flip isPrefix) (dottedLoc option)) spec)
+      (option: any ((flip hasPrefix) (dottedLoc option)) spec)
       (attrValues options);
     configurations = project: attrValues (project.nixos.configurations or {});
     packages = project: attrValues (project.packages or {});
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -1,4 +1,5 @@
 {
+  lib,
   options,
   pkgs,
   projects,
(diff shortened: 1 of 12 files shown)


lorenzleutgeb:
Don't get distracted by the failed deployment, I forgot to disable deploying for branches other than `main`, but at the same time also protected the deployment, resulting in an error to deploy. That's unrelated to this PR and fixed in 4ca2797f9b8327ae9654d004e6817ec446a3914e.

overview/default.nix
```diff
@@ -7,28 +8,32 @@
...
+    flip
     mapAttrsToList
     optionalString
-    unique
     ;
 
   empty = xs: assert isList xs; xs == [];
   heading = i: text: "<h${toString i}>${text}</h${toString i}>";
+  isPrefix = prefix: candidate: prefix == substring 0 (stringLength prefix) candidate;
+  concatStringsDot = concatStringsSep ".";
```
fricklerhandwerk, line 35:
```suggestion
  join = concatStringsSep;
```

Right?

This alias should actually go to Nixpkgs...

overview/default.nix
```diff
@@ -7,28 +8,32 @@
...
+  dottedLoc = option: concatStringsDot option.loc;
```
fricklerhandwerk, line 36: I think this can go into a `let` where it's used. Doesn't feel like it's worth a lib function.

overview/default.nix
```diff
@@ -7,28 +8,32 @@
...
+    flattenAttrsDot
+    flip
     mapAttrsToList
     optionalString
-    unique
     ;
 
   empty = xs: assert isList xs; xs == [];
   heading = i: text: "<h${toString i}>${text}</h${toString i}>";
+  isPrefix = prefix: candidate: prefix == substring 0 (stringLength prefix) candidate;
```
fricklerhandwerk, line 34: And I thought we have that in Nixpkgs?

fricklerhandwerk (approved):
Not sure if the services' flattened name convention is documented anywhere, because it's a bit weird/surprising at first glance. Looks good otherwise.

overview/default.nix
lorenzleutgeb: Ah, it's just called `hasPrefix`. Will fix.

overview/default.nix
lorenzleutgeb: Sure, will do.

lorenzleutgeb: `dottedLoc` is used in two places, and that's already the smallest scope that covers both.

overview/default.nix
lorenzleutgeb: Fixed in 70146c7.

overview/default.nix
lorenzleutgeb: I don't understand your suggestion. It removes a definition for `concatStringsDot` and adds a definition for `join` to alias `concatStringsSep`. But `concatStringsDot` was a partial application of `concatStringsSep`, not a plain alias.

I have removed `concatStringsDot` in 70146c7 because it was only used in one place anyway.

Regarding adding `join = concatStringsSep` to Nixpkgs, I do sympathize, but I am not sure I see the justification given that `concatStringsSep` is already there and adding another name adds complexity. Also, `concatStringsSep` is explicit about what is being concatenated here, while `join` is so generic that one might expect it to do other things, like operate on e.g. lists.

lorenzleutgeb: One would add it to `lib.strings`, yes, but it'd still be consumed as `lib.join`, see e.g. https://github.com/NixOS/nixpkgs/pull/254813#discussion_r1367015546

PR #190, opened 2024-04-17
overview: Add to checks

lorenzleutgeb:
Building the overview website is not very expensive (a lot cheaper than NixOS tests) but can catch some errors since it evals all modules.


lorenzleutgeb:
Don't get distracted by the failed deployment, I forgot to disable deploying for branches other than `main`, but at the same time also protected the deployment, resulting in an error to deploy. That's unrelated to this PR and fixed.

Issue #155, opened 2024-01-29
Disallow warnings in `nix flake check` or at least in CI


mightyiam:
To prevent issues such as #125 and #126.

mightyiam:
https://discourse.nixos.org/t/disallow-warnings-on-nix-flake-check/38964

mightyiam:
Thank you!

PR #191, opened 2024-04-17
ci: Set environment `NIX_ABORT_ON_WARN=true`

lorenzleutgeb:
Resolves ngi-nix/ngipkgs#155

See [`lib/trivial.nix`](https://github.com/NixOS/nixpkgs/blob/203fac824e2fdfed2e3a832b8123d9a64ee58b43/lib/trivial.nix#L700-L703).

Since this requires impure evaluation, and we want our flake to be pure in principle, split this into two separate evaluations, and also two jobs to hopefully save some time by parallelization. Caching the impure evaluation doesn't really make sense anyway.

### DIFF
--- a/pkgs/by-name/mcaptcha/frontend-package.json
+++ b/pkgs/by-name/mcaptcha/frontend-package.json
@@ -2,7 +2,7 @@
   "name": "vanilla",
   "main": "index.js",
   "version": "1.0.0",
-  "license": "AGPL-3.0",
+  "license": "AGPL-3.0-or-later",
   "scripts": {
     "build": "webpack --mode production",
     "lint": "yarn run eslint templates",
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -1,4 +1,4 @@
-name: nix flake check
+name: nix flake
 
 on:
   push:
@@ -8,6 +8,18 @@ on:
   workflow_dispatch:
 
 jobs:
+  eval:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: DeterminateSystems/nix-installer-action@main
+        with:
+          extra-conf: |
+            experimental-features = no-url-literals
+      - run: nix ${{ runner.debug && '--debug' }} flake check --impure --no-build --show-trace
+        env:
+          NIX_ABORT_ON_WARN: true
+
   check:
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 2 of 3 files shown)


pkgs/by-name/mcaptcha/frontend-package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vanilla",
   "main": "index.js",
   "version": "1.0.0",
-  "license": "AGPL-3.0",
+  "license": "AGPL-3.0-or-later",
```
fricklerhandwerk, line 5: Unrelated? Separate commit is fine though.

fricklerhandwerk:
I don't really understand the underlying problem. Is it that Nixpkgs/NixOS has impure expressions that fail hard on pure eval? How exactly does it help to split evaluations, and what is the partitioning?

pkgs/by-name/mcaptcha/frontend-package.json
lorenzleutgeb: Fixes evaluation for mCaptcha, to get to a successful build with the new constraint that all checks must eval without any warnings.

lorenzleutgeb:
> I don't really understand the underlying problem.

There's no problem per se. It's just that @mightyiam and me would like to avoid stuff reaching `main` that generates warnings, i.e. calls `lib.warn` somehow, during evaluation.

> Is it that Nixpkgs/NixOS has impure expressions that fail hard on pure eval?

In order to make `lib.warn` abort (like `-Werror` with `gcc`), you need to set the environment variable `NIX_ABORT_ON_WARN`. However, flakes evaluate purely by default, meaning that they don't have access to the network, information about time, the current machine, ... aaaand environment variables.

Therefore, we need to eval in impure mode to get the behavior we want, i.e. abort on warnings.

> How exactly does it help to split evaluations, and what is the partitioning?

There's one impure evaluation, which will error if there's a warning, and one pure evaluation, which will not error if there's a warning. Since they cannot reuse any evaluation results (well, one is impure the other isn't...) it does not make sense to somehow run them sequentially and benefit from evaluation caching (if that would be supported anyway, see https://github.com/NixOS/nix/issues/4279). Thus, run two jobs in parallel.

pkgs/by-name/mcaptcha/frontend-package.json
lorenzleutgeb: Upstream PR, just FYI, is https://github.com/mCaptcha/mCaptcha/pull/162

fricklerhandwerk:
I see, makes sense. Please add that sort of rationale into commit messages in the future (doesn't have to be that long).

PR #193, opened 2024-04-17
ci: Use Build Matrix

lorenzleutgeb:
Instead of building all `.#checks` with `nix flake check` directly, split this into two stages:

 1. Job "matrix":
    a. Evaluates `.#checks` and collects the digests of store paths.
    b. Contacts Cachix to find out which store paths are already cached.
    c. Prepares metadata for all store paths that are not cached.
    d. Submits the JSON from which the matrix is generated and all responses from Cachix as a GitHub artifact for inspection/debugging.
 2. Job "build" is instantiated once per flake output attribute path that must be built.
    a. Runs `nix build`.
    b. Pushes to Cachix.

The implementation is based on [nix-community/nix-github-actions](https://github.com/nix-community/nix-github-actions/), but crucially also exposes digests.

Relevant GitHub Actions documentation: [_Using a matrix for your jobs_](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)


flake.nix
```diff
@@ -187,8 +187,11 @@
         # `nix flake check` only evaluates packages and does not build them.
         (mapAttrs' (name: check: nameValuePair "packages/${name}" check) nonBrokenPkgs)
         // {
-          inherit (self.packages.${system}) overview;
```
lorenzleutgeb, line 190: Was already included via packages.

mightyiam:
Will be able to review beginning of May.

lorenzleutgeb:
I'll remove this again once #201 has landed and we have set up buildbot. In the meantime, it'll make GitHub Actions a little nicer.

PR #197, opened 2024-04-22
flake: replace treefmt-nix with pre-commit-hooks

wegank:


lorenzleutgeb:
Can we do the hook *in addition* to the check? It's very easy to `git commit --no-verify`...

fricklerhandwerk:
@lorenzleutgeb https://github.com/ngi-nix/ngipkgs/pull/197/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R187c does the check in CI, so that doesn't change semantics. And you can use pre-commit hooks to do lots of stuff, including YAML: https://github.com/cachix/git-hooks.nix

The main point is that it's done at commit time by default rather than waiting for CI to fail.

flake.nix
```diff
@@ -188,9 +184,21 @@
...
+            };
+          };
         };
 
+      devShell.${system} = pkgs.mkShell {
+        inherit (self.checks.${system}.pre-commit-check) shellHook;
+        buildInputs = self.checks.${system}.pre-commit-check.enabledPackages;
+      };
+
+      formatter.${system} = pkgs.alejandra;
```
lorenzleutgeb, line 200: @fricklerhandwerk please explain how, with this change, `nix fmt` will format YAML files. The way I understand this, only `alejandra` will be invoked, which will only format `*.nix` files.

Without this PR, using treefmt-nix, we could enable a YAML formatter in `treefmt.nix` and from that point on, `nix fmt` would format YAML files. The treefmt-nix formatter is currently registered (indirectly) in line 170, which is deleted in this PR as of a11678a.

fricklerhandwerk: I see. @wegank can we hook treefmt into pre-commit, or any other way of having the best of both worlds?

wegank: I've made `formatter.${system}` a shell script invoking `pre-commit` instead.

flake.lock
```diff
@@ -59,6 +96,22 @@
...
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-23.11",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs-stable_2": {
```
lorenzleutgeb, line 114: Looks like we're now pulling in two different versions of `release-23.11`. Could you please pin those down, so that they both follow the same input (maybe introducing `nixpkgs-stable` as an input to our Flake)?

flake.nix
```diff
@@ -188,9 +184,24 @@
...
+            };
+          };
         };
 
+      devShell.${system} = pkgs.mkShell {
+        inherit (self.checks.${system}.pre-commit-check) shellHook;
+        buildInputs = self.checks.${system}.pre-commit-check.enabledPackages;
+      };
+
+      formatter.${system} = pkgs.writeShellScriptBin "formatter" ''
```
lorenzleutgeb, line 200: Could you please you `pkgs.writeShellApplication`? It's not really an issue at the moment, but once people start adding stuff, this will shellcheck it.

flake.lock
wegank: Done.

flake.nix
wegank: Done.

lorenzleutgeb (approved):

PR #149, opened 2023-12-30
taler-wallet-core: init at 0.9.3

alejandrosame:
Fails with 

```
nix build .#wallet-core
warning: Git tree '/home/alesame/Dev/clients/nlnet/ngipkgs' is dirty
this derivation will be built:
  /nix/store/1ngx0bhhfhh98857qn2zbhffs3n3sr5v-wallet-core-0.9.3.drv
wallet-core> building '/nix/store/1ngx0bhhfhh98857qn2zbhffs3n3sr5v-wallet-core-0.9.3.drv'
wallet-core> Running phase: unpackPhase
wallet-core> unpacking source archive /nix/store/alsl49xzz2dwpljrq16ppcij3phqcsrf-wallet-core
wallet-core> source root is wallet-core
wallet-core> Running phase: patchPhase
wallet-core> applying patch /nix/store/c90i4h874w89wmpvwgmwh3ncadigna2y-no-git-bootstrap.patch
wallet-core> patching file bootstrap
wallet-core> applying patch /nix/store/jpx02dm4q43gyw7lx5m6v4fsb5rwn7d1-fix-broken-esbuild.patch
wallet-core> patching file pnpm-lock.yaml
wallet-core> Running phase: updateAutotoolsGnuConfigScriptsPhase
wallet-core> Running phase: configurePhase
wallet-core> Running phase: buildPhase
wallet-core> 
wallet-core> > @ compile /build/wallet-core
wallet-core> > pnpm run --filter '@gnu-taler/*' compile "--offline"
wallet-core> 
wallet-core> Scope: 17 of 18 workspace projects
wallet-core> packages/idb-bridge compile$ 
wallet-core> └─ Running...
wallet-core> packages/pogen compile$ 
wallet-core> │ 
wallet-core> └─ Failed in 406ms at /build/wallet-core/packages/pogen
wallet-core> packages/taler-util compile$ 
wallet-core> │ 
wallet-core> └─ Failed in 414ms at /build/wallet-core/packages/taler-util
wallet-core> /build/wallet-core/packages/pogen:
wallet-core>  ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @gnu-taler/pogen@0.0.5 compile: `tsc "--offline"`
wallet-core> Exit status 1
wallet-core>  WARN   Local package.json exists, but node_modules missing, did you mean to install?
wallet-core>  ELIFECYCLE  Command failed with exit code 1.

```

PR #198, opened 2024-04-25
wallet-core: init at 0.10.7

wegank:
An alternative of #149, without extra flake inputs.


pkgs/by-name/wallet-core-pnpm-deps/package.nix
fricklerhandwerk, line 1: Since it's not producing an actual stand-alone package, but just a bunch of dependencies, this file could be in `by-name/wallet-core`.

pkgs/by-name/esbuild_0_19_9/package.nix
fricklerhandwerk, line 1: Since this is a version specific to wallet-core, it may be worthwhile putting into the wallet core by-name directory. Optional

PR #203, opened 2024-04-27
flake: add aarch64-linux support

wegank:
Even if we don't want darwin, I guess we can at least have aarch64-linux?


lorenzleutgeb:
If we do this, then we should also enable CI for it. I think the argument against aarch64-linux was just to limit maintenance effort (less architectures, less work). @fricklerhandwerk please comment.

fricklerhandwerk:
I don't recall a discussion about aarch-linux. We had one about aarch-darwin, and decided to skip that because most services can be expected not to be run on Darwin, so fixing the plethora of build failures would not be worth it.

Unless there's a good reason to assume there will be substantial overhead, which would block other work or significantly eat into the budget, I think it should be fine to add aarch-linux.

lorenzleutgeb:
Alright, resolving conflicts and will merge soon.

PR #209, opened 2024-05-01
default.nix: init

wegank:
Too naive for now, not ready for review.


default.nix
```diff
@@ -0,0 +1,21 @@
+let
+  flakeLock = builtins.fromJSON (builtins.readFile ./flake.lock);
+  fetchFromGitHub = input:
+    with flakeLock.nodes.${input}.locked;
+    assert type == "github";
+      fetchTarball {
+        url = "https://github.com/${owner}/${repo}/archive/${rev}.tar.gz";
+        sha256 = narHash;
```
fricklerhandwerk, line 8: This re-implements [the lock-file parser in `flake-compat`](https://github.com/nix-community/flake-compat/blob/8bf105319d44f6b9f0d764efa4fdef9f1cc9ba1c/default.nix#L16-L77).

Unfortunately that code is not exported. Maybe fix upstream first by exporting that lock-file parser, and then [get `flake-compat` idiomatically](https://github.com/nix-community/flake-compat/blob/8bf105319d44f6b9f0d764efa4fdef9f1cc9ba1c/README.md#L58-L71).

fricklerhandwerk: There's apparently [a copy of that code in dream2nix](https://github.com/nix-community/dream2nix/blob/07bd44a3365efdbf4d17a14190636667e5fd3d48/dev-flake/flake-compat.nix#L21-L114)

default.nix
```diff
@@ -0,0 +1,7 @@
+{system ? builtins.currentSystem}: let
+  # Import nixpkgs and dream2nix
+  helpers = import ./helpers.nix {inherit system;};
+  inherit (helpers) importNgiPackages pkgs;
+  ngiPackages = importNgiPackages pkgs;
```
fricklerhandwerk, line 5: Defining inputs as arguments may be a slightly healthier pattern. This is a lot of speculation on my side though, because there's nothing established that I've seen, but all other patterns I've seen seem not as healthy. Here's [an example of what that could look like](https://github.com/fricklerhandwerk/lazy-drv/blob/6260dc5c0a542fe969174613df4b8f2352176002/default.nix#L18-L27).

One cool thing about the Nix language is that destructured function arguments have their own scope, and you can do arbitrary computation in default values.

helpers.nix
fricklerhandwerk, line 1: We may want something like a `./nix` directory for utilities. There you can lay out functions more liberally into separate files.

default.nix
```diff
@@ -0,0 +1,83 @@
+let
+  inputs = import ./flake-inputs.nix;
```
fricklerhandwerk, line 2: Incidentally I was working on exactly that question yesterday: should inputs be an argument or hard-coded? I came to the conclusion that they should be an argument, because you may have arbitrarily complex incantations in the default values for other arguments, and may only want to change the underlying source files. (Somewhat related, it may be good practice to export inputs in the output attrset, because that would allow patching the default value by calling the function without arguments, picking and updating the inputs, and calling the function with the updated inputs - essentially an `.override`)

The weird bit to me is the object flake-compat produces: it's already something evaluated rather than something akin to a lockfile entry (as in npins). Not sure we should open that can of worms here though...

default.nix
```diff
@@ -0,0 +1,83 @@
...
+in
+  {
+    nixpkgs ? inputs.nixpkgs,
+    dream2nix ? inputs.dream2nix,
+    rust-overlay ? inputs.rust-overlay,
+    sops-nix ? inputs.sops-nix,
+    self ? {
+      dirtyRev = "dirty";
+      lastModifiedDate = "19700101";
+    },
```
fricklerhandwerk, line 12: What exactly do we need that for?

default.nix
```diff
@@ -0,0 +1,83 @@
+let
+  inputs = import ./flake-inputs.nix;
+in
+  {
+    nixpkgs ? inputs.nixpkgs,
```
fricklerhandwerk, line 5: Is there any reason we can't create the actual values here already?

```suggestion
    pkgs ? import inputs.nixpkgs { inherit system; config = {}; overlays = []; },
```

default.nix
```diff
@@ -0,0 +1,83 @@
...
+    dream2nix ? inputs.dream2nix,
+    rust-overlay ? inputs.rust-overlay,
+    sops-nix ? inputs.sops-nix,
+    self ? {
+      dirtyRev = "dirty";
+      lastModifiedDate = "19700101";
+    },
+    system ? builtins.currentSystem,
+  }: let
+    lib = import (nixpkgs + "/lib");
```
fricklerhandwerk, line 15: If you have `pkgs` from arguments you get

```suggestion
    inherit (pkgs) lib;
```

default.nix
wegank: I only allow `outPath` from `flake.defaultNix.inputs`, though. I guess I'll put `fetchTree` somewhere in the tree to replace flake-compat.

default.nix
wegank: Overview, which inherits `inputs` from `flake.nix` to get info about the current repo (which I guess is invalid without Flakes).

wegank: Resolved by forking flake-compat...

default.nix
wegank: It's a lot more complicated than I thought, I ended up forking flake-compat and modifying it so that it doesn't touch `flake.nix` at all.

default.nix
wegank: There's the Rust overlay, I guess passing a different `pkgs` could break `atomic-server`?

wegank: Though I have no idea why the overlay is there, I'll open a test PR to see what happens if it's removed.

cleeyv:
I've been trying out this `default-nix-init` branch as part of testing the use of deploying NGIpkgs software with `npins` over at https://github.com/ngi-nix/npins-ngipkgs. 

Although I haven't gotten it working yet, since it can already be used to build packages I assume that it will be feasible to use this proposed `default.nix` to deploy packages from NGIpkgs to NixOS using npins. However, I get the impression that there would be more substantial changes required in order to use it to deploy _projects_ from NGIpkgs, which is currently one of the features of `flake.nix` version via `nixosModules`. Do you have a sense of how feasible it might be to implement something comparable for the `default.nix` as well?

PR #204, opened 2024-04-28
liberaforms: add NixOS tests

wegank:
Closes #199

PR #212, opened 2024-05-02
liberaformsEnv: drop

wegank:
The old `liberaforms` was defined with a ton of inheritances from `liberaformsEnv`, with the worry that something couldn't be done under dream2nix. Turns out I was wrong.

This PR also drops `liberaformsEnv.pyEnv` in favor of the new `liberaforms.env`.


wegank:
Tested with #204 (rebased on this branch) both on my machine and on CI. I'll think about how to separate the nixosTests change from that PR.

PR #170, opened 2024-04-11
atomic-browser: init at 0.37.0

albertchae:
- Use pnpm_8.fetchDeps to fetch dependencies and build
- Modeled after build steps in https://github.com/atomicdata-dev/atomic-server/blob/develop/browser/Earthfile

Co-authored-by: Anish Lakhwara <anish+git@lakhwara.com>

PR #220, opened 2024-05-03
Init atomic-server module

Chickensoupwithrice:
Building on work done in https://github.com/ngi-nix/ngipkgs/pull/170 (but not requiring it), this PR inits a module for atomic-server.
It's not ready for review but tested locally.

### DIFF
--- a/projects/AtomicData/service.nix
+++ b/projects/AtomicData/service.nix
@@ -1,0 +1,68 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.services.atomic-server;
+  # We need to add these for DirectoriesRS to pick up
+  # Since it doesn't pick up the ones set by systemd
+  envFile = pkgs.writeText ".env" ''
+    ATOMIC_CONFIG_DIR=/var/lib/atomic-server
+    ATOMIC_DATA_DIR=/var/lib/atomic-server
+    XDG_CACHE_HOME=/var/cache/atomic-server
+    ${generators.toINIWithGlobalSection {} {globalSection = cfg.settings;}}
+  '';
+in {
+  options = {
+    services.atomic-server = {
+      enable = mkEnableOption "Enable Atomic Server";
+      settings = mkOption {
+        default = {};
+        description = ''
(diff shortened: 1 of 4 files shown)


Chickensoupwithrice:
Alright! After many many months, I think this may finally be ready for review. 
We generate the .env file from settings, the systemd unit works on start up, and there is a test that passes. 

The main thing to note, is that the .env file does end up in the `/nix/store`, though there aren't any secrets in the .env file, so I think that's okay? Alternatively we could get the user to provide the environmentFile as a secret that doesn't end up in the nix store.

projects/AtomicData/service.nix
```diff
@@ -0,0 +1,67 @@
...
+    systemd.services.atomic-server = {
+      description = "Atomic Server";
+      after = [ "network.target" ];
+      wantedBy = [ "multi-user.target" ];
+      serviceConfig = {
+        ExecStart = "${pkgs.atomic-server}/bin/atomic-server";
+        User = "atomic-server";
+        EnvironmentFile = envFile;
+        StateDirectory = "atomic-server";
+        CacheDirectory = "atomic-server";
```
albertchae, line 51: So IIRC we discovered that this [systemd option sets the `CACHE_DIRECTORY` env var](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html), which turns out to ignored by the library atomic-server is using to configure these https://docs.rs/directories/latest/directories/struct.ProjectDirs.html#method.cache_dir, which is why we had to resort to adding `XDG_CACHE_HOME`.

Should we leave out these `*Directory` options then that won't actually impact the application? seems like they could be misleading

Chickensoupwithrice: The test fails without leaving `CacheDirectory` in as well, and I honestly can't explain it.

Chickensoupwithrice:
I'm not sure why the overview CI check is failing, or even what it does and how it's related to this PR. 
I will run a `nixfmt` once I'm back at my computer to fix the pre-commit checks.

fricklerhandwerk (approved):
Looks good! Amazing perseverance getting this done despite the stretched out timeline.

PR #227, opened 2024-05-14
projects: create a default interactive configuration for NixOS tests

wegank:
`interactive.nodes` looks pretty much the same in all NixOS tests, so why not make a default one.


fricklerhandwerk (approved):
Makes sense. Those are bells and whistles that do not interfere with the specifics of any test, so it's nice to have.

lorenzleutgeb (approved):
Yup, that was overdue. Thanks!

PR #201, opened 2024-04-26
infra: Migrate Repositories

lorenzleutgeb:
Towards #200:

 * Add `/CODOWNERS`, making @Erethon, @fricklerhandwerk, @Janik-Haag and me owners of `/infra`.
 * Merge the following two repositories into `/infra`, while preserving their history. Both repositories should be archived as soon as this PR is merged.
   * [ngi-nix/infra](https://github.com/ngi-nix/infra)
   * [ngi-nix/ngi0-infra](https://github.com/ngi-nix/ngi0-infra)
 * Remove most of the configuration of the "makemake" machine, in particular everything related to Hydra.
 * Start over with a new configuration.

I have deployed the new machine to a VM, it boots and is happy.

Since the PR diff doesn't really show what changes I made relative to the two previous infra repositories, [here](https://github.com/ngi-nix/ngipkgs/compare/d7d230f..infra)'s a more precise one: `git diff d7d230f..infra -- infra`

@Erethon: Happy to have your review. If you want, we can also set up a call and I walk you through.

@Ericson2314: Once this is deployed, which could be any time without further notice, [hydra.ngi0.nixos.org](https://hydra.ngi0.nixos.org/) will be gone for good. Are you still using it in anyway?

⚠️ Please don't squash merge this. I'd really like to keep the history of the NixOS/infra repo.

PR #230, opened 2024-05-16
flake.lock: deduplicate inputs

wegank:
https://github.com/ngi-nix/ngipkgs/pull/201#discussion_r1591183368


lorenzleutgeb (changes requested):
Thanks for cleaning up after me. But please let's have Hydra keep its own Nix and Nixpkgs.

PR #241, opened 2024-05-24
infra: Automatically deploy makemake

lorenzleutgeb:
As requested by @fricklerhandwerk.

Looks like this: https://github.com/ngi-nix/ngipkgs/deployments/makemake

I wanted to implement a fancier deployment via webhooks, but that turned out to be a rabbit hole. Can implement that when we have time. This works for now.

### DIFF
--- a/.github/workflows/makemake.yaml
+++ b/.github/workflows/makemake.yaml
@@ -1,0 +1,44 @@
+name: makemake
+
+on: { push: { branches: [ 'main' ] } }
+
+jobs:
+  deploy:
+    environment: makemake
+    runs-on: ubuntu-latest
+    env:
+      SSH_KEY: ${{ secrets.SSH_KEY }}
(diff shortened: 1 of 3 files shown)


Erethon (approved):
Looks reasonable to me.

lorenzleutgeb:
Security note: The secret which is the private key for deployment is scoped to the [environment "makemake"](https://github.com/ngi-nix/ngipkgs/settings/environments/3015629994/edit). This means that "regular" CI jobs don't have access, but only those that are [using that environment](https://docs.github.com/en/actions/using-jobs/using-environments-for-jobs) via `jobs.<job_id>.environment`. So, let's watch out for that when merging changes to `.github/workflows`.

fricklerhandwerk:
@lorenzleutgeb could you please document these details right in the code or contributor/maintainer instructions? Consider these threads ephemeral, no one will ever read them again.

PR #256, opened 2024-06-04
flake.nix: split lib.nix from nixpkgs lib

wegank:
Previously, `lib.nix` is merged into nixpkgs' [flake-dependent](https://github.com/NixOS/nixpkgs/blob/52a6b383fc60d6e75ee31457fd48b5a9495e5ba7/flake.nix#L18-L44) lib, which is then inherited in packages. Although this is not the case for now, it could make packages less upstreamable if non-nixpkgs / flake-dependent functions are accidentally used.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -32,7 +32,8 @@
     ...
   }: let
     # Take Nixpkgs' lib and update it with the definitions in ./lib.nix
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
Yes, makes sense.

PR #257, opened 2024-06-04
pkgs/by-name: make callPackage intrinsic

wegank:
Ideally, I would only need to pass the input sources (and the nixpkgs lib?) to `pkgs/by-name/default.nix` for it to work.

Tested with the following saved as `default.nix`, along with the `nix/flake-compat` folder from #209.

```nix
{
  system ? builtins.currentSystem,
  inputs ? import ./nix/flake-compat/inputs.nix,
  nixpkgs ? inputs.nixpkgs,
  dream2nix ? inputs.dream2nix,
}:
let
  pkgs = import nixpkgs { inherit system; };

  ngiPackages = import ./pkgs/by-name {
    inherit (pkgs) lib;
    inherit dream2nix pkgs;
  };
in
ngiPackages
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -69,17 +69,11 @@
         // args);
 
     # NGI packages are imported from ./pkgs/by-name/default.nix.
-    importNgiPackages = pkgs: let
-      callPackage = pkgs.newScope (
-        ngiPackages // {inherit callPackage;}
-      );
-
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
Ideally you could even pass the input sources as one argument. But generally that direction makes sense.

PR #258, opened 2024-06-05
configuraitons -> examples

fricklerhandwerk:
configurations are a legacy name, these really are examples for
consumer-facing documentation.


A-jay98:
We tried testing this locally by building overview and we noticed #259. So we're going to treat that as blocking this PR.

flake.nix
```diff
@@ -74,7 +74,7 @@
     importNgiProjects = {
       pkgs ? {},
       sources ? {
-        configurations = rawNixosConfigurations;
+        examples = rawNixosConfigurations;
```
lorenzleutgeb, line 77: Begs the question why the RHS is "configurations" then?

```suggestion
        examples = rawNixosExamples;
```

projects/Kbin/test.nix
```diff
@@ -20,7 +20,7 @@ in {
         sources.modules.default
         sources.modules."services.kbin"
         sources.modules.unbootable
-        sources.configurations."Kbin/base"
+        sources.examples."Kbin/base"
```
fricklerhandwerk, line 23: @lorenzleutgeb Unrelated, but that may be my next refactoring: is it for some reason *required* to flatten the nested attribute sets here? IIUC `checks` can be arbitrarily nested, and that's the only point where I see it would make a difference.

fricklerhandwerk: > `nixosConfigurations` cannot

Says who, actually? :thinking:

lorenzleutgeb: `nix flake show`, I guess? Would have to try it out again.

wegank: `nixosConfigurations` is something in the Flake output schema: https://github.com/NixOS/nix/blob/49c6f349116918d052a01b3c13cecbeaeeae657c/src/nix/flake.cc#L717

lorenzleutgeb: `checks` is too: https://github.com/NixOS/nix/blob/49c6f349116918d052a01b3c13cecbeaeeae657c/src/nix/flake.cc#L597-L617

Unfortunately that C++ code isn't something I can comprehend on the spot...

fricklerhandwerk: Yeah, that output schema officially doesn't exist, it's supposed to be defined by consumers, not Nix itself.

fricklerhandwerk: Okay, let me reword: The output schema is not part of any stable interface, and was explicitly intended to be user-defined - unfortunately, as things are, that is nowhere written down properly. The best I can give as proof as that more than half a year ago Nix maintainers discussed a [lack of output schema standardisation as a blocker for stabilising the CLI](https://discourse.nixos.org/t/2023-09-18-nix-team-meeting-minutes-87/33194). Please don't rely and don't make anyone rely on [experimental features](https://nix.dev/manual/nix/2.22/contributing/experimental-features): they are, and are supposed to be, subject to change, otherwise innovating becomes intractable.

fricklerhandwerk: To be clear, this is a tangential discussion, so let's not put too much brain power into it, but sometimes it's good to have it regardless:

From experience it's a fine line between cosmetics (a matter personal taste) and information structure (a property of our object of interest), where the second reflects on legibility, which in turn reflects on development agility. 

In this case I take issue with additional code being written to circumvent artificial constraints that should, in my opinion, never have existed, and on top of that produce an mismatch in how values are defined and accessed. That additional code imposes extra time on new readers and authors to understand what's happening, which ideally should be very little (apply some function to leaf attributes).

lorenzleutgeb: Sure, the code that flattens our project structure adds some overhead. The question is whether the benefit of conforming to a schema that (many) consumers are used to outweighs it. I do see value in the NGIpkgs flake behave like other flakes. We have traded users being able to observe an (experimental, though widespread) convention for maintaining some code that flattens names. Currently we don't have many users, but non-standard behavior, like violating the (experimental) flake schema, is what throws off users. They don't care whether it's experimental or not. If `nix flake show` complains and does not list configurations (which actually represent configuration examples) they will be confused.

PR #247, opened 2024-05-31
move NixOS VM tests to `checks` output

fricklerhandwerk:
this should capture all the tests to be run by buildbot

it also effectively deprecates Hydra builds

paired with @erethon; please run the formatter (development shell doesn't work on Darwin)

PR #265, opened 2024-06-06
factor out module for infra machine, remove hydra

fricklerhandwerk:
this puts information that belongs together next to each other rather
than having it in two completely different places. the `inputs` pattern
is not ideal and will need to be refactored for a non-flake setup, but
it works for now.

also remove the Hydra configuration, it's not used any more.

@wegank @lorenzleutgeb


lorenzleutgeb:
Conflicts with #247, so I'll wait for that other PR.

Ericson2314:
@fricklerhandwerk how come no hydra? I asked that we keep it before as this is our makeshift staging instance, and put in some work to try to keep its configuration out of the way.

fricklerhandwerk:
@Ericson2314 Ah, I was under the impression it's not actually getting used by anyone. We sure don't need it for this project, and I don't know who is even maintaining Hydra at all. Is there a way to move the staging to Nixpkgs or Nix, which seem to be the only major users?

infra/makemake/default.nix
```diff
@@ -0,0 +1,13 @@
+{inputs}:
```
lorenzleutgeb, line 1: I don't get it. This file doesn't seem to be used anywhere? `flake.nix` refers to `infra/makemake/configuration.nix` directly, see: https://github.com/ngi-nix/ngipkgs/pull/265/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R155

lorenzleutgeb (changes requested):

infra/makemake/default.nix
fricklerhandwerk: Maybe I missed something, have to take another look at the whole thing.

lorenzleutgeb (approved):

PR #273, opened 2024-06-07
sync: remove as it is upstreamed

wegank:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/sy/sync/package.nix


fricklerhandwerk (approved):
@wegank please add the reference to *where* it was upstreamed to the commit message.

PR #280, opened 2024-06-10
don't split description lines into paragraphs

fricklerhandwerk:
@wegank @lorenzleutgeb doing pre-processing of HTML inlined into markdown is asking for trouble


wegank (approved):

lorenzleutgeb:
The whole file is a wild mix of Markdown and HTML. So far it works out quite nicely.

PR #284, opened 2024-06-12
infra: Add fricklerhandwerk to sops config

lorenzleutgeb:
This is based on

```sh
% curl -s https://github.com/fricklerhandwerk.keys | ssh-to-age 2> /dev/null
age1028rq5gm5eakhycepydglqh0h2npxj3y4s4ls97ge02sw0aersysh3kqax
```

Please verify that this is indeed the key you would like to use (for example, compare with `ssh-to-age -i ~/.ssh/id_ed25519.pub`). If it is not, please update this PR accordingly (for example via GitHub's suggestion feature) to provide the desired key. Thanks!

### DIFF
--- a/infra/.sops.yaml
+++ b/infra/.sops.yaml
@@ -1,21 +1,24 @@
 .keys:
   .humans:
-    - &erethon       age187upwqdte7t0hkyec22jhac357m9y4fkcdvpg9sj5q9mekjupfnqg9a077
-    - &lorenzleutgeb age1c0g6s6daxy79dlm9uqczwlkh0hvjpghw5h8zzljc3vs275rvvqus30hv9l
+    - &erethon          age187upwqdte7t0hkyec22jhac357m9y4fkcdvpg9sj5q9mekjupfnqg9a077
+    - &fricklerhandwerk age1028rq5gm5eakhycepydglqh0h2npxj3y4s4ls97ge02sw0aersysh3kqax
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):

lorenzleutgeb:
Why do we *need* to? I certainly understand that it would be *possible* to do both in one PR. However, I wanted to ask @Janik-Haag next to add their key, and once we have all four keys, do the update once. The reason for asking them separately is to avoid a conflict (their GitHub usernames are neighbors when ordering the set of our four GitHub usernames lexicographically). Sure, in the meantime they won't be able to decrypt, but that's not worse than the status quo.

Erethon:
> Why do we _need_ to?

Ack, 'need' wasn't the best word for this. I meant it as a way to avoid turning this into a two step process where the second step can be forgotten. Also I wasn't aware that we'll add more people to the list with potential ordering conflicts.

lorenzleutgeb:
Yeah, sorry, I was not explicit about my intentions. My goal is to get all members of @ngi-nix/infra listed, and since the key update itself isn't that "interesting" (not human readable anyways) I think one of these commits is enough...

Janik-Haag:
Feel free to just add my key here :)
I can also do a PR later if you want me to, either works for me.
 
```plain
age1z7l86ldw44gjg3g6c4ztxcvmnf35q5umfam0w5rdc8pte60rzp4s5junff
```

fricklerhandwerk (approved):

Erethon (approved):

PR #292, opened 2024-06-20
kbin: mark as broken

### LINKED ISSUE #291: Kbin build fails with ` fatal: Authentication failed for 'https://github.com/nucleos/NucleosAntiSpamBundle.git/'`
https://buildbot.ngi.nixos.org/#/builders/183/builds/60#bb-step-1

This seems like an upstream problem. The project has not been updated for 4 months (it seems).

We will try to mark it as broken so that we can move on.

mightyiam:
closes #291

Co-authored-by: Adrien Faure <adrien.faure@protonmail.com>

### DIFF
--- a/pkgs/by-name/kbin-backend/package.nix
+++ b/pkgs/by-name/kbin-backend/package.nix
@@ -81,4 +81,6 @@ in
       inherit withS3;
       php = phpWithExtensions;
     };
+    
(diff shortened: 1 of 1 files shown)


stepbrobd:
I think what we want here is skipping packages that are marked as broken

PR #288, opened 2024-06-18
expose ntpd-rs

fricklerhandwerk:
closes #270

PR #289, opened 2024-06-18
expose mitmproxy

fricklerhandwerk:
closes #269

PR #305, opened 2024-06-25
expose forgejo

### LINKED ISSUE #268: Forgejo
Forgejo is [already in NixOS](https://search.nixos.org/options?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=forgejo). Re-expose it in the overview so it's listed as a supported project.

stepbrobd:
Closes #268

This is dirty, but modules from nixpkgs can be exported

If this is accepted, #288 can be modified to:

```diff
- modules.services.ntpd-rs = null;
+ modules.services.ntpd-rs = "${lib.modulePath}/services/networking/ntp/ntpd-rs.nix";
```

### DIFF
--- a/projects/Forgejo/default.nix
+++ b/projects/Forgejo/default.nix
@@ -1,0 +1,10 @@
+{
+  pkgs,
+  lib,
+  sources,
+} @ args: {
+  packages = {inherit (pkgs) forgejo;};
+  nixos = {
+    modules.services.forgejo = "${sources.inputs.nixpkgs}/nixos/modules/services/misc/forgejo.nix";
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -80,7 +80,7 @@
     # Note that modules and examples are system-agnostic, so import them first.
     rawNgiProjects = import ./projects {
       inherit lib;
-      sources = {};
+      sources = {inherit inputs;};
     };
 
     rawExamples = flattenAttrsSlash (mapAttrs (_: v: mapAttrs (_: v: v.path) v) (
@@ -118,6 +118,7 @@
       import ./projects {
         inherit lib pkgs;
         sources = {
+          inherit inputs;
           examples = rawExamples;
           modules = extendedNixosModules;
         };


stepbrobd:
Upstream NixOS tests can also be re-exposed like this, lmk if this wanted or not

projects/Forgejo/default.nix
```diff
@@ -0,0 +1,15 @@
...
+  sources,
+} @ args: let
+  inherit (lib) modulePath;
+in {
+  packages = {
+    inherit (pkgs) forgejo;
+  };
+
+  nixos = {
+    modules.services.forgejo = "${modulePath}/services/misc/forgejo.nix";
```
fricklerhandwerk, line 13: If we're taking this route, why can't we say more directly
 
```suggestion
    modules.services.forgejo = "${sources.nixpkgs}/nixos/modules/services/misc/forgejo.nix";
```

But I'm not sure whether hard-coding the paths is a good idea. Those may change arbitrarily, as they're not part of the external interface. Don't have to do it now, but it seems more explicit and non-magic to be able to say something like

```nix
modules.services.forejo = nixos.modules.services.forejo;
```

where we get `nixos` as an argument from somewhere

stepbrobd: I currently don't have a clear mental model of how to implement the following

> ```nix
> modules.services.forejo = nixos.modules.services.forejo;
> ```

I'll go with `"${sources.nixpkgs}/nixos/modules/services/misc/forgejo.nix";` for now

cleeyv:
I've done a build of the overview site for this and it looks good. I agree that making the path to the nixpkgs module explicit is much better than using `null`, and that the same should be done for the other PRs for exposing nixpkgs projects on the overview (#288 and #289). 

> Upstream NixOS tests can also be re-exposed like this, lmk if this wanted or not

That's an interesting idea, though I'm trying to think of what advantages there would be from doing it. Even if these exposed tests did run on our CI for ngipkgs, any changes to the projects would make more sense to happen via PRs to nixpkgs which would then run on their CI first. Our overview site currently only lists packages, options, and examples, and I'm not sure how useful it would be to add tests here.

cleeyv:
Also, if we are going to use the module paths from the nixpkgs input on all the nixpkgs exposure PRs, then this PR #305 should be merged first so that the `inherit inputs;` gets into the flake.nix first.

fricklerhandwerk (approved):

PR #306, opened 2024-06-25
PeerTube plugins: init official & livechat plugins, add module & test

OPNA2608:
For #8:

- Everything in [framasoft/peertube/official-plugins](https://framagit.org/framasoft/peertube/official-plugins) packaged
- [livechat plugin](https://github.com/JohnXLivingston/peertube-plugin-livechat) packaged
- `services.peertube.plugins` module added, plugins can be installed into peertube offline fully from the cache
- `nixosTests.PeerTube.peertube-plugins` test added

What's still missing from #8 is the [vosk-based transcription plugin](https://gitlab.com/apps_education/peertube/plugin-transcription). We've had some issues with packaging that, so we're submitting the whole suite of official plugins instead for now.

The Livechat package is an unholy mess, due to some issues with how the software expects to build itself. I've tried to add comments to explain why the different derivations are needed.

For getting the plugin installation to work, we're patching the peertube package to use `npm --offline` instead of `yarn`. This is because Yarn cannot handle the NPM cache format the packages are using, and (from my understanding) we can't build the packages via Yarn due to builder limitations (they all use `package.lock`, our Yarn builder expects `yarn.lock`). @Infinidoge is working on a separate patch to upstream for getting declarative plugin management starting, upon which we can hopefully build a better solution for this.

### DIFF
--- a/projects/PeerTube/test.nix
+++ b/projects/PeerTube/test.nix
@@ -1,0 +1,60 @@
+{
+  sources,
+  pkgs,
+  lib,
+  ...
+}: {
+  name = "peertube-plugins";
+
+  nodes = {
+    server = {config, ...}: {
+      imports = [
+        sources.modules.default
+        sources.modules."services.peertube.plugins"
+        sources.examples."PeerTube/base"
+      ];
+    };
+  };
+
+  testScript = {nodes, ...}: let
+    url = "http://${nodes.server.services.peertube.localDomain}:${toString nodes.server.services.peertube.listenWeb}";
+  in
+    ''
+      start_all()
+
+      with subtest("peertube works"):
+          server.wait_for_unit("peertube.service")
+          server.wait_for_console_text("Web server: ${url}")
+
+      # Eventually peertube-plugins-initial kicks in, sets up the initial state
+    ''
+    + (lib.strings.concatMapStringsSep "\n" (plugin: ''
+        with subtest("peertube plugin ${plugin.pname} installs"):
+            server.wait_for_console_text("Successful installation of plugin ${plugin}")
+      '')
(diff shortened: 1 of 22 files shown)


fricklerhandwerk (dismissed):
Great dirty hack, looking forward to the upstream fix!

projects/PeerTube/test.nix
```diff
@@ -0,0 +1,172 @@
...
+        windowManager.icewm.enable = true;
+      };
+
+      services.displayManager = {
+        defaultSession = "none+icewm";
+        autoLogin = {
+          enable = true;
+          user = localUser;
+        };
+      };
```
fricklerhandwerk, line 67: Is this needed for automated tests or just for logging into the test VM? If it's the second, one could add that in the VM test helpers: https://github.com/ngi-nix/ngipkgs/blob/main/projects/default.nix#L41-L52

projects/PeerTube/test.nix
```diff
@@ -0,0 +1,172 @@
...
+
+      environment = {
+        # To ease with interactive testing, use a fixed password
+        etc."peertube-envvars".text = ''
+          PT_INITIAL_ROOT_PASSWORD=changeme
+        '';
+      };
+
+      services.peertube = {
```
fricklerhandwerk, line 76: Please extract the configuration under test into a separate file that can be exposed as an example

OPNA2608: https://github.com/ngi-nix/ngipkgs/pull/306/commits/14dcedd1b597f8c6d5ba1de2710929db386e6cad#diff-59884480f9a83e10a7255d516e55d880d2d286e2d5134c1238294f4d35fb206c

fricklerhandwerk (approved):

PR #307, opened 2024-06-27
Allow makemake to be used for remote builds

Erethon:
This patchset changes the ssh-keys.nix structure a bit, adds a new user to `makemake` and adds some documentation. In particular:
- Following the reasoning from 5f64f82 and [this comment](https://github.com/ngi-nix/ngipkgs/pull/201#discussion_r1589802004), it removes @delroth's key from `ssh-keys.nix` to match nixos/infra#427.
- Add a new regular user called `remotebuild` to `makemake`.
- Change `ssh-keys.nix` to a `rec` and create two new lists, one for people that should have `root` access and one for people that should only have access to the newly created `remotebuild` user.
- Add documentation on how to get access to `makemake` in order to use it as a remote build host.


infra/makemake/configuration.nix
```diff
@@ -46,7 +46,11 @@
 
   users = {
     mutableUsers = false;
-    users.root.openssh.authorizedKeys.keys = import ../ssh-keys.nix;
+    users.root.openssh.authorizedKeys.keys = (import ../ssh-keys.nix).root;
+    users.remotebuild = {
+      isNormalUser = true;
+      openssh.authorizedKeys.keys = (import ../ssh-keys.nix).remotebuild;
```
fricklerhandwerk, line 52: A pattern that is very likely much easier to work with is having the keys in a directory, where each filename is the username. This is just a sketch:
```suggestion
      openssh.authorizedKeys.keyFiles = with lib;
        map (n: ./keys/${n}) (attrNames (builtins.readDir ./remotebuild));
```
You probably want to store those key groups in a `let` binding right here in the file, and combine them as needed, so whatever logic there is visible at first glance.

lorenzleutgeb (dismissed):

infra/makemake/configuration.nix
```diff
@@ -44,9 +44,19 @@
 
   time.timeZone = "Europe/Amsterdam";
 
-  users = {
+  users = let
+    groups = with lib.attrsets;
+      mapAttrs
+      (group: _:
+        mapAttrsToList (name: _: /keys/${group}/${name}) (builtins.readDir ./keys/${group}))
```
fricklerhandwerk, line 51:
```suggestion
        mapAttrsToList (name: _: ./keys/${group}/${name}) (builtins.readDir ./keys/${group}))
```

fricklerhandwerk (approved):
Thanks!

Issue #221, opened 2024-05-03
Migrate `weblate`


lorenzleutgeb:
### Repository URL

https://github.com/ngi-nix/weblate

### Notes

I saw that there's activity in that Flake by @erictapen and it's [on Flakestry](https://flakestry.dev/flake/github/ngi-nix/weblate).

wegank:
Does this mean introducing poetry2nix support in the repo? I tried to package Weblate with dream2nix, but failed because of nix-community/dream2nix#745.

erictapen:
Ah I think now I understand the goal of this issue. Feel free to ping me for questions on this.

lorenzleutgeb:
Sorry @erictapen, more context is here: #12

PR #309, opened 2024-07-02
weblate: init at 5.6.2

erictapen:
This adds the weblate package, module and integration test from my previous work at https://github.com/ngi-nix/weblate.

As this package attempt uses poetry2nix, which isn't part of Nixpkgs anymore, I had to add it as a Flake input to NgiPkgs. 
Also I noticed that I would benefit from the snakeoil TLS certificates used by many NixOS tests in Nixpkgs, so I figured it would make sense to expose nixpkgs as part of `sources` to every project?

I also update the package to the latest version 5.6.2.

Let me know if I should squash for inclusion, the commit messages are more or less written for myself.

Edit: Pinging @fricklerhandwerk as he asked me to do this, and @lorenzleutgeb as I saw your earlier branch.

Closes https://github.com/ngi-nix/ngipkgs/issues/221


projects/Weblate/tests/integration-test.nix
```diff
@@ -0,0 +1,123 @@
...
+    };
+
+    security.pki.certificateFiles = [certs.ca.cert];
+
+    networking.hosts."::1" = ["${serverDomain}"];
+    networking.firewall.allowedTCPPorts = [80 443];
+
+    # We need weblate-env available to the root user.
+    environment.systemPackages = config.users.users.weblate.packages;
+    users.users.weblate.shell = pkgs.bashInteractive;
```
fricklerhandwerk, line 67: Could you maybe pull out everything that's needed to self-host this into an example that will be listed in the overview? https://ngi-nix.github.io/ngipkgs/

projects/Weblate/service.nix
```diff
@@ -0,0 +1,408 @@
...
+          This should be a string, not a nix path, since nix paths are copied into the world-readable nix store.
+        '';
+        type = lib.types.path;
+      };
+
+      extraConfig = lib.mkOption {
+        type = lib.types.lines;
+        default = "";
+        description = ''
+          Text to append to <filename>settings.py</filename> Weblate config file.
```
wegank, line 161:
```suggestion
          Text to append to `settings.py` Weblate config file.
```

`<filename>` is not supported as child of `<td>`, generated in overview.

erictapen: Done!

projects/Weblate/tests/integration-test.nix
erictapen: Added a very basic but working one, is that enough? It's essentially what I have running at https://weblate.erictapen.name/ for testing.

erictapen:
Just noticed that the overview output auto-generates a URL to the project site: https://nlnet.nl/project/Weblate

But I can't find a Weblate project in https://nlnet.nl/project/? Is it just not listed maybe?

fricklerhandwerk:
Oh, it may never have been funded by NLnet. I suppose we'll have to adjust how we do the attribution.

projects/Weblate/tests/integration-test.nix
fricklerhandwerk: But that's not being tested, right? Users still have to somehow know all the other things that would go into a working setup, essentially everything in this test case here.

erictapen: No, everything else in the `server` configuration is purely there to make it work in the test environment. Big chunk is due to the absence of a publicly reachable Letsencrypt infrastructure (as we are in a sandbox).

fricklerhandwerk (approved):

erictapen:
Had to fix an override for aeidon, which probably broke due to a Python upgrade.

fricklerhandwerk (approved):

PR #326, opened 2024-08-01
atomic-server: refactor to rebuild from github source

albertchae:
This refactor builds atomic-server using the separately packaged
atomic-browser and then the github source. Not only is this more
reflective of the real build process used, but it is helpful when trying
to debug atomic-server by building a fork with additional logging, which
we did while working on https://github.com/ngi-nix/ngipkgs/pull/220

Previously we were building atomic-server from the published crate,
which is usually "from source". But in this case that meant only the
rust source, and separately built web assets were packaged into the
crate. This can be verified by running
`curl -L 'https://crates.io/api/v1/crates/atomic-server/0.37.0/download' | tar -tv`
and checking for the `assets_tmp` directory.

Part of https://github.com/ngi-nix/ngipkgs/issues/15

### DIFF
--- a/pkgs/by-name/atomic-server/package.nix
+++ b/pkgs/by-name/atomic-server/package.nix
@@ -1,7 +1,8 @@
 {
   lib,
-  fetchCrate,
+  fetchFromGitHub,
   rustPlatform,
+  atomic-browser,
 }: let
   inherit
     (lib)
@@ -12,12 +13,22 @@ in
     pname = "atomic-server";
     version = "0.37.0";
 
-    src = fetchCrate {
-      inherit pname version;
-      hash = "sha256-/OKYac0HA9EWDQ5qNyMPITN5iUdLM9SAVmOm6PVIFOk=";
+    src = fetchFromGitHub {
+      owner = "atomicdata-dev";
+      repo = pname;
+      rev = "v${version}";
+      hash = "sha256-+Lk2MvkTj+B+G6cNbWAbPrN5ECiyMJ4HSiiLzBLd74g=";
     };
 
-    cargoHash = "sha256-LwSyK/7EEoTf1x7KGtebPxYTqH3SCjXGONNMxcmdEv0=";
+    cargoHash = "sha256-cSv1XnuzL5PxVOTAUiyiQsMHSRUMaFDkW2/4Bt75G9o=";
+
+    # server/assets_tmp is the directory atomic-server's build will check for
+    # compiled frontend assets to decide whether to rebuild or not
+    # https://github.com/atomicdata-dev/atomic-server/blob/ba3c5959867a563d4da00bb23fd13e45e69dc5d7/server/build.rs#L22-L37
+    postUnpack = ''
+      mkdir -p source/server/assets_tmp
+      cp -r ${atomic-browser}/* source/server/assets_tmp
+    '';
 
     doCheck = false; # TODO(jl): broken upstream
 


pkgs/by-name/atomic-server/package.nix
```diff
@@ -12,12 +13,22 @@ in
...
+
+    cargoHash = "sha256-cSv1XnuzL5PxVOTAUiyiQsMHSRUMaFDkW2/4Bt75G9o=";
 
-    cargoHash = "sha256-LwSyK/7EEoTf1x7KGtebPxYTqH3SCjXGONNMxcmdEv0=";
+    # server/assets_tmp is the directory atomic-server's build will check for
+    # compiled frontend assets to decide whether to rebuild or not
+    # https://github.com/atomicdata-dev/atomic-server/blob/ba3c5959867a563d4da00bb23fd13e45e69dc5d7/server/build.rs#L22-L37
+    postUnpack = ''
+      mkdir -p $PWD/source/server/assets_tmp
+      cp -r ${atomic-browser}/* $PWD/source/server/assets_tmp
```
albertchae, line 30: both $PWD and $TMP seemed to work here, which one or is another option recommended?

fricklerhandwerk: You could even do:

```suggestion
      mkdir -p source/server/assets_tmp
      cp -r ${atomic-browser}/* source/server/assets_tmp
```

pkgs/by-name/atomic-server/package.nix
```diff
@@ -12,12 +13,22 @@ in
...
+
+    # server/assets_tmp is the directory atomic-server's build will check for
+    # compiled frontend assets to decide whether to rebuild or not
+    # https://github.com/atomicdata-dev/atomic-server/blob/ba3c5959867a563d4da00bb23fd13e45e69dc5d7/server/build.rs#L22-L37
+    postUnpack = ''
+      mkdir -p $PWD/source/server/assets_tmp
+      cp -r ${atomic-browser}/* $PWD/source/server/assets_tmp
+    '';
 
     doCheck = false; # TODO(jl): broken upstream
```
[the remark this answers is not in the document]
albertchae: not sure but I also saw this. Was going to verify it independently and resolve in a separate PR if so

pkgs/by-name/atomic-server/package.nix
```diff
@@ -1,7 +1,8 @@
 {
   lib,
-  fetchCrate,
+  fetchFromGitHub,
   rustPlatform,
+  atomic-browser,
 }: let
```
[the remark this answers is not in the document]
albertchae: I agree but I was just following the pattern introduced in https://github.com/ngi-nix/ngipkgs/pull/122 and still appears to be followed in several package definitions. If we change this, we should change it everywhere and figure out a way to enforce that

pkgs/by-name/atomic-server/package.nix
albertchae: 🤦 of course a relative path would implicitly point to `$PWD` by default. Verified your change still builds, if it's the favored style I can switch to it

pkgs/by-name/atomic-server/package.nix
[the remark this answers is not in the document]
albertchae: looks like it still fails. We do need to bump atomic-server to 0.38.0 soon (after https://github.com/ngi-nix/ngipkgs/pull/220 is merged), at that point I'll verify whether the tests are still failing and leave a more detailed comment why + contact upstream about fixing it

pkgs/by-name/atomic-server/package.nix
fricklerhandwerk: IIUC this is the common thing to do.

albertchae: thanks, updated!

fricklerhandwerk (approved):

PR #365, opened 2024-09-19
projects: sanitize

wegank:
This PR sanitizes the output of `import ./projects { ... }`, so that one doesn't need `or {}` or even auxiliary functions to extract modules/examples/tests from the output.


fricklerhandwerk (approved):
Oh yes, very nice.

PR #379, opened 2024-10-10
refactor: simplify attrs flattening

fricklerhandwerk:
this reduces unneeded generality. we probably shouldn't do the premature
attribute flattening at all, it's not clear to me why it's even done to
begin with. the only place where it seems unavoidable is for the
`checks` flake output.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -34,22 +34,12 @@
     lib' = import ./lib.nix {inherit lib;};
 
     inherit
-      (builtins)
-      mapAttrs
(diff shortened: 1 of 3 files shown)


wegank:
I once thought about renaming `ngipkgs.nixosModules."services.kbin"` to `ngipkgs.nixosModules.kbin`, but that's a breaking change that I'd like to finally avoid.

wegank (approved):

PR #405, opened 2024-11-19
add more packages to AtomicData

fricklerhandwerk:


wamirez (approved):

fricklerhandwerk:
It seems one of the packages does not have a `version` field, however that can happen

PR #412, opened 2024-11-20
Add Naja related library

wamirez:


fricklerhandwerk (commented):
Hm, is that by the same project or is exposed as a standalone component? Because if it's just a dependency that works behind the scenes, no need to list it here.

wamirez:
> Hm, is that by the same project or is exposed as a standalone component? Because if it's just a dependency that works behind the scenes, no need to list it here.

The latter. I see. I'll close this one out.

PR #415, opened 2024-11-21
Add CNSPRCY nixpkgs ref and missing artifacts

wamirez:


fricklerhandwerk (approved):
Thanks again, leaving a reference to how to configure the service is great!

PR #421, opened 2024-11-21
Add Agorakit nixpkgs and NixOS refs

wamirez:
There's [no dedicated test](https://github.com/NixOS/nixpkgs/pull/345014/files#diff-307c155c790abf7b54b5dec19dc230c6d692c6dca0604760d8389c7aed543c10) yet, therefore it was defined as `null`.

The build fails though with the following error.

> jq: parse error: Expected string key before ':' at line 1, column 71

Maybe something along the lines of jq breaking the css file due to too many items? Can't really tell from the get go.

### DIFF
--- a/projects/Agorakit/default.nix
+++ b/projects/Agorakit/default.nix
@@ -1,0 +1,14 @@
+{
+  pkgs,
+  lib,
+  sources,
+} @ args: {
+  packages = {
+    inherit (pkgs) agorakit;
+  };
+  nixos = {
+    modules.services.agorakit = "${sources.inputs.nixpkgs}/nixos/modules/services/web-apps/agorakit.nix";
+    tests.basic = "${sources.inputs.nixpkgs}/nixos/tests/web-apps/agorakit.nix";
+    examples = null;
+  };
+}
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
Our invocation of the tests fails because it seems we're not filtering out `null` when collecting them.

projects/Agorakit/default.nix
```diff
@@ -0,0 +1,14 @@
...
+  pkgs,
+  lib,
+  sources,
+} @ args: {
+  packages = {
+    inherit (pkgs) agorakit;
+  };
+  nixos = {
+    modules.services.agorakit = "${sources.inputs.nixpkgs}/nixos/modules/services/web-apps/agorakit.nix";
+    tests.agorakit = null;
```
fricklerhandwerk, line 11:
```suggestion
    tests = null;
```

Our invocation of the tests fails because you gave it `tests.agorakit` that was supposed to be a test but is `null`. Our filter only removes `tests` if the whole attribute set is `null`.

fricklerhandwerk (changes requested):

projects/Agorakit/default.nix
wamirez: It doesn't seem to be that though. Maybe a Nixpkgs bump is needed then?

fricklerhandwerk (approved):

PR #422, opened 2024-11-21
Fix projects names to match NLnet URLs

wamirez:


wamirez:
Actually, I'll do the same with `git mv` instead in order to preserve the commit history.

PR #427, opened 2024-11-23
Add KiKit Nixpkgs refs

wamirez:


projects/KiKit/default.nix
```diff
@@ -0,0 +1,8 @@
+{pkgs, ...} @ args: {
+  packages =
+    {
+      inherit (pkgs) kikit;
+      inherit (pkgs.kicadAddons) kikit-library;
+    }
+    // {kicad-kikit = pkgs.kicadAddons.kikit;};
```
fricklerhandwerk, line 7:
```suggestion
      kicad-kikit = pkgs.kicadAddons.kikit;
    };
```

No need to compute the attrset update

fricklerhandwerk (dismissed):

projects/KiKit/default.nix
wamirez: There's a `kikit` for both `pkgs` and `pkgs.kicadAddons`. Renaming the latter seemed the easiest approach. How would you call the `kikit` add-on?

PR #431, opened 2024-11-23
Add wireguard-rs ngipkgs ref

wamirez:


fricklerhandwerk:
Added everything we have from WireGuard. Depends on #439 because for some reason Cryptpad tests could never have worked but CI didn't catch it. @wegank any idea how that's even possible?

PR #435, opened 2024-11-24
Add arpa2 Ngipkgs refs

wamirez:


projects/arpa2/default.nix
```diff
@@ -0,0 +1,16 @@
+{pkgs, ...} @ args: {
+  packages = {
+    inherit
+      (pkgs)
+      freeDiameter
```
fricklerhandwerk, line 5:
```suggestion
```

this is only a dependency, not part of arpa2

fricklerhandwerk (approved):

wegank (approved):

PR #437, opened 2024-11-24
Add GNUnet-Msn Ngipkgs ref

wamirez:


fricklerhandwerk:
@wamirez it makes more sense to tuck it under GNUnet, because that's the parent application/project and will be recognized as such.

PR #447, opened 2024-11-27
Change formatting to rfc style

jasonodoom:
These changes bake a full-reformat commit and exclude that from blame.

In addition, the `alejandra` hook in `flakes.nix` is replaced with `nixfmt-rfc-style`. The hash from `nix fmt` formatting changes in `.git-blame-ignore-revs` is replaced with the hash from formatting changes from`nixfmt-rfc-style`. `.gitattributes` is added to ignore whitespace during bisect.


.git-blame-ignore-revs
lorenzleutgeb, line 1: You should not remove the old hash and comment, but rather append your new hash and comment to this file so that _both_ formatting commits are ignored. Just because you do another reformatting, that does not mean that the older reformatting now somehow gets relevant again.

lorenzleutgeb (dismissed):

.git-blame-ignore-revs
jasonodoom: > Just because you do another reformatting, that does not mean that the older reformatting now somehow gets relevant again.

Aah, that was my understanding initially. Thanks for the clarification. Please see: [8166d9f](https://github.com/ngi-nix/ngipkgs/pull/447/commits/8166d9f7e779eb7a1dc876e82d0170ba2ebecfb6)

wegank:
Merge conflict, please rebase.

wegank:
Merge conflict again...

fricklerhandwerk:
Oh this won't be any fun as long as we keep working on `main`... :)

jasonodoom:
@wegank thanks :)

fricklerhandwerk:
Nice, thanks a lot @jasonodoom!

fricklerhandwerk (approved):

PR #450, opened 2024-11-27
makemake: store key files in flat directory

fricklerhandwerk:
group membership is not exclusive, and should be managed at a different
layer, but the mapping from name to SSH key is neatly implemented with
file names and their content


Erethon (approved):
Haven't tested this, but looks reasonable to me

PR #453, opened 2024-11-28
get all projects from default.nix

fricklerhandwerk:
this reduces the number of moving parts and finally removes a code path that was unnecessarily hard to reason about

towards https://github.com/ngi-nix/ngipkgs/pull/209

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,9 +28,9 @@
     buildbot-nix,
     ...
   } @ inputs: let
+    classic' = import ./. {system = null;};
     # Take Nixpkgs' lib and update it with the definitions in ./lib.nix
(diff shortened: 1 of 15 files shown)


flake.nix
```diff
@@ -211,7 +189,7 @@
           in
             checksForNixosTests // checksForNixosExamples;
         in
-          concatMapAttrs checksForProject (importNgiProjects (pkgs // nonBrokenPackages));
```
fricklerhandwerk, line 214: Note that this is not a problem. If we had broken packages filtered out that tests happen to rely on, everything would collapse anyway. We'd need to filter broken tests as well.

wamirez (approved):

PR #459, opened 2024-11-29
Rosenpass: increase timeout for wireguard test

fricklerhandwerk:
5s is too flaky


fricklerhandwerk:
This fixes Rosenpass, but leaves a bunch of broken builds left over from the GC

PR #462, opened 2024-11-29
fix builds

fricklerhandwerk:
- fix how modules are collected
- fix overview build
- fix checks

squashing attribute paths into strings considered harmful


fricklerhandwerk:
Also nuked the eval cache, because false-negative build failures kept accumulating

PR #487, opened 2025-02-04
projects: init deliverables

eljamm:
Python models for the projects' deliverables.

### DIFF
--- a/projects/deliverables.py
+++ b/projects/deliverables.py
@@ -1,0 +1,62 @@
+from enum import Enum
+
+
+class URL(str):
+    pass
+
+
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Thanks for putting this up. This is our core data structure and everyone on the @ngi-nix/team should be closely familiar with it.

Implemnetation note: we probaby want it as e.g. a [yants](https://code.tvl.fyi/about/nix/yants) type in `default.nix`, and then see from there.

eljamm:
> Is this a duplicate of #507 now?

Yes, let's close this.

Issue #504, opened 2025-02-12
Extract non-sensitive project data from NLnet sources and store it in tree


erictapen:
We need a non-sensitive version of the metadata extracted by the script `extract-project-metadata.py` from https://github.com/ngi-nix/summer-of-nix/pull/114. The result of that process should live in tree, probably in the `overview/` directory.

eljamm:
Added the option to extract data from the NLnet dashboard in a way that's suitable for the project overview in https://github.com/ngi-nix/summer-of-nix/pull/146

erictapen:
So the kind of basic structure that @eljamm and I discussed yesterday would look like this.

`SoftwareProject` is a *project* identifier, so it's something we invent manually. `subgrants` contains the identifier that are used on https://nlnet.nl/project/.

```
{
  "SoftwareProject": {
    "subgrants": [ "SoftwareProjectBlib", "SoftwareProjectBlub" ],
    "name": "Software Project",
    "summary": "A software Project that makes everything",
    websites: ["https://github.com/foo/bar", "https://foo.bar"],
    …
  },
  …
}
```

erictapen:
Sounds reasonable!

eljamm:
> So the kind of basic structure that @eljamm and I discussed yesterday would look like this.

Yup, this looks good to me.  

> SoftwareProject is a project identifier, so it's something we invent manually.

One thing to note here is that since a `project` is not a concept that exists in the data, but rather something we're inventing, its summary will be something we'll have to manually create, or we'd need to figure out which summary from the subgrants to use.

Also, `subgrants` doesn't necessarily need to be a list as we could have it as a dict, which will allow us more freedom when accessing the data. As such, the structure could be something like this:

```json
{
  "SoftwareProject": {
    "name": "Software Project",
    "summary": "A software project that makes everything",
    "subgrants": {
      "SoftwareProjectBlib": {
        "name": "SoftwareProjectBlib",
        "summary": "Part of the software project that makes everything"
      },
      "SoftwareProjectBlub": {
        "name": "SoftwareProjectBlub",
        "summary": "Another part of the software project that makes everything"
      }
    },
    "websites": ["https://github.com/foo/bar", "https://foo.bar"]
  }
}
```

eljamm:
I agree that we should continue discussing this in code, but I think a better place for this would be in https://github.com/ngi-nix/ngipkgs/pull/507

PR #507, opened 2025-02-14
Add typed models for projects

eljamm:
Can be tested with:

```sh
nix-instantiate --eval --strict --expr "let self = import ./default.nix {}; in self.project-models.example"
```

You can also use [watchexec](https://search.nixos.org/packages?channel=unstable&show=watchexec&from=0&size=50&sort=relevance&type=packages&query=watchexec) to continuously monitor the output as the file changes:

```sh
watchexec '<COMMAND>'
```

If this is the correct way to go, project deliverables will be added next.

### DIFF
--- a/projects/models.nix
+++ b/projects/models.nix
@@ -1,0 +1,53 @@
+{
+  lib,
+  pkgs,
+  sources,
+}:
+let
+  yants = import sources.yants { };
+
+  inherit (yants)
+    string
+    list
+    option
+    attrs
+    ;
+in
+rec {
+  project =
+    p: with p; {
+      name = string name;
+      metadata = with metadata; {
+        summary = string summary;
+        fund = string fund;
+        status = string status;
+
+        websites = with websites; {
+          repo = string repo;
+          docs = option string (websites.docs or null);
+          other = list string other;
+
+          contact = option (attrs (option string)) {
+            email = contact.email or null;
+            forum = contact.forum or null;
+            matrix = contact.matrix or null;
+            blog = contact.blog or null;
+          };
+        };
+      };
+    };
+
+  example = project {
+    name = "";
+    metadata = {
+      summary = "";
+      websites = {
+        repo = "";
+        contact = null;
+        other = [ ];
+      };
+      fund = "";
+      status = "";
+    };
+  };
+}
--- a/default.nix
+++ b/default.nix
@@ -63,6 +63,8 @@ rec {
     };
   };
 
+  project-models = import ./projects/models.nix { inherit lib pkgs sources; };
+
   # TODO: find a better place for this
   metrics = with lib; {
     projects = attrNames raw-projects;
--- a/flake.lock
+++ b/flake.lock
@@ -267,7 +267,8 @@
         "nixpkgs-stable": "nixpkgs-stable",
         "pre-commit-hooks": "pre-commit-hooks",
         "sops-nix": "sops-nix",
-        "systems": "systems"
+        "systems": "systems",
+        "yants": "yants"
       }
     },
     "slimlock": {
@@ -346,6 +347,22 @@
         "owner": "numtide",
         "repo": "treefmt-nix",
         "type": "github"
+      }
+    },
+    "yants": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1645270620,
+        "narHash": "sha256-wwkl3K200UbW9Z7BRlVH8HOEXCaVYP2MqZpsF9EhgZg=",
+        "ref": "refs/heads/canon",
+        "rev": "efeb6dc11eb1a1e88d41dc2093fc5aa31f7abd35",
+        "revCount": 15,
+        "type": "git",
+        "url": "https://code.tvl.fyi/depot.git:/nix/yants.git"
+      },
+      "original": {
+        "type": "git",
+        "url": "https://code.tvl.fyi/depot.git:/nix/yants.git"
       }
     }
   },
--- a/flake.nix
+++ b/flake.nix
@@ -13,6 +13,8 @@
   inputs.sops-nix.url = "github:Mic92/sops-nix";
   inputs.buildbot-nix.inputs.nixpkgs.follows = "nixpkgs";
   inputs.buildbot-nix.url = "github:nix-community/buildbot-nix";
+  inputs.yants.url = "git+https://code.tvl.fyi/depot.git:/nix/yants.git";
+  inputs.yants.flake = false;
 
   # See <https://github.com/ngi-nix/ngipkgs/issues/24> for plans to support Darwin.
   inputs.systems.url = "github:nix-systems/default-linux";


projects/models.nix
```diff
@@ -0,0 +1,53 @@
...
+        summary = string summary;
+        fund = string fund;
+        status = string status;
+
+        websites = with websites; {
+          repo = string repo;
+          docs = option string (websites.docs or null);
+          other = list string other;
+
+          contact = option (attrs (option string)) {
```
erictapen, line 30: Why is a contact in the websites attrs?

projects/models.nix
```diff
@@ -0,0 +1,53 @@
...
+    ;
+in
+rec {
+  project =
+    p: with p; {
+      name = string name;
+      metadata = with metadata; {
+        summary = string summary;
+        fund = string fund;
+        status = string status;
```
erictapen, line 23: is this something like packaged/not yet packaged/not planned? Could be an enum then.

erictapen (commented):
Nice, looking forward to have some kind of model in tree that we can start iterating on.

We might want to consider adding extensive doc comments in `projects/models.nix`, as this is a file that is likely to be looked at by contributors a lot.

Also I'm not sure I yet understand the exact scope of the project model. Is it supposed to eventually contain only the external data we get from NLnet or also the kind of information that we have from the package set itself, e.g. versions, vulnerabilities, labels?

projects/models.nix
fricklerhandwerk, line 23: Funds can be calculated from subgrants, and we should probably have those as a list of enum here.

fricklerhandwerk: There's a status field in the NLnet dashboard, but that's not very useful for us here, and really just an accounting thing. It's supposed to say something about handling status by partners. We keep this in our internal system already and that's good enough.

projects/models.nix
```diff
@@ -0,0 +1,56 @@
...
+    p: with p; {
+      name = string name;
+      metadata = with metadata; {
+        summary = string summary;
+        funds = list (enum [
+          "Commons"
+          "Core"
+          "Entrust"
+          "Review"
+        ]) funds;
```
fricklerhandwerk, line 29: Maybe leave funds out as long as we don't have code/processes for adding values. As discussed elsewhere, that information is attached to subgrants, and we want those separate anyway so we can link to https://nlnet.nl/project.

projects/models.nix
```diff
@@ -0,0 +1,56 @@
...
+          repo = string websites.repo;
+          docs = option string (websites.docs or null);
+          blog = option string (websites.blog or null);
+          forum = option string (websites.forum or null);
+          matrix = option string (websites.matrix or null);
+          other = list string (websites.other or [ ]);
+        };
+        contact = {
+          email = option string (contact.email or null);
+        };
```
fricklerhandwerk, line 40: It's alright to keep the field if we don't expose that information from the NLnet dashboard. In any case I suggest not adding that if we don't have a concrete use case for it.

projects/models.nix
```diff
@@ -0,0 +1,45 @@
...
+    either
+    ;
+in
+rec {
+  project =
+    p: with p; {
+      name = string name;
+      metadata = with metadata; {
+        summary = string summary;
+        websites = attrs (either (option string) (list string)) {
```
eljamm, line 24: In https://github.com/ngi-nix/ngipkgs/issues/504#issuecomment-2666515871, it was metioned that 

> `websites` will now be denormalised because those actually come from each subgrant -- if we keep them there, we can re-aggregate them in presentation if needed.

Does it make sense to look at this PR's model as the representation?

projects/models.nix
```diff
@@ -0,0 +1,126 @@
...
+    name = string;
+    documentation = optionalStruct {
+      config = option string;
+    };
+    examples = nonEmtpyAttrs (option exampleType);
+  };
+
+  exampleType = struct "example" {
+    description = option string; # TODO: should this be non-optional?
+    path = either string path;
```
eljamm, line 42: I forgot, what did we say `path` will be used for?

fricklerhandwerk: The path to the expression we use as an example (see some projects that have examples, e.g. Atomic Data)

projects/models.nix
```diff
@@ -0,0 +1,128 @@
...
+    enum
+    either
+    struct
+    drv
+    path
+    restrict
+    ;
+
+  programType = struct "program" {
+    name = string;
```
fricklerhandwerk, line 23:
```suggestion
    name = option string;
```

I don't see the need for a human-readable name *yet*, so we could leave that optional.

projects/models.nix
```diff
@@ -0,0 +1,128 @@
...
+    source = drv;
```
fricklerhandwerk, line 24: We decided programs should also be NixOS modules (`programs.<symbolic application name>`), so the actual sources would be an implementation detail we later think about whether and how expose it in the UI. So arguably that could just as well be

```suggestion
    module = defun [attrs attrs];
```

fricklerhandwerk:
I guess the next meaningful step to get this merged is to hook up (and refactor as needed) at least one actual project description with real values. This sort of slightly-larger-scale refactoring can be kicked off by adding an entirely new code path for constructing the projects collection, and adding a migration so the old stuff doesn't break.

E.g. we could move `//projects` to `//projects-old`, and have the `//projects-old/default.nix` convert the new format (after the type check!) to the old format, and migrate projects incrementally `//projects-old` is empty. That way
- we're always exercising the new code without having to rework the overview rendering code
- have a progress indicator right there in the code
- the old code appropriately looks like a garbage dump that needs to be gotten rid of until it's not needed any more and can be discarded

projects/models.nix
```diff
@@ -0,0 +1,128 @@
...
+    function
+    any
+    ;
+
+  programType = struct "program" {
+    name = option string;
+    module = defun [
+      (attrs any)
+      (attrs any)
+    ] function any;
```
fricklerhandwerk, line 29: I'm not sure on the exact syntax here, can you please add an example to check?

fricklerhandwerk: I'd like to see the shape of an actual module, namely `{...}: {}`. Maybe I read the docs too superficially, but I understood we'd write `defun [(attrs any) (attrs any)]` and then apply `{...}: {}` as a value to be type-checked.

projects/Aerogramme/default.nix
```diff
@@ -1,10 +1,15 @@
...
+  pkgs,
+  sources,
+}@args:
+{
+  name = "Aerogramme";
   nixos = {
     # https://aerogramme.deuxfleurs.fr/documentation/cookbook/config/
     # https://aerogramme.deuxfleurs.fr/documentation/cookbook/service-manager/
     modules.services = null;
+    modules.packages = { inherit (pkgs) aerogramme; };
```
fricklerhandwerk, line 12: This needs to be a proper `programs` module, but not sure if Aerogramme even has standalone executables, so likely this line just needs to go away (which would correctly reflect taht we're not particularly far with packaging that one)

```suggestion
```

eljamm: Does that mean that we should keep `modules.packages`? If not, where do we put the packages in the new structure for the current overview?

fricklerhandwerk: As discussed in one of the many discussions so far, the Nixpkgs notion of packages (which is largely a derivation with a bunch of data tacked on) is not particularly useful for us apart from providing files to work with, and it's merely an implementation detail for the our collection of NGI applications.

projects/models.nix
eljamm: There were issues making `defun` work inside structs, but when it's it's defined outside of it, then it works as you've described. I've included a slightly more elaborate example than a `{...}: {}` to show the result.

projects/Aerogramme/default.nix
```diff
@@ -1,6 +1,10 @@
...
-  packages = { inherit (pkgs) aerogramme; };
+  lib,
+  pkgs,
+  sources,
+}@args:
+{
+  name = "Aerogramme";
   nixos = {
     # https://aerogramme.deuxfleurs.fr/documentation/cookbook/config/
     # https://aerogramme.deuxfleurs.fr/documentation/cookbook/service-manager/
```
fricklerhandwerk, line 10:
```suggestion
    # https://aerogramme.deuxfleurs.fr/documentation/cookbook/service-manager/
    # derivation at https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/by-name/ae/aerogramme/package.nix
```

Maybe worth mentioning

projects/models.nix
```diff
@@ -0,0 +1,152 @@
...
+      config = option string;
+    };
+    examples = nonEmtpyAttrs (option exampleType);
+  };
+
+  exampleType = struct "example" {
+    description = option string; # TODO: should this be non-optional?
+    path = either string path;
+    documentation = option string;
+    tests = nonEmtpyAttrs drv;
```
fricklerhandwerk, line 66: Aha, tying tests to example right there in the is actually very clever. But do we want standalone "non-example" tests at all then?

projects/models.nix
```diff
@@ -0,0 +1,152 @@
...
+    documentation = optionalStruct {
+      config = option string;
+    };
+    examples = nonEmtpyAttrs (option exampleType);
+  };
+
+  exampleType = struct "example" {
+    description = option string; # TODO: should this be non-optional?
+    path = either string path;
+    documentation = option string;
```
fricklerhandwerk, line 65: Generally for reference we probably want a structured URL type along these lines:

```nix
  urlType = struct "URL" {
    text = string;
    # we may later want to do a fancy syntax check in a custom `typdef`
    link = string;
  };
```

projects/models.nix
```diff
@@ -0,0 +1,152 @@
...
+    name = option string;
+    path = either string path;
+    documentation = optionalStruct {
+      config = option string;
+    };
+    examples = nonEmtpyAttrs (option exampleType);
+  };
+
+  exampleType = struct "example" {
+    description = option string; # TODO: should this be non-optional?
```
fricklerhandwerk, line 63: Yes, bare code is useless to most people.

```suggestion
    description = string;
```

projects/models.nix
```diff
@@ -0,0 +1,152 @@
...
+
+          foobar-cli = {
+            name = "foobar-cli";
+            module = mkModule {
+              name = "foobar";
+              src = "";
+
+              # not included in the final attribute
+              test = "";
+            };
```
fricklerhandwerk, line 136: I expect it to be an actual module:

```suggestion
            module = { lib, ... }: {
              enable = lib.mkEnableOption "foobar CLI";
            };
```

fricklerhandwerk:
Looks to me like this is almost there, thank you very much.

Some side notes simply to drop them somewhere, not relevant for merging this:
- We're missing libraries and plugins here, as well as a few (currently fringe) deliverables such as mobile apps or arbitrary files (e.g. like with Libre-SOC), i.e. stuff that is not necessarily NixOS specific. This okay and we can add them as we go.
- Plugins could be a per-project type defined in the relevant program/service module
- Libraries could be values for use in a per-project service/module field (akin to `python.withPackages`), but would be specified in a per-ecosystem type, e.g. under `<project>.libraries.<ecosystem>.<name>`.
  - Sigh, we'd need to curate ecosystems, but fortunately we're only dealing with a few there and can go about this incrementally.
  - We'll probably need a migration path from "Nixpkgs code goo where everything is a derivation with random attrs attached" though, since we already quite a bunch of those libraries packaged and would want to show them even if we don't have a proper type figured out.

projects/models.nix
fricklerhandwerk: Yeah sorry I actually did misread the docs. `defun` does runtime input/output typechecks, and while that's nice we don't need or want that here. See the other comment

projects/Aerogramme/default.nix
```diff
@@ -1,9 +1,14 @@
...
+    # derivation at https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/by-name/ae/aerogramme/package.nix
     modules.services = null;
```
fricklerhandwerk, line 12: Thinking about it, in order to not lose already existing information we may want an absolutely minimal module here, such as

```suggestion
    modules.services = { lib, pkgs, ... }: {
        options = {
          enable = lib.mkEnableOption "Aerogramme " ;
          package = lib.mkPackageOption pkgs "aerogramme" { };
        }
    };
```

projects/models.nix
eljamm: This makes sense to me. I assume we should do the same for `programs` and `services`, but do we want to make them into an `attrs` or restrict which types of links are allowed with `struct`?

> ... and possibly later auto-add a __toString attribute or even various output formats so we can string interpolate this in e.g. example descriptions for correct-by-construction URLs.

Yup, I believe this is possible as well.

projects/models.nix
eljamm: That's a good question. I currently think of `project.tests` as a pool which we can be reused in other places. Two examples might need the same test, so in that case it makes sense to just link them there:

```nix
examples.basic = {
  tests = {
    inherit (tests) basic;
  };
};
examples.advanced = {
  tests = {
    inherit (tests) basic advanced;
  };
};
tests.basic = "...";
tests.advanced = "...";
```

This also means we can have internal tests that aren't necessarily meant to be exposed through examples.

On the other end, we could put such tests in something like `project.examples._tests` or somewhere similar:

```nix
examples = {
  _tests = {
    description = "Tests that don't have examples";
    module = { ... }: { };
    tests.basic = "";
    tests.advanced = "";
  };
};
```

projects/models.nix
fricklerhandwerk: Where we know (or expect) a certain set of URLs I'd go for struct. Free form stuff such as additinal references can be attrs.

projects/models.nix
```diff
@@ -0,0 +1,155 @@
...
+  optionalStruct = set: option (struct set);
+  optionalAttrs = set: option (attrs set);
+  nonEmtpyAttrs = t: restrict "non-empty-attrs" (a: a != { }) (attrs t);
+  absPath = restrict "absolute-path" (p: lib.pathExists p) (either path string);
+in
+rec {
+  project = struct {
+    name = string;
+    metadata = optionalStruct {
+      summary = option string;
```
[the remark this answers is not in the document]
fricklerhandwerk: As we discussed at some point, the source is an implementation detail somewhere deep inside a derivation. And as I mentioned in the top level comment, I'd prefer proper structured data for ecosystem stuff. A framework will have certain properties and refer only to a certain part of the project, which may have multiple components written in different languages.

projects/AtomicData/default.nix
```diff
@@ -1,20 +1,22 @@
 { pkgs, ... }@args:
 {
-  packages = {
-    inherit (pkgs)
-      atomic-server
-      atomic-browser
-      atomic-cli
-      ;
-  };
+  # TODO: put these in modules
```
fricklerhandwerk, line 3: Let's not leave it in a broken state. Either port it completely or ditch it

fricklerhandwerk:
Left "review comments" in the form of commits. None of the TODOs should get addressed in this PR, but it does need fixing the build by adapting the example and ported projects.

fricklerhandwerk (approved):

eljamm:
CI now fails because [MarginaliaSearch](https://github.com/ngi-nix/ngipkgs/tree/aa262ae08f9029acb2aaf82bf75e847ed8617e2a/projects/MarginaliaSearch) will exist in the `projects` folder after the merge, and not `projects-old`. Should I rebase and move it or do we merge this first and then fix that later?

projects/Aerogramme/default.nix
```diff
@@ -28,7 +28,8 @@
             package = lib.mkPackageOption pkgs "aerogramme" { };
           };
           # TODO: add a service definition
-          meta.broken = true;
+          # NOTE: enabling this fails with "The option `meta.broken' does not exist"
```
fricklerhandwerk, line 31: Hm, I wonder how NixOS does it then, because it certainly has those meta fields...

projects/models.nix
```diff
@@ -45,23 +50,17 @@ let
   serviceType = struct "service" {
     name = option string;
     module = moduleType;
-    references = optionalStruct {
-      config = option urlType;
-    };
-    examples = nonEmtpyAttrs (option exampleType);
+    references = optionalAttrs (option urlType);
```
fricklerhandwerk, line 53: Please rename to `links` (for brevity) as mentioned earlier

projects/Aerogramme/default.nix
eljamm: I don't think it was breaking before. We can investigate this later, though.

fricklerhandwerk:
Please make it cleanly mergeable. I recommend resetting and slicing new commits.

fricklerhandwerk (approved):

PR #543, opened 2025-03-05
Add prefix and update label of NGI project issue template

imincik:
Make NGI project tasks more visible.

This change will require rename of currently existing `project` label.


fricklerhandwerk (approved):

fricklerhandwerk:
I merged that a bit too quickly. Thinking about it, since we already have labels for the various funds which all start with NGI, maybe we should call it just "project" -- or if that's too vague, maybe "application" since that's slightly less ambiguous with respect to what *we* call subgrants, which NLnet calls projects.

PR #555, opened 2025-03-07
add CryptoLyzer

fricklerhandwerk:
Closes https://github.com/ngi-nix/ngipkgs/issues/549

PR #556, opened 2025-03-07
Fix project mapping

eljamm:
Fixes: https://github.com/ngi-nix/ngipkgs/pull/555

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -76,30 +76,29 @@ rec {
 
           examples-from =
             value:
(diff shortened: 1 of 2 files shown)


fricklerhandwerk (approved):

fricklerhandwerk:
The next thing would be updating the backwards-compat mapping so we can see examples from `nixos.modules.programs` in the overview.

Issue #548, opened 2025-03-07
Implement project metadata for Private Key Operations for Keyoxide


imincik:
Implement project metadata in 'projects/Private Key Operations for Keyoxide/default.nix'

PR #558, opened 2025-03-07
Add Keyoxide

eljamm:
Closes: https://github.com/ngi-nix/ngipkgs/issues/548

~~Depends on https://github.com/ngi-nix/ngipkgs/pull/560 being merged~~

### DIFF
--- a/projects/Keyoxide/default.nix
+++ b/projects/Keyoxide/default.nix
@@ -1,0 +1,20 @@
+{
+  lib,
+  pkgs,
+  sources,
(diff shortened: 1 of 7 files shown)


eljamm:
Apparently, `services` and `programs` can't have the same name since in [extendedNixosModules](https://github.com/ngi-nix/ngipkgs/blob/716f0f56604f76aa731b4dd47386028a91a7a0c5/flake.nix#L79-L80), the former will always override the latter.

PR #559, opened 2025-03-07
Libervia-backend: Extract from old Libervia, add to new project structure

### LINKED ISSUE #553: Implement project metadata for Libervia
Implement project metadata in 'projects/Libervia-backend/default.nix'

OPNA2608:
Closes #553

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -48,6 +48,16 @@ rec {
     }
     // foldl recursiveUpdate { } (map (project: project.nixos.modules) (attrValues projects));
 
+  extendedNixosModules =
+    with lib;
+    [
+      nixos-modules.ngipkgs
+      # TODO: needed for examples that use sops (like Pretalx)
+      sops-nix
+    ]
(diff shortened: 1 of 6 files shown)


fricklerhandwerk:
Would it make sense to tuck under Libervia, or are the other components distinct applications as opposed to just components of the same application?

OPNA2608:
#553 named "Libervia-backend" specifically, so I extracted only that part of the project here. If that was meant to include the desktop client in that, then I have no issue with migrating that into here as well, putting it all under "Libervia", and dropping `/projects-old/Libervia`.

fricklerhandwerk:
Yes, please put everything together! :)

imincik (approved):
Thanks ! Code looks good to me.

My only concern is that we lack some conventions or templates for file naming and code structure. Currently, all project metadata looks slightly different.

fricklerhandwerk:
We'll get to stronger patterns once there's enough material to generalize over.

Issue #561, opened 2025-03-10
project-models.example is failing to evaluate


imincik:
When trying to evaluate example code in [projects/models.nix](https://github.com/ngi-nix/ngipkgs/blob/716f0f56604f76aa731b4dd47386028a91a7a0c5/projects/models.nix#L108) it fails with following error message:

```
nix-instantiate --eval --strict --expr "let self = import ./default.nix {}; in self.project-models.example"

error:
       … while evaluating the attribute 'project-models.example'
         at /home/imincik/Projects/ngi/ngipkgs/code/projects/models.nix:108:3:
          107|
          108|   example = project {
             |   ^
          109|     name = "foobar";

       … while calling a functor (an attribute set with a '__functor' attribute)
         at /home/imincik/Projects/ngi/ngipkgs/code/projects/models.nix:108:13:
          107|
          108|   example = project {
             |             ^
          109|     name = "foobar";

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: expected 'anon'-struct, but found:
       field 'nixos': expected 'nixos'-struct, but found:
       field 'examples': expected type option<attrs<example>>, but value does not conform to 'attrs<example>': expected type attrs<example>, but found:
       {
         description = ''
           This is how you can run `foobar` in the terminal.
         '';
         links = {
           website = {
             text = "FooBar Documentation";
             url = "https://foo.bar/docs";
           };
         };
         module = <function>;
         tests = {
           foobar-cli = <derivation myname>;
         };
       }: expected 'example'-struct, but found:
       field 'module': expected type 'either<path, string>', but value '<function>' is of type 'lambda'


       field 'modules': expected 'modules'-struct, but found:
       field 'programs': expected type option<attrs<option<program>>>, but value does not conform to 'attrs<option<program>>': expected type attrs<option<program>>, but found:
       {
         examples = {
           foobar-cli = {
             description = ''
               This is how you can run `foobar` in the terminal.
             '';
             links = {
               website = {
                 text = "FooBar Documentation";
                 url = "https://foo.bar/docs";
               };
             };
             module = <function>;
             tests = {
               foobar-cli = <derivation myname>;
             };
           };
           foobar-tui = null;
         };
         module = <function, args: {lib}>;
         name = "foobar-cli";
       }: expected type option<program>, but value does not conform to 'program': expected 'program'-struct, but found:
[...]

PR #562, opened 2025-03-10
Fix project-models.example

eljamm:
Type expects an absolute path or a string, not a function

Fixes: https://github.com/ngi-nix/ngipkgs/issues/561


imincik (approved):
Thanks. Looks good and it works.

PR #566, opened 2025-03-10
Aerogramme: fix module option

eljamm:
This PR makes the `Aerogramme` service option appear in the overview. It also removes `meta.broken` from the module as I don't think is a thing for NixOS modules (correct me if I'm wrong).

### DIFF
--- a/projects/Aerogramme/default.nix
+++ b/projects/Aerogramme/default.nix
@@ -23,13 +23,13 @@
       module =
         { lib, pkgs, ... }:
         {
-          options = {
+          options.services.aerogramme = {
             enable = lib.mkEnableOption "Aerogramme";
             package = lib.mkPackageOption pkgs "aerogramme" { };
           };
           # TODO: add a service definition
-          # NOTE: enabling this fails with "The option `meta.broken' does not exist"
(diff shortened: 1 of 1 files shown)


projects/Aerogramme/default.nix
```diff
@@ -23,13 +23,11 @@
...
         { lib, pkgs, ... }:
         {
-          options = {
+          options.services.aerogramme = {
             enable = lib.mkEnableOption "Aerogramme";
             package = lib.mkPackageOption pkgs "aerogramme" { };
           };
           # TODO: add a service definition
-          # NOTE: enabling this fails with "The option `meta.broken' does not exist"
-          # meta.broken = false;
```
fricklerhandwerk, line 32: It's deliberately left as a todo, we need a way to express something doesn't work as intended yet.

fricklerhandwerk (approved):

projects/Aerogramme/default.nix
eljamm: Made the comment more clear.

imincik (approved):
Looks good, thanks !

PR #569, opened 2025-03-10
Cleanup nixos modules; pass sops-nix directly

eljamm:
Since we're only interested in the attribute sets' values, we can directly use lists, instead of updating the attrs, then calling `attrsValues`.

With this, we avoid shadowing `programs` and `services` that share the same name.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -52,6 +52,8 @@ rec {
     with lib;
     [
       nixos-modules.ngipkgs
+      # TODO: needed for examples that use sops (like Pretalx)
+      sops-nix
     ]
     ++ attrValues nixos-modules.programs
     ++ attrValues nixos-modules.services;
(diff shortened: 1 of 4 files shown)


default.nix
```diff
@@ -48,6 +48,16 @@ rec {
     }
     // foldl recursiveUpdate { } (map (project: project.nixos.modules) (attrValues projects));
 
+  extendedNixosModules =
+    with lib;
+    [
+      nixos-modules.ngipkgs
+      # TODO: needed for examples that use sops (like Pretalx)
+      sops-nix
```
eljamm, line 56: Can we perhaps make the examples import the modules they need individually? This would make them more self-contained.

fricklerhandwerk: Yes absolutely. IIUC&IIRC historically the reason to export them was that examples were built as full configurations, and collecting them at the flake level was practical to thread them into CI. We don't need that any more, the examples are just files we want people to copy and edit. The tests are still needed for CI but we can collect them out of the evaluated fix point.

fricklerhandwerk (approved):
Didn't review in depth but if there's no diff in output store paths it's good.

PS: having a helper script around to check store path diff would be nice...

PR #570, opened 2025-03-10
overview: Stable project urls

### LINKED ISSUE #505: Stable project URLs
As a project author, I need the overview page for my project to be advertisable on one stable URL.

My idea would be to go with `https://{NGIPKGS_DOMAIN}/{PROJECT_NAME}` for maximum brevity but maybe  `https://{NGIPKGS_DOMAIN}/project/{PROJECT_NAME}` is more wise.

The initial page should contain information that is already available in tree:

- Project name
- pnames
- versions
- options
- examples

erictapen:
Closes #505 

Beware that this currently mixes the names of `projects/` and `projects-old/`, even though the names in `projects-old/` are actually grant names, which we don't want to use. As NGIpkgs is experimental and the overview isn't even hosted in it's designated place yet, I didn't care yet to write logic to only use names from `projects/`.

Currently this uses the following URL format for the URL of e.g. the Canaille project: `https://ngi.nixos.org/Canaille`

tmp C0CpqKLLiM

tmp UPjDaWf5V4

### DIFF
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -30,6 +30,8 @@ let
     mapAttrsToList
     optionalString
     recursiveUpdate
+    mapAttrs'
+    nameValuePair
     ;
 
   empty =
@@ -146,17 +148,23 @@ let
         '';
     };
 
-    projects = rec {
+    projects = {
       one = name: project: ''
-        <section><details><summary>${heading 2 name}</summary>
+        ${heading 1 name}
         <https://nlnet.nl/project/${name}>
 
         ${render.packages.many (pick.packages project)}
         ${render.options.many (pick.options project)}
         ${render.examples.many (pick.examples project)}
-        </details></section>
       '';
-      many = projects: concatLines (mapAttrsToList one projects);
+      # Many projects are renderes as links to their individual project sites
+      many =
+        projects:
+        concatLines (
+          mapAttrsToList (name: _: ''
+            <a href="${name}">${name}</a>
+          '') projects
+        );
     };
   };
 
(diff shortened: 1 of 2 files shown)


overview/default.nix
```diff
@@ -146,34 +148,79 @@ let
...
         ${render.examples.many (pick.examples project)}
-        </details></section>
       '';
-      many = projects: concatLines (mapAttrsToList one projects);
+      # Many projects are renderes as links to their individual project sites
+      many =
+        projects:
+        concatLines (
+          mapAttrsToList (name: _: ''
+            <a href="${name}">${name}</a>
```
fricklerhandwerk, line 165:
```suggestion
            <a href="project/${name}">${name}</a>
```

The issue discussed namespacing for better extensibility, and I agree with that.

fricklerhandwerk (dismissed):
Nice!

eljamm:
Well done! This works great when using a HTTP server and it looks and behaves just as expected:

```
$ nix build .#overview
$ nix-shell -p httplz --run 'httplz ./result/'
```

That said, this isn't the case when opening the overview file directly:

```
$ xdg-open ./result/index.html
```

- When I click on a project link, it doesn't take me directly to its page, but rather:

image

Then I'd have to click on `index.html` to be taken there.

- In the project page, the arrows look a bit out of place:

image

overview/default.nix
eljamm, line 165: We can make the page link take us directly to the project page (see https://github.com/ngi-nix/ngipkgs/pull/570#issuecomment-2713083802):

```suggestion
            <a href="${name}/index.html">${name}</a>
```

erictapen: Yeah agree, going for extensibility would be the safer bet here.

erictapen: Thanks, I'll respond to your review here. Opening the pages as `file://` in the browser doesn't work and I think it's reasonable to require contributors to use some kind of minimal webserver (e.g. I use `darkhttpd`). Opening as `file://` also brings the disadvantage of not allowing AJAX and even is inconsistently implemented in different browsers.
On a different note, I'd rather not have `index.html` fragments in URLS. These are implementation details and shouldn't be visible to the user (and they make the URL longer).

eljamm: Alright, that seems reasonable to me. Then what do you think about adding another package that runs a HTTP server on the overview? Something like `overview-preview` that users can run to see how it would look in production.

We could alternatively rename `overview` to `overview-static` and have `overview` be the package that starts the server.

erictapen: Mh how about we just add `darkhttpd` to `devShells.default` and document how to build the overview and run the server?

eljamm: Sure, would that be better for reloading changes faster?

PR #572, opened 2025-03-11
Update CONTRIBUTING.md

eljamm:
We should probably add new project/package templates somewhere, instead of linking examples. With this, the documentation links won't point to outdated formats and users can just copy that and start working.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,6 +85,7 @@ Instead, write one sentence per line, as this makes it easier to review changes.
      ```
 
      The module will then be accessible from `nixosModules.services.some-project`.
+     Note that for new projects, it's ideal that you follow the [triaging template](#triaging-an-ngi-project) workflow and create a new issue, detailing some information about this project.
 
    - Add the module tests in `test.nix`, or under a test directory, and reference that file in `default.nix`.
      For example:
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
Looks good already, what else is missing?

CONTRIBUTING.md
```diff
@@ -66,10 +66,20 @@ Instead, write one sentence per line, as this makes it easier to review changes.
 
    An existing example is [libgnunetchat](https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/by-name/libgnunetchat/package.nix).
 
-1. To add a NixOS service module, start by creating a `default.nix` file in the directory `projects/some-project` where `some-project` is the project name corresponding to the last URL component in the [NLnet project listing](https://nlnet.nl/project/).
+1. When contributing to a project, start by checking if it has an entry in `projects/some-project`.
+   If the entry does not exist, you should copy the project template and edit it with relevant details:
```
fricklerhandwerk, line 70:
```suggestion
   If the entry does not exist, copy the project template and edit it with relevant details:
```

Always use imperative in instructions. It's fewer words and unambiguous. If you write "you should" it sounds like there may be other options.

Issue #573, opened 2025-03-11
US: Project template


eljamm:
As a contributor, I want an example project template to be available, so that I can easily add a new project and have a better idea of what's expected.

## Acceptance criteria

Given contributor documentation for adding a new project,
when I copy the project template as per the instructions,
the template passes CI checks and I can modify it to fit the project information.

## Notes

This is meant to be a more structured and clear version of the example in [projects/models](https://github.com/ngi-nix/ngipkgs/blob/ff95b2c147a7f337ef2083831c493bb853974790/projects/models.nix#L108), which was designed to be more of a test for types.

PR #574, opened 2025-03-11
Add project template

eljamm:
This adds a template for projects which can easily be copied to `projects/`. It also adds the template to the flake check to make sure that it's typed correctly. 

Related: https://github.com/ngi-nix/ngipkgs/issues/573

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -224,6 +224,8 @@
                 };
                 "infra/makemake" = toplevel self.nixosConfigurations.makemake;
                 "infra/overview" = self.packages.${system}.overview;
+                # fake derivation for flake check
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -224,6 +224,8 @@
                 };
                 "infra/makemake" = toplevel self.nixosConfigurations.makemake;
                 "infra/overview" = self.packages.${system}.overview;
+                # fake derivation for flake check
+                "infra/templates" = pkgs.writeText "dummy" (lib.strings.toJSON classic'.templates.project);
```
eljamm, line 228: Flake checks require a derviation. This is the best way I found to do this so far, so I'm happy for any suggestions ...

fricklerhandwerk (approved):
LGTM

PR #581, opened 2025-03-12
overview: render subgrant names if they exist in the new project model

erictapen:
tmp xCpD4nTpRx

### DIFF
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -148,11 +148,25 @@ let
         '';
     };
 
+    subgrants = rec {
+      one = subgrant: ''
+        <li>
+          <https://nlnet.nl/project/${subgrant}>
+        </li>
+      '';
+      many =
+        subgrants:
+        optionalString (subgrants != [ ]) ''
+          <ul>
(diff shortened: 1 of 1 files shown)


overview/default.nix
```diff
@@ -148,11 +148,25 @@ let
...
+          </ul>
+        '';
+    };
+
     projects = {
       one = name: project: ''
         ${heading 1 name}
-        <https://nlnet.nl/project/${name}>
-
+        ${render.subgrants.many (project.metadata.subgrants or [ ])}
```
eljamm, line 169: Would it be feasible to fall back to using `<https://nlnet.nl/project/${name}>`  if `project.metadata.subgrants` doesn't exist? As such, we won't break the links for the old projects.

fricklerhandwerk: I think it's okay to break it temporarily (since it's a partial break) and in one of the next steps to invert the compatibility wrapper so `//projects-old` are converted to the new format.

fricklerhandwerk (approved):

Issue #583, opened 2025-03-13
Data type for binary blobs


eljamm:
Some NGI projects (Libre-SOC, Nitrokey, Heads, ...) ship binary blobs as aretefacts, but we don't currently have an effective/accurate way to represent them in our data model. What would be the best way to do this?

PR #584, opened 2025-03-13
models: add binary type

eljamm:
Closes: https://github.com/ngi-nix/ngipkgs/issues/583

### DIFF
--- a/projects/models.nix
+++ b/projects/models.nix
@@ -90,6 +90,7 @@ rec {
     metadata = optionalStruct {
       summary = option string;
       subgrants = list string;
+      links = optionalAttrs (option urlType);
     };
     nixos = struct "nixos" {
       # TODO: Tests should really only be per example, in order to clarify that we care about tested examples more than merely tests.
--- a/projects/models.nix
+++ b/projects/models.nix
@@ -39,6 +39,11 @@ let
     (attrs any)
   ];
 
+  binaryType = struct "binary" {
+    name = option string;
+    data = option (either absPath drv);
+  };
+
   # TODO: plugins are actually component *extensions* that are of component-specific type,
   #       and which compose in application-specific ways defined in the application module.
   #       we can't express that with yants, but with the module system, which gives us a bit of dependent typing.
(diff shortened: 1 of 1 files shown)


projects/models.nix
```diff
@@ -96,6 +107,7 @@ rec {
         programs = optionalAttrs (option programType);
         services = optionalAttrs (option serviceType);
       };
+      blobs = option (list blobType);
```
eljamm, line 110: According to [Wikipedia](https://en.wikipedia.org/wiki/Binary_blob):

> In the context of [free and open-source software](https://en.wikipedia.org/wiki/Free_and_open-source_software), [proprietary software](https://en.wikipedia.org/wiki/Proprietary_software) only available as a [binary executable](https://en.wikipedia.org/wiki/Executable) is referred to as a blob or binary blob.

But what are binaries coming from non-proprietary software? just binaries?

```suggestion
      binary = option (list binaryType);
```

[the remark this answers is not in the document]
fricklerhandwerk: The idea behind namespacing Nixpkgs and NixOS are the separate composition mechanisms. Binaries would have no dependencies, so they're artefacts on their own.

fricklerhandwerk: Yes, just binaries

fricklerhandwerk, line 110:
```suggestion
      blobs = optionalAttrs blobType;
```

There's no inherent ordering, it should be an attrset. Also lists don't allow access by symbolic name

projects/models.nix
```diff
@@ -78,31 +83,41 @@ let
   optionalAttrs = set: option (attrs set);
   nonEmtpyAttrs = t: restrict "non-empty attribute set" (a: a != { }) (attrs t);
   absPath = restrict "absolute path" (p: lib.pathExists p) (either path string);
+
+  dummyDerivation = derivation {
```
fricklerhandwerk, line 87: Please move to a let block in the example it self, this is confusing here.

fricklerhandwerk (approved):

PR #586, opened 2025-03-13
Re-Isearch: Add a VM test

### LINKED ISSUE #571: Re-Isearch: Test: Index and Search a document
We'd like to write a test for the [Re-Isearch module](https://github.com/ngi-nix/ngipkgs/blob/ff95b2c147a7f337ef2083831c493bb853974790/projects/Re-Isearch/default.nix#L23), which should index some test data and successfully return a search result. This should be easy to do, following the 5th step in the [re-Isearch/INSTALLATION](https://github.com/re-Isearch/re-Isearch/blob/master/INSTALLATION) document.

OPNA2608:
Closes #571

I tried to make it easy to add tests for further document formats, I hope this isn't overengineered.

I noticed that PDF indexing seemingly doesn't work. The code [looks for an external filter tool called `pdftomemo`](https://github.com/re-Isearch/re-Isearch/blob/2.20220925.4.0a/doctype/adobe_pdf.cxx#L59). On re-isearch master, there is an in-tree source checkout of `xpdf` at [`/filters/xpdf-3.01-bsn`](https://github.com/re-Isearch/re-Isearch/tree/4cf2bd99d7c3f7a782b621dd64450e3ecdde0d21/filters/xpdf-3.01-bsn) which can build this `pdftomemo`, but the package in Nixpkgs is too old to have this checkout.

Additionally, we do have an `xpdf` package in Nixpkgs… It is marked insecure due to having multiple known vulnerabilities. And that package has a version number higher than what's included in the repo… So we prolly don't want this anyway.


fricklerhandwerk:
Having that note in the code would be great so we don't lose track of that information.

fricklerhandwerk (approved):

projects/Re-Isearch/test.nix
```diff
@@ -0,0 +1,98 @@
+{
+  sources,
+  lib,
+  ...
+}:
+let
+  docDir = "documents";
+  docPath = filename: "${docDir}/${filename}";
+  docs = [
+    # PDF support not functional, needs custom pdftomemo from /filters/xpdf-3.01-bsn to be built & installed
```
OPNA2608, line 10: I have a note here that PDF support doesn't work & needs the seemingly custom `pdftomemo` from the source. Do you want me to expand this with the info that it needs a `re-isearch.src` newer than what's in Nixpkgs to appear there, and the questionable security of that project?

fricklerhandwerk: Yes, essentially what you wrote in the PR description

OPNA2608: Expanded the comment with more information.

fricklerhandwerk (approved):

PR #588, opened 2025-03-14
Update project metadata template

imincik:


templates/project/default.nix
```diff
@@ -37,11 +41,26 @@
...
-  # NOTE: same structure as programs
+  # Services
   nixos.modules.services = {
+    foobar = {
+      name = "foobar";
+      module = ./services/foobar/module.nix;
+      examples.foobar = {
+        module = ./services/foobar/examples/foobar/module.nix;
+        description = "";
+        tests.basic = ./services/foobar/examples/foobar/tests/basic.nix;
```
eljamm, line 56: You need to create the same folder structure for services as programs for the CI checks to pass as well. I've opted to keep `nixos.modules.services` empty because:
1. I don't have to make the template structure complicated, especially when projects don't have that many artefacts
2. It's more digestible as a template, at least from a beginner's perspective
3. It's easier to delete if you don't need it

eljamm:
I've resolved the merge conflict and improved the naming in followup commits, so there should be less confusion with new contributors.

fricklerhandwerk (approved):

PR #604, opened 2025-03-14
overview: simplify per-project display

fricklerhandwerk:


overview/default.nix
```diff
@@ -262,7 +262,7 @@ let
         --from=markdown+raw_html \
         --to=html \
         --standalone \
-        --css="/style.css" \
+        --css="./style.css" \
```
erictapen, line 265: This breaks css for project URLs: http://localhost:8080/project/Nitrokey-3.html

overview/default.nix
```diff
@@ -237,9 +237,9 @@ let
     }
     // mapAttrs' (
       name: project:
-      nameValuePair "project/${name}/index.html" {
```
erictapen, line 240: I really don't like these kind of URLs. They are longer and they contain the `html` bit which is an implementation detail imo and should be hidden from the user.

overview/default.nix
fricklerhandwerk: Ugh... this needs more thought. But absolute URLs bad because you can't use the styles locally without web server shenanigans as it will think it's `file:///style.css` rather than the store path.

overview/default.nix
fricklerhandwerk: The hiding can be done at the web server side with rewrites. The document structure is independent of that and should be explicit at the file system level. The extra nesting has no meaning in our document structure and thus shouldn't exist.

overview/default.nix
erictapen: I'd advise against opening stuff as `file://` in general. It breaks stuff in weird ways and doesn't even behave the same way across browsers. Instead we should put a simple web server into shell.nix, something like `darkttpd` that works out of the box with no config required.

PR #610, opened 2025-03-17
marginalia-search: Build slop from source during the build

OPNA2608:
See https://github.com/ngi-nix/ngipkgs/pull/597#issuecomment-2725197006 for details.


OPNA2608:
Actually, doing the long term fix doesn't seem that hard. I'll repurpose this PR to implement that instead.

eljamm (approved):

PR #616, opened 2025-03-18
openXC7: Import upstream packages, add project

### LINKED ISSUE #552: Implement project metadata for OpenXC7
Implement project metadata in 'projects/OpenXC7/default.nix'

OPNA2608:
Closes #552

https://github.com/openXC7/toolchain-nix contains upstream's Nix code for the toolchain, but we can't just `fetchFromGitHub` that & pass everything through as-is

- We can't do `import` or `callPackage` any of the downloaded files, because that would re-introduce IFD to the repo after it was previously removed from it treewide.
- We *could* just copy the src into `out` and make a shell script that wraps Nix to run `nix develop` inside of the downloaded checkout, but
  1. `nix` is dumb and sees `nix develop /nix/store/...` as "oh, you want me to build a derivation based on this store path… sorry, i can't parse this into a derivation!", necessitating some directory changing shenanigans
  2. It makes it impossible to test that setup in a VM, since that would require internet access for Nix to resolve URLs for downloads… unless we re-introduce IFD and finangle the VM environment to have everything pre-downloaded maybe?

So that left me with one option: Copy the relevant files from the repo into here, and replicate some setup code from the flake to build a wrapper that adds things to the environment, similar to a development shell (whose output we can't just use as-is, because the generated file in the store is commented with "this is an implementation detail, do not have any expectations for this path).

As a result, `src` is kinda useless, only serving as a pointer to what version of the repo the code was copied from.

Formatting of the copied files differs from upstream due to our pre-commit hook forcing a reformatting.

Really not happy with any of this, but given the situation described above, it is what it is…

### DIFF
--- a/pkgs/by-name/openxc7/nix/fasm/default.nix
+++ b/pkgs/by-name/openxc7/nix/fasm/default.nix
@@ -1,0 +1,54 @@
+# https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md
+
+{
+  lib,
+  buildPythonPackage,
+  fetchFromGitHub,
+  pythonOlder,
+  cmake,
+  textx,
+  cython,
+  fetchpatch,
+}:
+
+buildPythonPackage rec {
+  name = "fasm";
+  version = "0.0.2.r98.g9a73d70";
+  format = "setuptools";
+
+  disabled = pythonOlder "3.7";
+
+  src = fetchFromGitHub {
+    inherit name;
+    owner = "openxc7";
+    repo = "fasm";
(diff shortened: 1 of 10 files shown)


fricklerhandwerk:
@opna2608 this is the correct approach from what I gather after reading or participating in discussions with Nixpkgs maintainers. @imincik and @julienmalka may chip in there with 👍👎.

Nixpkgs (and by extension NGIpkgs) a distribution, and we curate how exactly we integrate the actual code, and that's expressed with the Nix language.

fricklerhandwerk:
Makes sense, thank you!

fricklerhandwerk (approved):

PR #633, opened 2025-03-21
overview: better rendering for subgrants

erictapen:
tmp zTTZQsMxIi

### DIFF
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -171,7 +171,7 @@ let
         '')
         + (optionalString (metadata ? subgrants && metadata.subgrants != [ ]) ''
           <p>
-            This project is funded through these NGI subgrants:
+            This project is funded by NLnet through these subgrants:
 
             ${render.subgrants.many metadata.subgrants}
           </p>
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -151,16 +151,14 @@ let
     subgrants = rec {
       one = subgrant: ''
         <li>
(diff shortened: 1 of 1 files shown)


eljamm (approved):
Looks nice!

overview/default.nix
```diff
@@ -171,8 +169,12 @@ let
             ${metadata.summary}
           </p>
         '')
-        + (optionalString (metadata ? subgrants) ''
-          ${render.subgrants.many metadata.subgrants}
+        + (optionalString (metadata ? subgrants && metadata.subgrants != [ ]) ''
+          <p>
+            This project is funded through these NGI grants:
```
eljamm, line 174: Just one small thing, I think this should be:

```suggestion
            This project is funded through these NGI subgrants:
```

erictapen: Is subgrant really a name that Nlnet uses? I thought that was just our terminology.

eljamm: I think they call them `projects`, but our definition of what projects are are different. Calling them `grants` in this case might be a bit confusing since that's already used to refer to Review, Entrust, ... in the project pages. Example:
image

IMO we should stick to one terminology, and preferably ours.

erictapen: Makes sense :+1:

fricklerhandwerk (dismissed):

fricklerhandwerk (approved):

PR #643, opened 2025-03-25
Improve module template

themadbit:
While exposing `mox,` I found that the current `module.nix` template structure can be improved as it has been used widely and will be in the future. So I have. @eljamm


fricklerhandwerk (approved):
Thanks, makes sense.

(At some point we'll have to watch out not to grow the template too much, otherwise it will become mostly noise. Really, templates are only any good if you have written them yourself for yourself. Upstream can maintain the interface and tutorials for learning how to use it.)

Issue #614, opened 2025-03-18
US: NGI software demo runtime environment for services


erictapen:
This is a subset of what #575 describes.

As a user of NGIpkgs running Ubuntu, I want to launch and experiment with an NGI application running as a service, in order to assess whether it is fit for purpose.

Target audience: small-business sysadmins

- proficient with the Linux command line and plain text editing, have their own tools
- will gladly run `podman` or VMs
- good enough with Git
- not familiar with Nix or functional programming
- may be security conscious
- will be reluctant to `curl | sh` and pollute their existing environment

Operating conditions:

- uses some off-the-shelf major Linux distribution: Debian, Ubuntu, Fedora, Arch
    - NGI targets FLOSS and we can expect people interested in that to show up at NGIpkgs
    - also, macOS/WSL introduces more moving parts we’d have to care about. We can
- for the purpose of this user story, pick one to run an integration test on
    - See https://github.com/numtide/nix-vm-test for NixOS VM tests on a Debian/Ubuntu/Fedora image
        - Advantage over GitHub Actions runners:
            - no re-build/re-download if nothing changed
            - full control over the source, runs on our own infrastructure
        - Disadvantages:
            - more involved setup
            - increases load on our infrastructure

Demo application:

- Pick exactly one, which is simple to configure and doesn’t incur maintenance overhead
    - Example: Galene, doesn’t need secrets or persistent state
        - Still needs to be put into NGIpkgs
    - Example: [Cryptpad](http://ngi.nixos.org/project/Cryptpad/)
        - Need it for ourselves anyway, may as well make it nice
        - Just make sure we don’t spend more time on the application than on the user story

## Acceptance criteria

Given I have picked an NGI application to experiment with,

When I follow the instructions on the project page,

Then the software will run on my machine, I can try out different settings, and remove the software without traces afterwards.

## Implementation notes

The project page should guide through roughly these steps:

1. Download an artifact with the NGIpkgs environment
    - For example an OCI image that has NixOS preconfigured to build a certain configuration
2. Launch the environment, using pre-existing tools from my own system
    - For example using podman
3. Interact with the applications from my own environment, e.g. browse the web interface
    - For example by mapping ports from the container to the host
4. Customize the NGIpkgs environment, using my own tools
    - For example by mounting my locally edited `configuration.nix` file into the container and running `nixos-rebuild` inside the container
5. Remove all artefacts

[...]

PR #658, opened 2025-03-25
Cryptpad: move to projects

erictapen:
I figured this would be beneficial, as we want to use Cryptpad as testing ground for service demos: #614


eljamm (approved):
LGTM. Thanks!

PR #661, opened 2025-03-25
Cryptpad: add example and test

eljamm:
Tested inside an lxc container generated with [nixos-generators](https://github.com/nix-community/nixos-generators):

image

**Note:** Instructions for reproducing this are available [here](https://github.com/eljamm/ngipkgs-container-generators#cryptpad-lxc).


erictapen:
I worked on something similar yesterday to show something in the project overview. I think this will not work as expected without a reverse proxy, as the Nginx config currently ties a few routes together? My solution to this was

```nix
  services.nginx = {
    virtualHosts."cryptpad.localhost" = {
      enableACME = false;
      forceSSL = false;
    };
  };
```

so that we don't depend on ACME working.

projects/Cryptpad/default.nix
```diff
@@ -23,7 +23,13 @@
     cryptpad = {
       name = "cryptpad";
       module = "${sources.inputs.nixpkgs}/nixos/modules/services/web-apps/cryptpad.nix";
-      examples.basic = null;
+      examples.basic = {
```
erictapen, line 26:
```suggestion
      examples.demo = {
```

I had this idea to reserve the `demo` name for service demos, as they would be rendered differently in the overview than other examples. What do you think?

eljamm: Yup, that's a great idea! I like it.

eljamm:
> I worked on something similar yesterday to show something in the project overview. I think this will not work as expected without a reverse proxy, as the Nginx config currently ties a few routes together? My solution to this was
> 
> ```nix
>   services.nginx = {
>     virtualHosts."cryptpad.localhost" = {
>       enableACME = false;
>       forceSSL = false;
>     };
>   };
> ```
> 
> so that we don't depend on ACME working.

The current config doesn't enable nginx for Cryptpad (even though it has an option to) and instead uses:

```
services.cryptpad.setting.httpSafeOrigin = "http://localhost:${toString servicePort}";
```

Which seems to work. Is there something that wouldn't work without this?

erictapen:
Yeah I know. I didn't check the functionality, but the Nginx config also contains the `/cryptpad_websocket` route which wouldn't be reachable in your code, as the app seems to deliver websockets separately.

https://github.com/NixOS/nixpkgs/blob/b613a5f7db5a7ad8507a2d20bec11db892193f25/nixos/modules/services/web-apps/cryptpad.nix#L266-L292

erictapen:
Oh, no I never tested it this far, sorry. Ideally we would be able give the name a proper host name like cryptpad.local, because my suspicion would be that using `localhost` messes with CORS.

erictapen (approved):
Let's merge for now and iterate on the Nginx problem later.

PR #664, opened 2025-03-26
Omnom: move from projects-old/ to projects/

OluchiTheAnalyst:


OluchiTheAnalyst:
@eljamm I was trying  to drop the flakes.lock then i mistakening force pushed and the other PR closed. so i opened another one but i am still seeing flake.lock i just dont kone how this came about

PR #665, opened 2025-03-26
meta-press: 1.8.17.1 -> 1.8.17.4

### LINKED ISSUE #652: Update `meta-press`
## Instructions

1. Open the [package derivation](https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/by-name/meta-press/package.nix) in your text editor

```nix
  |
  v
sha256 = "";
```

5. Build the package

```
```

```
```

3pleX-dev:
This should resolve #652.
@eljamm, please review.
Note: I only verified that the extension (firefox_addon.xpi) compiled correctly. There was no executable in the result folder.

### DIFF
--- a/pkgs/by-name/meta-press/package.nix
+++ b/pkgs/by-name/meta-press/package.nix
@@ -6,12 +6,12 @@
 }:
 stdenv.mkDerivation (finalAttrs: {
   pname = "meta-press";
-  version = "1.8.17.1";
+  version = "1.8.17.4";
 
   src = fetchgit {
     url = "https://framagit.org/Siltaar/meta-press-ext.git";
-    rev = "v${finalAttrs.version}";
-    hash = "sha256-5/TotS+Je4AlKVPq1H4BVEc2gkJ8NSmDuKJzRN7qR1M=";
+    tag = "v${finalAttrs.version}";
+    hash = "sha256-sl84E38swbzdlAd6zfm/wwYgzmMwSttri2Gi8Ljcn/k=";
   };
 
   # The Makefile moves the output to the enclosing folder


pkgs/by-name/meta-press/package.nix
```diff
@@ -6,12 +6,12 @@
 }:
 stdenv.mkDerivation (finalAttrs: {
   pname = "meta-press";
-  version = "1.8.17.1";
+  version = "1.8.17.4";
 
   src = fetchgit {
     url = "https://framagit.org/Siltaar/meta-press-ext.git";
     rev = "v${finalAttrs.version}";
```
eljamm, line 13: This was introduced a while ago to avoid conflicts with branches that share the same name as the tag. Although this package's source repo doesn't have such branches, IMO it's a good pattern to adopt when working with tags.

```suggestion
    tag = "v${finalAttrs.version}";
```

eljamm (commented):
Congrats on your first contribution! Would you please rename the commit to fit the [Nixpkgs commit conventions](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#commit-conventions)?

pkgs/by-name/meta-press/package.nix
eljamm: You can do `git commit --amend` and it will include the changes to the latest commit. Then you can force-push the branch. Alternatively we can rebase, but amending is simpler.

3pleX-dev: I just did that and it worked. Thank you for the review @eljamm ❤️

eljamm (approved):
Just changed the title to fit the conventions.

LGTM, thanks!

Issue #651, opened 2025-03-25
Update `corestore`


eljamm:
## Instructions
 
1. Open the [package derivation](https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/by-name/corestore/dream2.nix) in your text editor
1. Go to the homepage (`meta.homepage`) or source repository (`src`) of the package  
2. Replace the version in the derivation with the most up-to date tag or commit revision if tags aren't available
3. Replace hashes with empty strings. Example:

```nix
sha256 = "sha256-18FKwP0XHoq/F8oF8BCLlul/Xb30sd0iOWuiKkzpPLI=";
  |
  v
sha256 = "";
```

4. Build the package

```
nix build .#checks.x86_64-linux.packages/<package_name>
```

5. The build will fail because the hashes are empty, but it will return the correct hash. Replace the empty hash with the correct hash and build again. Example:

```
error: hash mismatch in fixed-output derivation '/nix/store/xxkj74gnza5rw5xyawzvlafbvbb76qdq-source.drv':
        likely URL: https://github.com/holepunchto/corestore/archive/v7.0.23.tar.gz
         specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
            got:    sha256-oAsyv10BcmInvlZMzc/vJEJT9r+q/Rosm19EyblIDCM=
``` 

6. Verify that it works, if possible (at least with `--version`)

themadbit:
I am working on this. The bump is to a significantly higher version, so it's taking quite some time as many dependencies are needed.

eljamm:
It seems we'll also need to update the lockfile as well, which I think needs to be done manually. Would it be possible to just use [buildNpmPackage](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/javascript.section.md#buildnpmpackage-javascript-buildnpmpackage), instead of dream2nix?

themadbit:
I've updated the lock file too. Check PR commits?
Let me look into using `buildNpmPackage`.

eljamm:
Yup, the update looks good. Well done! You just need to squash the commits and I also left a comment about the node version.

PR #666, opened 2025-03-26
corestore: 6.15.9 -> 7.0.23

themadbit:
Resolves [issue](https://github.com/ngi-nix/ngipkgs/issues/651).

### DIFF
--- a/pkgs/by-name/corestore/dream2.nix
+++ b/pkgs/by-name/corestore/dream2.nix
@@ -6,7 +6,7 @@
 }:
 rec {
   name = "corestore-${version}";
-  version = "6.15.9";
+  version = "7.0.23";
 
   imports = [
     dream2nix.modules.dream2nix.nodejs-package-lock-v3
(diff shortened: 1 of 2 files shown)


pkgs/by-name/corestore/dream2.nix
```diff
@@ -41,6 +41,7 @@ rec {
         stdenv
         ;
 
+      # Update to Node.js 22
       npm = nixpkgs.nodejs_16.pkgs.npm;
```
eljamm, line 45: Is this a note or did you forget to do this?

themadbit: Yes, I was experimenting with nodejs_22.
It shouldn't be there in squashed commit(learning how to :sweat_smile: ). As I look into using npmBuildPackage, I might update it.

eljamm: That's certainly a sound approach, as it's recommended to use the same node version as the upstream package when possible (see [docs](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/javascript.section.md#general-principles-javascript-general-principles)). That said, you don't need to look into rewriting this with `npmBuildPackage` right now. Let's just focus on merge this, first.

eljamm (approved):
LGTM, thanks!

PR #667, opened 2025-03-26
Omnom: Move to projects

OluchiTheAnalyst:


OluchiTheAnalyst:
Hi @eljamm i have restored this PR but i can't tell the root of this conflict
could you please have a look

eljamm:
It's because the flake update commit was merged with one to move the project. I'll resolve that real quick and we'll just focus on reviewing the migration.

projects/Omnom/example.nix
```diff
@@ -0,0 +1,24 @@
...
+      port = 8080;
+
+      settings = {
+         app = {
+           disable_signup = true; #restrict CLI user-creation
+           results_per_page = 50;
+         };
+        smtp = {
+          tls = true;
+          host = "127.0.0.1";
```
eljamm, line 16: There should be a port option here, why is it gone from the example?

eljamm (commented):
Thank you for your effort on this @OluchiTheAnalyst. After addressing my review, please format this with `nix fmt`.

OluchiTheAnalyst:
Hi @eljamm Thank you for your feedback on the PR. I’ve implemented the requested changes,could you please review the changes and let me know if any further adjustments are needed?

projects/Omnom/default.nix
```diff
@@ -0,0 +1,28 @@
...
+    summary ="Omnom is a webpage bookmarking and snapshotting service.";
+    subgrants = [
+      "Omnom"
+      "Omnom-ActivityPub"
+    ];
+  };
+  nixos.modules.service = {
+    Omnom = {
+      name = "Omnom";
+      module = "${sources.inputs.nixpkgs}/nixos/modules/services/misc/omnom.nix};
```
eljamm, line 17:
```suggestion
      module = "${sources.inputs.nixpkgs}/nixos/modules/services/misc/omnom.nix";
```

projects/Omnom/default.nix
```diff
@@ -0,0 +1,28 @@
...
+      examples.base = null;
+      };     
+        description = "Basic Omnom configuration,mainly used for testing purposes.";
+        };
+        test = null;
+
+        };
+      };
```
eljamm, line 25:
```suggestion
```

eljamm (commented):
Please make sure that the example is building locally, first, with:

```nix
nix build .#checks.x86_64-linux.projects/Omnom/nixos/examples/base
```

If it succeeds, you'll know that you're on the right path, else it will give you some useful information on what to fix.

projects/Omnom/default.nix
```diff
@@ -0,0 +1,26 @@
...
+
+  nixos.modules.services = {
+    Omnom = {
+     # https://github.com/asciimoo/omnom/blob/master/config/config.go
+      module = "${sources.inputs.nixpkgs}/nixos/modules/services/misc/omnom.nix";
+      examples.base = {
+        path = ./example.nix;
+        description = "Basic Omnom configuration mainly used for testing purposes.";
+      };
+    }
```
fricklerhandwerk, line 24:
```suggestion
    };
```

projects/Omnom/example.nix
```diff
@@ -0,0 +1,18 @@
+{ ... }:
+{
+  service.Omnom.enable = true;
+    openFirewall = true;
+  }
```
fricklerhandwerk, line 5:
```suggestion
```

projects/Omnom/example.nix
```diff
@@ -0,0 +1,18 @@
+{ ... }:
+{
+  service.Omnom.enable = true;
```
fricklerhandwerk, line 3:
```suggestion
  services.omnom = {
    enable = true;
```

projects/Omnom/example.nix
```diff
@@ -0,0 +1,18 @@
...
+        disable_signup = true; 
+        results_per_page = 50;
+      };
+      smtp = {
+        tls = true;
+        host = "127.0.0.1";
+        port = 1025;
+        username = "testUser";
+      };
+    };
```
fricklerhandwerk, line 18:
```suggestion
    };
  };
}
```

projects/Omnom/default.nix
```diff
@@ -0,0 +1,26 @@
...
+  metadata = {
+    summary = "Omnom is a webpage bookmarking and snapshotting service.";
+    subgrants = [
+      "Omnom"
+      "Omnom-ActivityPub"
+    ];
+  };
+
+  nixos.modules.services = {
+    Omnom = {
```
fricklerhandwerk, line 17:
```suggestion
    omnom = {
```

While we're really flexible with calling our attributes whatever we like, and while it surely makes sense to use exactly the project name, the NixOS convention is to keep it in lower case (which also happens to be ever so slightly simpler to type), so I suggest we follow that to avoid surprise.

OluchiTheAnalyst:
Hello @eljamm  Finally my examples are running, I've made necessary changes, this PR is ready to reviewed again.
thank you so much

eljamm (approved):
Just dropped the flake lock stuff that was causing the merge conflict, but this should be good. Thanks!

eljamm:
This also needed a format with `nix fmt` for the CI to pass.

OluchiTheAnalyst:
> This also needed a format with `nix fmt` for the CI to pass.

Alright

eljamm (approved):

PR #670, opened 2025-03-26
openfire: 4.9.0 -> 4.9.2

themadbit:
Resolves [issue](https://github.com/ngi-nix/ngipkgs/issues/653)
I avoided updating to `v5.0.0,` which is in alpha. It's better to bump to a stable release.

### DIFF
--- a/pkgs/by-name/openfire/package.nix
+++ b/pkgs/by-name/openfire/package.nix
@@ -7,13 +7,13 @@
 }:
 maven.buildMavenPackage rec {
   pname = "openfire";
-  version = "4.9.0";
+  version = "4.9.2";
 
   src = fetchFromGitHub {
     owner = "igniterealtime";
     repo = "Openfire";
-    rev = "v${version}";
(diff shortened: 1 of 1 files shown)


pkgs/by-name/openfire/package.nix
```diff
@@ -7,13 +7,13 @@
 }:
 maven.buildMavenPackage rec {
   pname = "openfire";
-  version = "4.9.0";
+  version = "4.9.2";
 
   src = fetchFromGitHub {
     owner = "igniterealtime";
     repo = "Openfire";
     rev = "v${version}";
```
eljamm, line 15: This was introduced a while ago to avoid conflicts with branches that share the same name as the tag. Although this package's source repo doesn't have such branches, IMO it's a good pattern to adopt when working with tags.

```suggestion
    tag = "v${version}";
```

themadbit: Yes. It's the better pattern.

eljamm (approved):
Very nice!

PR #671, opened 2025-03-26
taldir: 0-unstable-2024-02-18 -> 1.0.3

### LINKED ISSUE #655: Update `taldir`
## Instructions

1. Open the [package derivation](https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/by-name/taldir/package.nix) in your text editor

```nix
  |
  v
sha256 = "";
```

5. Build the package

```
```

```
```

3pleX-dev:
Should resolve #655 
@eljamm Please help review


pkgs/by-name/taldir/package.nix
```diff
@@ -6,19 +6,19 @@
 }:
 buildGoModule {
   pname = "taldir";
-  version = "0-unstable-2024-02-18";
+  version = "0-unstable-2025-01-25";
```
eljamm, line 9: Since we're using a stable release with the tag, this should correspond to that version (see [Nixpkgs docs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#versioning)):

```suggestion
  version = "1.0.3";
```

pkgs/by-name/taldir/package.nix
```diff
@@ -6,19 +6,19 @@
...
 
   src = fetchgit {
     url = "https://git.taler.net/taldir.git";
-    rev = "9c5230f64d16d46c000c6d4f5842170c51b697ce";
-    hash = "sha256-Ri0R7kxP/FitBRKtrM48Cbks63mqpXsRc3r98M0sMus=";
+    tag = "v1.0.3";
```
eljamm, line 13: We can reuse the version attribute here since it nicely corresponds to the upstream tag, unlike the previous unstable version:

```suggestion
    tag = "v{finalAttrs.version}";
```

What's this `finalAttrs` you say? It's a way for us to reuse attributes from within the derivation (see my comment above).

pkgs/by-name/taldir/package.nix
```diff
@@ -6,19 +6,19 @@
 }:
 buildGoModule {
```
eljamm, line 7: Here, we can use `finalAttrs` to refer to attributes from within the derivation (e.g finalAttrs.version). We can also use `rec`, which will allow us to access `version` directly, but that comes with its [own issues](https://nix.dev/guides/best-practices#recursive-attribute-set-rec), so it's best to use `finalAttrs` when we can.

```suggestion
buildGoModule (finalAttrs: {
```

Note: `finalAttrs` is also better for overriding the package, but maybe that's a topic for another time 😄

pkgs/by-name/taldir/package.nix
```diff
@@ -6,19 +6,19 @@
...
   src = fetchgit {
     url = "https://git.taler.net/taldir.git";
-    rev = "9c5230f64d16d46c000c6d4f5842170c51b697ce";
-    hash = "sha256-Ri0R7kxP/FitBRKtrM48Cbks63mqpXsRc3r98M0sMus=";
+    tag = "v1.0.3";
+    hash = "sha256-axVQ687cGvxGEKECh3HmbTAFwI/YaCtPvtRU5bWsYYI=";
   };
 
   postPatch = ''
     cp ${./go.sum} go.sum
```
eljamm, line 18: Can we perhaps add a comment here for how to generate this file again?

eljamm (commented):
Really nice. Please keep up the good work.

pkgs/by-name/taldir/package.nix
3pleX-dev: Thank you for the feedback. The issues and your feedbacks are helping me learn more about nix as I go 😅

pkgs/by-name/taldir/package.nix
```diff
@@ -4,21 +4,45 @@
...
+
+  5. Copy the new go.sum file into the Nix package directory.
+
+  6. Retry the Nix build to get the correct vendorHash:
+      nix build .#packages.x86_64-linux.taldir
+
+  This ensures that Nix can properly resolve Go dependencies and avoids hash mismatch errors.
+*/
+
+  vendorHash = "sha256-V/UzUS3eMKnEhgaipsHoodAlZxuEkZM/ALliU1TNuYg=";
```
eljamm, line 45: Could you move `vendorHash` to be positioned right after `src`? It's nice to keep them close so it's easier for you to update them.

eljamm:
One more change and I think this is good to merge. The commit/PR titles should correspond exactly to the `version` attribute change, which doesn't include the `v` (that's for the tag name):

```diff
- taldir: 0-unstable-2024-02-18 -> v1.0.3
+ taldir: 0-unstable-2024-02-18 -> 1.0.3

eljamm (approved):
Awesome, well done!

Issue #601, opened 2025-03-14
Triage data for OWASP


eljamm:
Following the instructions for [triaging an NGI project](https://github.com/ngi-nix/ngipkgs/blob/main/CONTRIBUTING.md#triaging-an-ngi-project), we'd like to collect some relevant information about this project so we could expose its packaging state in our repository.

khabdrick:
Hello there, I am part of the students participating in the outreach program. Can I take this up ?

eljamm:
Yup, absolutely.

eljamm:
Ideally, we shouldn't duplicate our efforts when someone has already picked something up, but this issue didn't see any activity for a week, now. What's your progress @khabdrick? Are you still interested in this or can @Prinuel pick it up?

**PS:** It's not worth it to spend more than 1H for the triaging task. If a project seems too difficult to triage, it might better to move on to something easier, while also leaving a comment on what the difficulties are so other contributors can either provide assistance and help get it over the finish line or gain more insight if they continue the work, themselves.

khabdrick:
@eljamm I understand. Let me have one more day.
By this time tomorrow if I don't have anything @Prinuel can take it up.

khabdrick:
I just wanted to let you know that I'm done with this. @eljamm 
Couldn't find some information, though. 🤔

eljamm:
Thanks. It's fine if we don't collect everything as we realistically won't be able to, anyways. That's why it's best to limit the task to 1H maximum, because any more than that and we'll probably be wasting time.

eljamm:
Done in https://github.com/ngi-nix/ngipkgs/issues/660

Prinuel:
oh just seeing this, I already  submitted though, yeah i will pick up something else to work on. Thank you

PR #672, opened 2025-03-26
Add OWASP triage report (Issue #601)

Prinuel:
This is an initial and packaging feasibility report of OWASP tools in Nixpkgs. 
Here's the summary:
I added a triage report for OWASP tools in Nixpkgs and then identified if there are any available OWASP tools and which needs packaging. Afterwards I documented build requirements and dependency management and then rounded it off by highlighting licensing considerations and potential challenges that could be faced. 
Feedbacks and suggestions are welcome!

### DIFF
--- a/overview/owasp-triage.md
+++ b/overview/owasp-triage.md
@@ -1,0 +1,108 @@
+# OWASP Triage Report for NGIpkgs
+# 1. Brief Introduction
+The Open Web Application Security Project, recently renamed to the Open Worldwide Application Security Project [OWASP](https://owasp.org/about/) is a non-profit organization that works to improve the security of software. 
+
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Hi @Prinuel, please use the triage issue template to submit this information, in the format requested. This is a decidedly manual task because projects are very different and require careful attention to details that are both project- and Nix-specific. So far, unfortunately this can't be helped much with LLMs since their output is too noisy and can't be trusted or easily verified, otherwise we'd be mostly done already.

Prinuel:
Ok  @fricklerhandwerk, I'll check out the format in the "triage issue template" and format this correctly also with other issues I'm working on. Thank you

PR #674, opened 2025-03-27
change libgnunetchat package version

### LINKED ISSUE #657: Update `libgnunetchat`
## Instructions

1. Open the [package derivation](https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/by-name/libgnunetchat/package.nix) in your text editor

```nix
  |
  v
sha256 = "";
```

5. Build the package

```
```

```
```

khabdrick:
Fixed #657 
Updated `libgnunetchat` version.

### DIFF
--- a/pkgs/by-name/libgnunetchat/package.nix
+++ b/pkgs/by-name/libgnunetchat/package.nix
@@ -18,7 +18,7 @@
 # From https://github.com/ngi-nix/gnunet-messenger-cli/blob/main/flake.nix
 stdenv.mkDerivation (finalAttrs: {
   name = "libgnunetchat";
-  version = "0.5.0-unstable-2025-01-07";
+  version = "0.5.3";
 
   src = fetchgit {
     url = "https://git.gnunet.org/libgnunetchat";


pkgs/by-name/libgnunetchat/package.nix
```diff
@@ -18,7 +18,7 @@
 # From https://github.com/ngi-nix/gnunet-messenger-cli/blob/main/flake.nix
 stdenv.mkDerivation (finalAttrs: {
   name = "libgnunetchat";
-  version = "0.5.0-unstable-2025-01-07";
+  version = "0.5.3";
 
   src = fetchgit {
```
eljamm, line 23: Changing the version is not enough here. We also need to update the hashes as well, replacing the commit rev with the tag since we're targeting a stable release.

eljamm:
The gnunet PR has been merged in nixpkgs, so we can continue working on this. @khabdrick, can you address the review above?

eljamm:
Done in https://github.com/ngi-nix/ngipkgs/pull/694/commits/dff53448e716aa566f3a8af5ae23af9b06a5ede7

Issue #578, opened 2025-03-12
Package Mox as a NixOS service


wamirez:
- [static configuration file](https://www.xmox.nl/config/#hdr-mox-conf)
- [dynamic configuration file](https://www.xmox.nl/config/#hdr-domains-conf)

eljamm:
The config file seems to be using the [sconf](https://pkg.go.dev/github.com/mjl-/sconf) format. Is there a way to handle that in nixpkgs or do we need to that ourselves?

eljamm:
> create the conf file, source it in the module, and use home-manager to place them

Reading a pre-generated config file should work for now, even without home-manager, so you should definitely give your idea a go. That said, it would be ideal to be able to configure it directly with Nix in the future.

> can we raise a feature enhancement issue for sconf to be included in supported formats

Sure, it doesn't hurt to open an issue for this in nixpkgs, but I think that whether someone will work on it or not depends on how much the format is needed/used and how easy it is to implement.

Issue #577, opened 2025-03-11
Expose Mox


wamirez:
We'd like to [add the project](https://github.com/ngi-nix/ngipkgs/blob/main/CONTRIBUTING.md#how-to-create-pull-requests-to-ngipkgs) for Mox in our repository from its [triaged data](https://github.com/ngi-nix/ngipkgs/issues/576), using our [project template](https://github.com/ngi-nix/ngipkgs/tree/main/templates/project).

themadbit:
I'm a first time user and contibutor of Nix(OS/pkgs) so feel free to yell at me if I ask not so great questions. 😅 @eljamm 
Anyway, I've already gone through the contibuting doc for the ngi-nixpkgs and created a project for Mox. (FYI, i have gained some initial fundamental knowledge of working through nixpkg and have a maintainer approved PR waiting merge over at nixpkgs).

Q: Should I add the mox package to ngi-nixpkgs before exposing it as a program?

eljamm:
We mainly expose projects in order to tack how far we are in packaging them. As such, if something is missing, we first need to mark it as such, then we can start working on implementing it later on. So, although we don't necessarily need to package anything to expose Mox, it would be quite a nice bonus to have Mox packaged as well!

Also, please don't hesitate to ask anything no matter how silly it may sound. We're all here to learn after all 😁

PR #677, opened 2025-03-27
Package Mox as a service

### LINKED ISSUE #576: NGI Project: Mox
### Short Summary

Modern full-featured open source secure mail server for low-maintenance self-hosted email.

### NLnet page(s)

- https://nlnet.nl/project/Mox/
- https://nlnet.nl/project/Mox-Automation/

### Website

- https://www.xmox.nl/

### Source repository

- https://github.com/mjl-/mox

### Metadata

- Language/Framework: Go
- Dependency management: [`go.mod`](https://github.com/mjl-/mox/blob/main/go.mod)
- Development environment: N/A

### Documentation

- Usage Examples:
  - https://www.xmox.nl/commands/#hdr-usage
- Build from source/Development:
  - https://github.com/mjl-/mox/tree/main?tab=readme-ov-file#compiling

### Artefacts

_No response_

### Nixpkgs/NixOS

   - Packages:
     - [mox](https://github.com/NixOS/nixpkgs/blob/cafa38037f8053b843664a4ed969c59f3425a49f/pkgs/by-name/mo/mox/package.nix#L8)
   - Services:
     - N/A

### Extra Information

_No response_

themadbit:
Closes #576 #577 #578

- Exposed Mox as a service to resolve [issue(s)](https://github.com/ngi-nix/ngipkgs/issues/576).
- Initial tests are running, and a local rebuild of NixOS with the module included runs just fine.
i.e:

image

However, it fails to start/survive reboot with:

```
[themadbit@nixos:~]$ systemctl status mox
○ mox.service
     Loaded: loaded (/etc/systemd/system/mox.service; enabled; preset: ignored)
     Active: inactive (dead)

Mar 27 14:40:30 nixos systemd[1]: Dependency failed for mox.service.
Mar 27 14:40:30 nixos systemd[1]: mox.service: Job mox.service/start failed with result 'dependency'.
```
So, I'll leave this as a draft PR as I investigate further. Putting it out here in case you've had a similar headache  and know the painkiller :sweat_smile:

### DIFF
--- a/projects/Mox/default.nix
+++ b/projects/Mox/default.nix
@@ -1,0 +1,50 @@
+{
+  lib,
+  pkgs,
+  sources,
+}@args:
+
+{
+  metadata = {
+    summary = "Modern full-featured open source secure mail server";
+    subgrants = [
+      "Mox"
+      "Mox-API"
+      "Mox-Automation"
+    ];
+    links = {
+      install = {
+        text = "Mox Install Documentation";
+        url = "https://www.xmox.nl/install/";
+      };
+      source = {
+        text = "Mox Github Repository";
+        url = "https://github.com/mjl-/mox";
+      };
+    };
+  };
+
+  nixos.modules.programs = {
+    mox = {
+      name = "mox";
+      module = ./programs/mox/module.nix;
+      examples.basic = {
+        module = ./programs/mox/examples/basic.nix;
+        description = "Use Mox subcommands to manage/debug the Mox server";
+        tests.basic = import ./programs/mox/tests/basic.nix args;
+      };
+    };
+  };
+
+  nixos.modules.services = {
+    mox = {
+      name = "mox";
+      module = ./services/mox/module.nix;
+      examples.basic = {
+        module = ./services/mox/examples/basic.nix;
+        description = "Mox server with optional hostname and user";
+        tests.basic = import ./services/mox/tests/basic.nix args;
+      };
+    };
+  };
+}
--- a/projects/Mox/programs/mox/tests/basic.nix
+++ b/projects/Mox/programs/mox/tests/basic.nix
@@ -1,0 +1,29 @@
+{
+  sources,
+  ...
+}:
+
+{
+  name = "Mox";
+
+  nodes = {
+    machine =
+      { ... }:
+      {
+        imports = [
+          sources.modules.ngipkgs
+          sources.modules.programs.mox
+          sources.examples.Mox.basic
+        ];
+      };
+  };
+
+  testScript =
+    { nodes, ... }:
+    ''
+      start_all()
+
+      # List available SMTP config examples
+      machine.succeed("mox config example")
+    '';
+}
--- a/projects/Mox/services/mox/module.nix
+++ b/projects/Mox/services/mox/module.nix
@@ -13,6 +13,7 @@ in
       # NOTE: The module uses contextual generated config files for the mox server.
       #       This is currently the most reproducible way to get mox running. However, it might
       #       be possible to use a declarative approach when the sconf config file is supported.
+      #       If addition configuration is needed, please edit the file and restert the service.
 
       enable = lib.mkEnableOption "Mox";
       package = lib.mkPackageOption pkgs "mox" { };
@@ -35,6 +36,21 @@ in
         type = lib.types.str;
         description = "*Required* Email user as (user@domain) to be created.";
       };
+      openPorts = lib.mkOption {
+        type = lib.types.listOf lib.types.int;
+        default = [
+          25
+          80
+          143
+          443
+        ];
+        description = "Ports to be opened for the Mox Mail Server";
+      };
+      openFirewall = lib.mkOption {
+        type = lib.types.bool;
+        default = true;
+        description = "Open firewall for Mox Mail Server";
+      };
     };
   };
 
@@ -42,6 +58,12 @@ in
     environment.systemPackages = [
       cfg.package
     ];
+
+    networking.firewall = lib.mkIf cfg.openFirewall {
+      enable = true;
+      allowedTCPPorts = cfg.openPorts;
+      allowedUDPPorts = [ 53 ];
+    };
 
     users.users.mox = {
       isSystemUser = true;
--- a/projects/Mox/services/mox/module.nix
+++ b/projects/Mox/services/mox/module.nix
@@ -6,6 +6,15 @@
 }:
 let
   cfg = config.services.mox;
+  mkOptionalPort =
+    name:
+    lib.mkOption {
+      description = ''
+        The ${name} port. Set to null if we should leave it unset.
+      '';
+      type = lib.types.nullOr lib.types.port;
+      default = null;
+    };
 in
 {
   options = {
@@ -13,9 +22,9 @@ in
       # NOTE: The module uses contextual generated config files for the mox server.
       #       This is currently the most reproducible way to get mox running. However, it might
       #       be possible to use a declarative approach when the sconf config file is supported.
-      #       If addition configuration is needed, please edit the file and restert the service.
+      #       If addition configuration is needed, please edit the file and restart the service.
 
-      enable = lib.mkEnableOption "Mox";
+      enable = lib.mkEnableOption "Mox server";
       package = lib.mkPackageOption pkgs "mox" { };
       configFile = lib.mkOption {
         type = lib.types.str;
@@ -36,20 +45,15 @@ in
         type = lib.types.str;
         description = "*Required* Email user as (user@domain) to be created.";
       };
-      openPorts = lib.mkOption {
-        type = lib.types.listOf lib.types.int;
-        default = [
-          25
-          80
-          143
-          443
-        ];
-        description = "Ports to be opened for the Mox Mail Server";
+      ports = {
+        http = mkOptionalPort "http";
+        https = mkOptionalPort "https";
+        smtp = mkOptionalPort "smtp";
       };
       openFirewall = lib.mkOption {
         type = lib.types.bool;
         default = true;
-        description = "Open firewall for Mox Mail Server";
+        description = "Open firewall for the ports defined in `ports`";
       };
     };
   };
@@ -61,7 +65,13 @@ in
 
     networking.firewall = lib.mkIf cfg.openFirewall {
       enable = true;
-      allowedTCPPorts = cfg.openPorts;
+      allowedTCPPorts =
+        with lib;
+        (
+          optional (cfg.ports.http != null) cfg.ports.http
+          ++ optional (cfg.ports.https != null) cfg.ports.https
+          ++ optional (cfg.ports.smtp != null) cfg.ports.smtp
+        );
       allowedUDPPorts = [ 53 ];
     };
 
--- a/projects/Mox/services/mox/module.nix
+++ b/projects/Mox/services/mox/module.nix
@@ -1,0 +1,87 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.mox;
+in
+{
+  options = {
+    services.mox = {
+      # NOTE: The module uses contextual generated config files for the mox server.
+      #       This is currently the most reproducible way to get mox running. However, it might
+      #       be possible to use a declarative approach when the sconf config file is supported.
+
+      enable = lib.mkEnableOption "Mox";
+      package = lib.mkPackageOption pkgs "mox" { };
+      configFile = lib.mkOption {
+        type = lib.types.str;
+        default = "/var/lib/mox/config/mox.conf";
+        description = ''
+          Mox `quickstart` generates configuration files to quickly set up a mox instance.
+          This, is by far the easiest and most reproducible way to get mox running. All output from `quickstart`
+          is written to `quickstart.log`, including initial admin accounts and passwords. In this module, options
+          hostname and user are passed to `quickstart` as arguments and the config file are generated in that context.
+        '';
+      };
+      hostname = lib.mkOption {
+        type = lib.types.str;
+        default = "mail";
+        description = "Hostname for the Mox Mail Server";
+      };
+      user = lib.mkOption {
+        type = lib.types.str;
+        description = "*Required* Email user as (user@domain) to be created.";
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [
+      cfg.package
+    ];
+
+    users.users.mox = {
+      isSystemUser = true;
+      name = "mox";
+      group = "mox";
+      home = "/var/lib/mox";
+      createHome = true;
+      description = "Mox Mail Server User";
+    };
+    users.groups.mox = { };
+
+    systemd.services.mox-setup = {
+      description = "Mox Setup";
+      wantedBy = [ "multi-user.target" ];
+      requires = [
+        "network-online.target"
+      ];
+      after = [
+        "network-online.target"
+      ];
+      before = [ "mox.service" ];
+      serviceConfig = {
+        WorkingDirectory = "/var/lib/mox";
+        Type = "oneshot";
+        RemainAfterExit = true;
+        User = "mox";
+        Group = "mox";
+        ExecStart = "${pkgs.mox}/bin/mox quickstart -hostname ${config.services.mox.hostname} ${config.services.mox.user}";
+      };
+    };
+
+    systemd.services.mox = {
+      wantedBy = [ "multi-user.target" ];
+      after = [ "mox-setup.service" ];
+      requires = [ "mox-setup.service" ];
+      serviceConfig = {
+        WorkingDirectory = "/var/lib/mox";
+        ExecStart = "${pkgs.mox}/bin/mox -config /var/lib/mox/config/mox.conf serve";
+        Restart = "always";
+      };
+    };
+  };
+}
--- a/projects/Mox/services/mox/tests/basic.nix
+++ b/projects/Mox/services/mox/tests/basic.nix
@@ -24,13 +24,6 @@
           nameserver 127.0.0.1
         '';
 
-        networking.firewall.allowedTCPPorts = [
-          25 # SMTP
-          80 # HTTP
-          143 # IMAP
-          443 # HTTPS
-        ];
-        networking.firewall.allowedUDPPorts = [ 53 ];
         networking.nameservers = [ "127.0.0.1" ];
         networking.hosts = {
           "127.0.0.1" = [
--- a/projects/Mox/services/mox/tests/basic.nix
+++ b/projects/Mox/services/mox/tests/basic.nix
@@ -13,7 +13,7 @@
         imports = [
           sources.modules.ngipkgs
           sources.modules.services.mox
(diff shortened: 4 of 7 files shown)


themadbit:
Update: I've run the test interactively for debugging and got access to the VM. The VM might not have a running nameserver, and I'd like to know if there's a way I can get one running as part of the machine configs before running the test(s). @eljamm, any pointers for further debugging?

ss of log:
image

eljamm:
Turns out that it was just a matter of modifying the service order since the dns configuration is not over by the time the `mox-setup` service tries to start. With the following patch, the setup completes, the server runs and the NixOS Test succeeds:

<details><summary>fix-mox-service-order.patch</summary>
<p>

```diff
---
 projects/Mox/module.nix | 9 ++++++---
 1 file changed, 6 insertions(+), 3 deletions(-)

diff --git a/projects/Mox/module.nix b/projects/Mox/module.nix
index 43bb8d7..91d38d7 100644
--- a/projects/Mox/module.nix
+++ b/projects/Mox/module.nix
@@ -48,23 +48,26 @@
     systemd.services.mox-setup = {
       description = "Setup Mox Mail Server";
       wantedBy = [ "multi-user.target" ];
+      requires = [ "network-online.target" ];
+      after = [ "network-online.target" ];
       before = [ "mox.service" ];
       serviceConfig = {
         Type = "oneshot";
         RemainAfterExit = true;
+        User = "mox";
+        Group = "mox";
       };
       script = ''
         mkdir -p /var/lib/mox
         cd /var/lib/mox
         ${pkgs.mox}/bin/mox quickstart -hostname ${config.services.mox.hostname} ${config.services.mox.user}
-        chown -R mox:mox /var/lib/mox
       '';
     };

     systemd.services.mox = {
       wantedBy = [ "multi-user.target" ];
-      after = [ "network.target" "mox-setup.service" ];
-      requires = [ "mox-setup.service" ]; # This ensures mox-setup must succeed
+      after = [ "mox-setup.service" ];
+      requires = [ "mox-setup.service" ];
       serviceConfig = {
         WorkingDirectory = "/var/lib/mox";
         ExecStart = "${pkgs.mox}/bin/mox -config /var/lib/mox/config/mox.conf serve";
--
2.47.2
```

</p>
</details>

eljamm:
> I added services.dnsmasq and enabled dnssec and had the mox-setup unit wait for that to be up and running.

That sounds like it could be useful for a separate example config that uses DNSSEC, for those interested in that

themadbit:
> That sounds like it could be useful for a separate example config that uses DNSSEC, for those interested in that

Yess. I've stashed changes using `dnsmasq` to have a running service first, and then I can build upon it slowly. Ultimately a user should be able to specify all their desired config(s) so which means supporting sconf. Nonetheless, the current implementation, if successfully built, a user with a working and configured domain can have `Mox` running and setup working emails.

projects/Mox/test.nix
```diff
@@ -0,0 +1,61 @@
...
+  name = "mox";
+
+  nodes = {
+    machine =
+      { ... }:
+      rec {
+        imports = [
+          sources.modules.ngipkgs
+          # Sources module path depends on your folder structure
+          ./module.nix
```
eljamm, line 16: We can just import the module and the example:

```suggestion
          sources.modules.ngipkgs
          sources.modules.services.mox
          sources.examples.Mox.mox
```

projects/Mox/test.nix
```diff
@@ -0,0 +1,61 @@
...
+        networking.firewall.allowedTCPPorts = [ 25 80 143 443 587 993 ];
+        networking.firewall.allowedUDPPorts = [ 53 ];
+
+        # Configure the mox service
+        services.mox = {
+          enable = true;
+          hostname = "mail.example.com";
+          user = "admin@example.com";
+        };
+      };
```
eljamm, line 29: The test is meant to the example, not write another one

```suggestion
```

projects/Mox/example.nix
```diff
@@ -0,0 +1,6 @@
+{ ... }:
+{
+  services.mox.enable = true;
+  services.mox.hostname = "mail";
+  services.mox.user = "admin@example.ke";
```
eljamm, line 5: Maybe we should use another hostname, since (from running `mox quickstart`):

> Quickstart assumes mox will run on the machine you run quickstart on and uses
its host name and public IPs. On many systems the hostname is not a fully
qualified domain name, but only the first dns "label", e.g. "mail" in case of
"mail.example.org". If so, quickstart does a reverse DNS lookup to find the
hostname, and as fallback uses the label plus the domain of the email address
you specified. Use flag -hostname to explicitly specify the hostname mox will
run on.

```suggestion
  services.mox.hostname = "localhost";
  services.mox.user = "admin@localhost";
```

eljamm:
I think the test fails in CI because it's not running with `.driver` or `.driverInteractive` as those give the VM an internet connection. To reproduce this, one can run:

```
nix build .#checks.x86_64-linux.projects/Mox/nixos/tests/basic
```

which will fail.

projects/Mox/module.nix
```diff
@@ -0,0 +1,84 @@
...
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+{
+  options = {
+    services.mox = {
+      enable = lib.mkEnableOption "Enable Mox Mail Server";
```
eljamm, line 11:
```suggestion
      enable = lib.mkEnableOption "Enable Mox Mail Server";
      package = lib.mkPackageOption pkgs "mox" { };
```

projects/Mox/module.nix
```diff
@@ -0,0 +1,84 @@
...
+        description = "Hostname for the Mox Mail Server";
+      };
+      user = lib.mkOption {
+        type = lib.types.str;
+        description = "*Required* Email user as (user@domain) to be created.";
+      };
+    };
+  };
+
+  config = lib.mkIf config.services.mox.enable {
```
eljamm, line 29:
```suggestion
  config = lib.mkIf cfg.enable {
```

projects/Mox/module.nix
```diff
@@ -0,0 +1,84 @@
...
+      before = [ "mox.service" ];
+      serviceConfig = {
+        Type = "oneshot";
+        RemainAfterExit = true;
+        User = "mox";
+        Group = "mox";
+      };
+      script = ''
+        mkdir -p /var/lib/mox
+        cd /var/lib/mox
```
eljamm, line 68: We can just use [serviceConfig.stateDirectory](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=) which is relative to `/var/lib`

```suggestion
```

projects/Mox/module.nix
```diff
@@ -0,0 +1,84 @@
...
+      after = [
+        "network-online.target"
+        "unbound.service"
+      ];
+      before = [ "mox.service" ];
+      serviceConfig = {
+        Type = "oneshot";
+        RemainAfterExit = true;
+        User = "mox";
+        Group = "mox";
```
eljamm, line 64: We can just use [stateDirectory](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=) (which is relative to `/var/lib`) and [WorkingDirectory](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#WorkingDirectory=)

```suggestion
        Group = "mox";
        StateDirectory = "mox";
        WorkingDirectory = "/var/lib/mox";
```

projects/Mox/module.nix
```diff
@@ -0,0 +1,84 @@
...
+      '';
+    };
+
+    systemd.services.mox = {
+      wantedBy = [ "multi-user.target" ];
+      after = [ "mox-setup.service" ];
+      requires = [ "mox-setup.service" ];
+      serviceConfig = {
+        WorkingDirectory = "/var/lib/mox";
+        ExecStart = "${pkgs.mox}/bin/mox -config /var/lib/mox/config/mox.conf serve";
```
eljamm, line 79:
```suggestion
        ExecStart = "${lib.getExe cfg.package} -config ${cfg.configFile} serve";
```

projects/Mox/module.nix
```diff
@@ -0,0 +1,84 @@
...
+      serviceConfig = {
+        Type = "oneshot";
+        RemainAfterExit = true;
+        User = "mox";
+        Group = "mox";
+      };
+      script = ''
+        mkdir -p /var/lib/mox
+        cd /var/lib/mox
+        ${pkgs.mox}/bin/mox quickstart -hostname ${config.services.mox.hostname} ${config.services.mox.user}
```
eljamm, line 69:
```suggestion
        ${lib.getExe cfg.package} quickstart -hostname ${cfg.hostname} ${cfg.user}
```

projects/Mox/module.nix
```diff
@@ -0,0 +1,84 @@
...
+  lib,
+  pkgs,
+  ...
+}:
+
+{
+  options = {
+    services.mox = {
+      enable = lib.mkEnableOption "Enable Mox Mail Server";
+      configFile = lib.mkOption {
```
eljamm, line 12: This is fine at the moment, but in the future we should just make the state directory (/var/lib/mox) configurable, since it's not enough to pass a config file (try passing a config file using this option, and it will complain about missing files).

projects/Mox/module.nix
```diff
@@ -0,0 +1,84 @@
...
+      isSystemUser = true;
+      name = "mox";
+      group = "mox";
+      home = "/var/lib/mox";
+      createHome = true;
+      description = "Mox Mail Server User";
+    };
+    users.groups.mox = {
+      name = "mox";
+    };
```
eljamm, line 46:
```suggestion
    users.groups.mox = { };
```

projects/Mox/test.nix
```diff
@@ -0,0 +1,97 @@
...
+  pkgs,
+  ...
+}:
+{
+  name = "mox";
+
+  nodes = {
+    machine =
+      { ... }:
+      rec {
```
eljamm, line 12: `rec` is not necessary, here

```suggestion
      {
```

projects/Mox/default.nix
```diff
@@ -0,0 +1,39 @@
+{
+  lib,
+  pkgs,
+  sources,
+}@args:
+{
+  metadata = {
+    summary = ''
+      Mox is a modern, secure, and open-source email server that implements all modern email protocols.
+      It makes it easy for people and organizations to run their own email server in minutes using mox quickstart.
```
eljamm, line 10: I'd say this is short and descriptive enough

```suggestion
      Modern, secure, and open-source email server that implements all modern email protocols.
```

projects/Mox/default.nix
```diff
@@ -0,0 +1,39 @@
...
+    '';
+    subgrants = [
+      "Mox"
```
eljamm, line 13:
```suggestion
      "Mox"
      "Mox-API"
      "Mox-Automation"
```

projects/Mox/test.nix
```diff
@@ -0,0 +1,97 @@
...
+
+        # Allow necessary ports through the firewall
+        networking.firewall.allowedTCPPorts = [
+          25
+          80
+          143
+          443
+          587
+          993
+          5335
```
eljamm, line 27: Are all of these necessary? Can we add comments on why they're needed?

eljamm (commented):
Well done @themadbit! This is shaping up really nicely.

projects/Mox/default.nix
```diff
@@ -0,0 +1,50 @@
...
+    mox = {
+      name = "mox";
+      module = ./programs/mox/module.nix;
+      examples.basic = {
+        module = ./programs/mox/examples/basic.nix;
+        description = "";
+        tests.basic = import ./programs/mox/tests/basic.nix args;
+      };
+    };
+  };
```
fricklerhandwerk, line 37: Can this application be used standalone, without the server running? Because if not, we don't need an extra `programs`. The description should contain an interaction, otherwise it's not clear at all what to do with it. It's also possible to attach `links` to that so users can check upstream documentation for reference.

projects/Mox/programs/mox/tests/basic.nix
```diff
@@ -0,0 +1,28 @@
...
+        ];
+      };
+  };
+
+  testScript =
+    { nodes, ... }:
+    ''
+      start_all()
+
+      machine.succeed()
```
fricklerhandwerk, line 26: This isn't testing anything

projects/Mox/services/mox/tests/basic.nix
```diff
@@ -0,0 +1,116 @@
...
+  testScript =
+    { nodes, ... }:
+    ''
+      start_all()
+
+      # Wait for machine to be available
+      machine.wait_for_unit("multi-user.target")
+
+      # Verify the mox-setup service has run successfully
+      machine.wait_for_unit("mox-setup.service")
```
fricklerhandwerk, line 96: Isn't that implied by `mox.service` being ready?

projects/Mox/services/mox/module.nix
```diff
@@ -0,0 +1,85 @@
...
+  cfg = config.services.mox;
+in
+{
+  options = {
+    services.mox = {
+      enable = lib.mkEnableOption "Mox";
+      package = lib.mkPackageOption pkgs "mox" { };
+      configFile = lib.mkOption {
+        type = lib.types.str;
+        default = "/var/lib/mox/config/mox.conf";
```
fricklerhandwerk, line 17: Is it the desired mode of interaction for the configuration file to be mutable? Because it would be a very unusual thing to do on NixOS. You'd instead compose the configuration file from the module options, and invoke the service such that it uses that file from the Nix store.

projects/Mox/services/mox/tests/basic.nix
```diff
@@ -0,0 +1,116 @@
...
+        environment.etc."resolv.conf".text = ''
+          nameserver 127.0.0.1
+        '';
+
+        # Networking
+        networking.firewall.allowedTCPPorts = [
+          25 # SMTP
+          80 # HTTP
+          143 # IMAP
+          443 # HTTPS
```
fricklerhandwerk, line 33: There's a convention to set that in the service module, guarded by `openPorts` -- check https://search.nixos.org/options for examples.

projects/Mox/services/mox/tests/basic.nix
```diff
@@ -0,0 +1,116 @@
...
+    machine =
+      { ... }:
+      {
+        imports = [
+          sources.modules.ngipkgs
+          sources.modules.services.mox
+          sources.examples.Mox.mox
+        ];
+
+        # Environment packages
```
fricklerhandwerk, line 19:
```suggestion
```

No need for those LLM-generated comments, this is perfectly conventional code and the comments won't help anyone because they say almost literally what the code says.

projects/Mox/services/mox/tests/basic.nix
```diff
@@ -0,0 +1,116 @@
...
+            ];
+            local-data = [
+              "\"example.com. IN A 12.34.56.78\""
+              "\"mail.example.com. IN A 127.0.0.1\""
+              # Provide a static response for com.
+              "\"com. IN NS localhost\""
+              "\"localhost. IN A 127.0.0.1\""
+            ];
+          };
+        };
```
fricklerhandwerk, line 70: It's not entirely clear to me what that code is for, explanatory comments would help here.

projects/Mox/services/mox/module.nix
themadbit: Mox uses `sconf,` which is not supported in NixOS config file types yet. Also, autogenerating the config files for Mox using the options already in the module is more reproducible. I suggest making the exception and documenting that.

projects/Mox/services/mox/tests/basic.nix
themadbit: True. Was helpful in debugging. I'll remove that

projects/Mox/default.nix
themadbit: Yes, it can be used. MOx has many other subcommand that talk to the running service. A program for that would be helpful.

projects/Mox/services/mox/tests/basic.nix
themadbit: I'll document it better

projects/Mox/services/mox/tests/basic.nix
themadbit: I did not find this in networking options. What service were you referring to?

eljamm: The option is usually called [openPorts](https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=openPorts) or [openFirewall](https://search.nixos.org/options?channel=24.11&from=0&size=50&sort=relevance&type=packages&query=openfirewall). Check out those services' modules for the implementation.

themadbit: I'm having trouble implementing this. Using `services.prometheus.exporters.unbound.openFirewall` allows only one port to be opened and  `mox`  requires additional ports. What's the downside of using the current implementation?

eljamm: You're just defining a `services.mox.openFirewall` option which will open the necessary `networking.firewall.allowedTCPPorts` and `networking.firewall.allowedUDPPorts` when it's enabled. The `services.unbound` service shouldn't matter in this case.

themadbit: now, I get what you mean.

projects/Mox/services/mox/module.nix
```diff
@@ -0,0 +1,109 @@
...
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [
+      cfg.package
+    ];
+
+    networking.firewall = lib.mkIf cfg.openFirewall {
```
eljamm, line 62: It's redundant to have both `openFirewall` and `openPorts` options as they're essentially the same thing. A switch/toggle that enables some necessary ports for the service. In this case, I don't know how desirable it is to make those ports configurable to users.

themadbit: I get that. My earlier approach was to explicitly allow the ports in the test and leave the in-production configuration for users to figure out. However, I believe it is more desirable to have an openPorts option, as Mox relies on a couple of ports, and it is ultimately extensible if more ports are required. Noted on the redundancy, thanks. What do you think?

eljamm: In this case, I think it would be better to have each port be configurable by itself (for example, see [this](https://github.com/NixOS/nixpkgs/blob/398ebeaf3620da9fb78d22d7de1c55c2f42d2735/nixos/modules/services/networking/vwifi.nix#L83-L88) and [this](https://github.com/NixOS/nixpkgs/blob/398ebeaf3620da9fb78d22d7de1c55c2f42d2735/nixos/modules/services/networking/vwifi.nix#L162-L165)) as it makes it clear what the purpose of each one is.

themadbit:
> The issue with the ubuntu test was resolved in #1054 and I'm in the process of fixing CI, now.

ooh, okay

eljamm (approved):
Looks good, well done @themadbit!

PR #681, opened 2025-03-27
WIP heads: init at 0.2.1-unstable-2025-03-12

### LINKED ISSUE #547: Implement project metadata for Heads-OpenPGP
Implement project metadata in 'projects/Heads-OpenPGP/default.nix'

OPNA2608:
Closes #547

This is WIP and doesn't fully build yet, just opening this to show progress as I switch development back to my (hopefully-revived, will see…) desktop.

We're running upstream's build setup, which requires significant fetching & some patching to behave nicely.

- We can't download things at build-time, so all the download that heads, and subproject that heads fetches & builds as part of its build process, are entered into `deps.nix` and placed/patched in where needed before starting the build - or in the case of subprojects that get unpacked during the build, have their patch files prepared & put into place for heads to apply during the build.
- Ahead of the build, heads inspects some things about the system and uses the information for scheduling the build, along with printing this information. For the sake of making it behave the same on different systems & at different times on the same system, we're patching some of those checks & outputs


pkgs/by-name/heads/deps.nix
```diff
@@ -0,0 +1,274 @@
+{
```
eljamm, line 1: How would one update the deps, and is it possible to do it automatically?

OPNA2608: Right now, it's all collected by iteratively running the build for the selected `BOARD`. seeing what it tries to fetch, and adding it to the list. It should be possible to automate this though:

- `/modules/*` contains Makefile code with all the heads-managed dependencies. Right now I have only entered into `deps.nix` what I actually needed for the picked `BOARD` target, but other targets require different versions of coreboot & linux. They're nonetheless all listed in the corresponding `modules` file though, so it should be possible to have them extracted from the files… I don't think Makefile has a nice way of requesting fully resolved variable values, so it'll involve some custom Makefile code.
- The `musl-cross-make` has its [own list of dependencies that it tries to fetch, which is not managed by heads](https://github.com/richfelker/musl-cross-make/blob/fd6be58297ee21fcba89216ccd0d4aca1e3f1c5c/Makefile#L4-L79) and depends on the fetched version of `musl-cross-make`. It's also managed in the Makefile, so it'll require similar custom code to extract the necessary information.

OPNA2608: Do we have any infra for running `passthru.updateScript`s of packages in ngipkgs, in the same way that you would in Nixpkgs (i.e. `nix-shell maintainers/scripts/update.nix --argstr package heads`)? They aren't necessarily derivations (for example, when needing to use `_experimental-update-script-combinators.sequence` - see `venera` in Nixpkgs), so `nix build`ing doesn't really work.

fricklerhandwerk: Not yet. We have https://github.com/ngi-nix/summer-of-nix/issues/121 to track it but this is not planned/prioritised yet. Pinging @Erethon @wegank again, maybe you all could jot down a few notes or prepare when work on that item can start.

OPNA2608: `deps.nix` is now 100% generated by an update script. Until support for properly running `.passthru.updateScript` is added, the `deps.nix`-relevant part can be called manually by building `heads.<any board>.passthru.updateDepsScript` and running the executable in the resulting output.

eljamm:
Would it be possible to split this into smaller PRs? Both so that CI can build the packages properly without timing out and for us to review it better.

pkgs/by-name/heads/package.nix
```diff
@@ -0,0 +1,302 @@
...
+  nix,
+  nixfmt-rfc-style,
+  nix-prefetch-scripts,
+  perl,
+  pkg-config,
+  rsync,
+  xz,
+  zip,
+  zlib,
+  brandName ? "Heads",
```
eljamm, line 42: It would probably be nice to sort the deps for better readability

```suggestion
  lib,
  stdenv,

  fetchgit,
  fetchFromGitHub,
  fetchurl,

  replaceVars,
  runCommandNoCC,
  newScope,

  bashNonInteractive,

  # build deps
  bc,
  bison,
  ccache,
  cmake,
  cpio,
  curl,
  flex,
  gnum4,
  git,
  imagemagick,
  ncurses,
  perl,
  pkg-config,
  rsync,
  zip,
  gnat,
  gnat-bootstrap,

  # runtime deps
  elfutils,
  zlib,

  # update script
  unstableGitUpdater,
  writeShellApplication,
  _experimental-update-script-combinators,
  envsubst,
  getopt,
  gnumake,
  jq,
  nix,
  nixfmt-rfc-style,
  nix-prefetch-scripts,
  xz,

  brandName ? "Heads",
```

OPNA2608: Not a fan of these group labels (i.e. `# update script`) - dependencies may over time move between them, or belong into multiple at the same time. But I will try to resort them into roughly:
```nix
{
  lib,
  stdenv,
  # Nix functions
  # Packages used to build time
  # Variables
}:
```

pkgs/by-name/heads/package.nix
```diff
@@ -0,0 +1,432 @@
...
+}:
+
+let
+  # Theoretically, it should be possible to extract these in the updateScript by running the corresponding download
+  # scripts in /blobs/* with a curl/wget that prints out the given URL, and collecting all the requested URLs & their
+  # output names
+  blobs = [
```
OPNA2608, line 54: Note: I am *unhappy* to announce that while `deps.nix` is still 100% generated, the new `blobs` attrset in `package.nix` likely belongs in there logically… It was just easier to extract those few downloads with their peculiarities by hand than to hack together an automatable solution for now.

Unlike the *mountain of dependencies* that is `deps.nix`, manually updating these by skimming the scripts in `/blobs/*` should at least be fairly doable.

OPNA2608:
> Would it be possible to split this into smaller PRs? Both so that CI can build the packages properly without timing out and for us to review it better.

To help with CI load, I've only allowed a single board target to be generated for now - `heads.qemu-coreboot-fbwhiptail-tpm1-hotp`.

The bulk of the code here is still necessary for building even this mostly-emulated board target. The only things I could cut for now are:

- All the `blobs` files, their collecting & patching of blobs-related code - that's only relevant for boards that *have* blobs
- Maybe some of the `coreboot` entries in `patches` - only need the one that's relevant for the currently allowed board
- I think the `openssl` patch is also not yet relevant for the board?

If you think cutting them for now is worth it, then I can go ahead with that.

pkgs/by-name/heads/package.nix
eljamm: I know that this might be a matter of preference so this is merely a suggestion, but even having them separated by spaces is more readable to me and easier to discern which parts are related, as opposed to having a long list of everything.

OPNA2608: Spaces added.

projects/Heads/default.nix
```diff
@@ -0,0 +1,55 @@
...
+
+  nixos.modules.programs = {
+    heads = {
+      name = "heads";
+      module = ./module.nix;
+      examples.qemu-coreboot-fbwhiptail-tpm1-hotp = {
+        module = ./example.nix;
+        description = ''
+          Builds heads for the example qemu-coreboot-fbwhiptail-tpm1-hotp board, and makes the ROM image available
+          at a fixed location. For testing it in a VM.
```
fricklerhandwerk, line 25:
```suggestion
          at a fixed location, for testing in a VM.
```

projects/Heads/default.nix
```diff
@@ -0,0 +1,55 @@
...
+          url = "https://github.com/linuxboot/heads/blob/594abed8639b4f4a7fc9b7898d85eb48acbd0072/targets/qemu.md";
+        };
+      };
+    };
+  };
+  # TODO: Referencing `pkgs` here is currently causing eval issues all over ngipkgs.
+  # Resolve this first before enabling this.
+  /*
+    binary = {
+      # TODO: Maybe lib.attrsets.mapAttrs over derivations in `heads` scope to
```
fricklerhandwerk, line 49: Yes

projects/Heads/module.nix
```diff
@@ -0,0 +1,22 @@
...
+let
+  cfg = config.programs.heads;
+in
+{
+  # Note: Heads produces ROM images intended to be flashed onto real hardware.
+  # This module only exists to allow automated testing of a QEMU-targeting
+  # Heads ROM.
+  options.programs.heads = {
+    enable = lib.mkEnableOption "symlinking the Heads ROM for qemu-coreboot-fbwhiptail-tpm1-hotp";
+  };
```
fricklerhandwerk, line 16: We do have binary artefacts in our data model now!

OPNA2608: This is used in the test, to have a fixed location where we can find the ROM at. @eljamm said that currently, it's not possible to refer to the `binary` attrset.

Also, referencing `pkgs` in `binary` (in order to point it at the artifact from the build), currently makes eval blow up, which is why it's commented out: https://github.com/ngi-nix/ngipkgs/pull/773

eljamm: > @eljamm said that currently, it's not possible to refer to the binary attrset.

Because it's [not mapped back](https://github.com/ngi-nix/ngipkgs/blob/51793ff4b032d718e21a5f15620171126f5a26f8/default.nix#L110-L119), but we can do this if we need to.

eljamm (approved):
This looks good to me. As a next step, we need to fix the overview in https://github.com/ngi-nix/ngipkgs/pull/773 so we can use the binary type.

PR #683, opened 2025-03-27
Openfire: move projects-old/ -> projects

### LINKED ISSUE #592: Migrate Openfire to projects
The project needs to be moved from `/projects-old` into `/projects` and its artefacts tracked to match our [project template](https://github.com/ngi-nix/ngipkgs/tree/main/templates/project).

## Instructions

1. Copy the `templates/project/` template to the `projects/` directory and re-name it to be the same as the project.
1. Use the data from the `projects-old/<project_name>` entry,  to fit the template.
1. Remove any extra/unneeded things from the template that do not apply to the migrated project (e.g. a project has no services, ...).
1. Check that the code is valid by running the test locally:

```
# For examples
$ nix build .#checks.x86_64-linux.projects/<project_name>/nixos/examples/<example_name>

# For tests
$ nix build .#checks.x86_64-linux.projects/<project_name>/nixos/tests/<test_name>
```

5. Delete the `projects-old/<project_name>` directory and open a PR

3pleX-dev:
should resolve #592 
@eljamm please help review


projects/Openfire/default.nix
```diff
@@ -0,0 +1,25 @@
...
+  };
+
+  nixos.modules.services = {
+    openfire-server = {
+      name = "openfire-server";
+      module = ./module.nix;
+      examples.openfire-server = {
+        module = ./example.nix;
+        description = "";
+        tests.basic = import ./test.nix args;
```
3pleX-dev, line 21: @eljamm, just curious—why do we use `import ./test.nix args` here (as you suggested on the call) but not for something like `./example.nix`?

Sorry for the noob question! I'm just trying to understand better—maybe you can point me to some relevant parts of the Nix docs?

eljamm: That's a totally valid question! You see, [tests need to be derivations](https://github.com/ngi-nix/ngipkgs/blob/9b4ae90d8cb79bb53f9770b9cdb2020d4d006ffa/flake.nix#L163-L166) in order to be be checked by our CI. Currently, we handle [different types of tests](https://github.com/ngi-nix/ngipkgs/blob/9b4ae90d8cb79bb53f9770b9cdb2020d4d006ffa/default.nix#L241-L252). If it's already a derivation, pass it. If it's a string, import it (usually [references nixpkgs tests](https://github.com/ngi-nix/ngipkgs/blob/9b4ae90d8cb79bb53f9770b9cdb2020d4d006ffa/projects/Mastodon/default.nix#L22-L23)) and for everything else convert it to a [nixosTest](https://github.com/ngi-nix/ngipkgs/blob/9b4ae90d8cb79bb53f9770b9cdb2020d4d006ffa/default.nix#L205-L227).

A path like `./test.nix` is not a string and is not a derivation, so it will be passed to nixosTest. One of the internal functions of nixosTest (called [simpleTest](https://github.com/NixOS/nixpkgs/blob/5a1a99b415290cc45ddbbcaf1c0c37666b8b1014/nixos/lib/testing-python.nix#L99)) takes an attribute set as input, and will error out if we give it a path:

```
         at /nix/store/gs0dh6x7x828j6qg9fnsv4khic9agmc0-source/nixos/lib/testing-python.nix:99:23:
           98|
           99|     simpleTest = as: (makeTest as).test;
             |                       ^
          100|

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: expected a set but found a path: /home/kuroko/Files/System/Development/Git/Personal/nix/ngipkgs/.worktree/review/projects/Openfire/test.nix
```

Note: Even the `simpleTest`'s input is called `as`, which I assume stands for attribute set.

So, by [importing the file](https://nix.dev/tutorials/nix-language#import) we evaluate it and that gives us back an attribute set that we then pass to `nixosTest`, which will turn it into a derivation which will be ran by our CI to check the test.

Please let me know if this adequately answers your question 😁

projects/Openfire/test.nix
```diff
@@ -5,17 +5,10 @@
   ...
 }:
 {
-  # NOTE:
```
eljamm, line 8: IMO this is a useful note, is there any reason to remove it?

projects/Openfire/test.nix
```diff
@@ -5,17 +5,10 @@
...
-  # - Diable `Restrict Admin Console Access` in the `Server Settings`, else you won't be able to login.
-
   name = "openfire";
-  meta = {
-    maintainers = [ ];
-  };
 
   nodes = {
-    server =
+    machine =
```
eljamm, line 11: Curious why the naming was changed from `server` to a more generic `machine` 🤔

projects/Openfire/default.nix
3pleX-dev: Ah! I see, very detailed explanation 👍 
Thank you so much!

projects/Openfire/test.nix
3pleX-dev: Haha! this was before I understood the flow, I was trying to strictly follow the template 😆.

eljamm: Guess the template was more restrictive than I anticipated 😅

projects/Openfire/default.nix
```diff
@@ -0,0 +1,25 @@
+{
+  lib,
+  pkgs,
+  sources,
+}@args:
+{
+  metadata = {
+    summary = "Openfire is a real time collaboration (RTC) server licensed under the Open Source Apache License.";
+    subgrants = [
+      "Openfire"
```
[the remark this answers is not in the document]
3pleX-dev: Ah nice! I was going to ask this on the call too but it skipped my mind, thanks for bringing it up 👍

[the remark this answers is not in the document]
3pleX-dev: Just checked now, I see we're using the paths and not the actual titles yeah?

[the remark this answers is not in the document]
eljamm: Yup. Ideally one would start with the [triaging](https://github.com/ngi-nix/ngipkgs/issues/new?template=project-triaging.yaml) of the project, which gives us the necessary information for the template (like the subgrants) before adding creating its entry in the source code, but it's not a big deal since we can iterate on this. Plus, migrating old projects is more important than triaging because that would allow us to drop some [nasty mapping code](https://github.com/ngi-nix/ngipkgs/blob/9b4ae90d8cb79bb53f9770b9cdb2020d4d006ffa/default.nix#L64-L123).

projects/Openfire/default.nix
```diff
@@ -0,0 +1,27 @@
...
```
eljamm, line 10: This should be removed as there is no such subgrant:

```suggestion
```

eljamm (approved):
LGTM. Thanks!

PR #684, opened 2025-03-27
overview: rebuild the rendering flow so that we can use markdown strings

erictapen:
As discussed with @fricklerhandwerk 

tmp WTMliGJu6y

This brings back markdown formatting in option descriptions.


fricklerhandwerk (approved):
This is properly cursed, but seems like the right thing to do here.

Issue #575, opened 2025-03-11
Spike: Installation instructions for projects


erictapen:
As a user of NGIpkgs, I want to be able to run software that is part of NGI projects. As we package many different kinds of software projects, this can mean different things:

- *Executables:* Make an executable available in the terminal of my system, so that I can try it out
- *Libraries:* Use a library in my project
- *Applications:* Install an application on my system, so that I can use it permanently
- *Services:* Run a service as part of my NixOS system closure

This can mean different things on different platforms.

```mermaid
graph TD;
  subgraph Start
    direction TB
    A[User uses macOS]
    X[User uses non-NixOS Linux]
    D[User uses NixOS]
  end

  A --> B[Aquire Nix];
  X --> C[Aquire Nix];
  C --> Y[Start a KVM VM with a preconfigured service];
  Y --> M;
  D --> E[nix-shell/nix shell command];
  B --> E;
  E --> F[Executable];
  B --> G[Example code for getting a derivation to be available in your Nix project];
  D --> G;
  G --> H[Library];
  D --> I[Using NGIpkgs in your system configuration];
  I --> J[Example code to configure the application in your system configuration];
  J --> K[Application];
  I --> L[Example code to configure the service in your system configuration];
  L --> M[Service];
```

<hr>

Related: https://github.com/ngi-nix/ngipkgs/issues/516
Inspiration: https://geospatial-nix.today/

## Dod

Given the list of NGI projects exists, when choose some NGI project, then I am able to run it on my non-Nix Linux machine.

PR #685, opened 2025-03-27
overview: render project snippets with deliverable type and summary text

erictapen:
For now I'd only show the type of existing deliverables and the summary text if existent.

tmp kCljCTf0fn

### DIFF
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -209,16 +209,16 @@ let
       '';
       many =
         project:
-        optionalString (project.packages != { }) (one "Executable")
+        optionalString (project.packages != { }) (one "program")
         +
           # TODO is missing in the model yet
-          optionalString false (one "Library")
+          optionalString false (one "library")
         + optionalString (project.nixos.modules ? services && project.nixos.modules.services != { }) (
-          one "Service"
+          one "service"
         )
         +
           # TODO is supposed to represent GUI apps and needs to be distinguished from CLI applications
-          optionalString false (one "Application");
+          optionalString false (one "application");
     };
 
     # The snippets for each project that are rendered on https://ngi.nixos.org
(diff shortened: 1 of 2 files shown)


overview/default.nix
```diff
@@ -184,26 +184,56 @@ let
...
       many =
-        projects:
-        concatLines (
-          mapAttrsToList (name: _: ''
-            <article>
-              <a href="/project/${name}">${name}</a>
-            </article>
-          '') projects
-        );
+        project:
```
eljamm, line 203: Just a naming nit, but I believe the input should be plural (`projects`) for the many , right?

eljamm (approved):
This looks very cool. Thanks!

overview/default.nix
erictapen: No, because this is for many tags that belong to a singular project. I decided to extract the list of tags in the `render.deliverableTags` section as this logic is about nothing else.

eljamm: Ah, that makes sense.

fricklerhandwerk (approved):
Nice, except for the bikeshed.

overview/default.nix
```diff
@@ -186,26 +186,56 @@ let
...
+        optionalString (project.packages != { }) (one "Executable")
```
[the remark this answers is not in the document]
erictapen: I'm not even sure where I got these four categories from, but I also used these in https://github.com/ngi-nix/ngipkgs/issues/575.

I don't care much about how we call them currently, as we get the data right from the model and changing the implementation is still easy. So let's go with `program` for now. We might stumble over better terms through talking with users/explaining NGIpkgs to them.

fricklerhandwerk (approved):

PR #687, opened 2025-03-27
Fix naming for consistency

themadbit:
Update wording (`service` to `module`) for consistency with project template naming.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,12 +85,12 @@ Instead, write one sentence per line, as this makes it easier to review changes.
 
    Make sure to:
 
-   - Add the module options in `service.nix`, and reference that file in `default.nix`.
+   - Add the module options in `module.nix`, and reference that file in `default.nix`.
      For example:
 
      ```nix
      nixos.modules = {
-       services.some-project.module = ./service.nix;
+       services.some-project.module = ./module.nix;
      };
      ```
 


fricklerhandwerk:
The naming was correct: a module is the data type which a service -- an application component -- is implemented in. Programs, another application component type, are also implemented in terms of modules.

CONTRIBUTING.md
```diff
@@ -85,12 +85,12 @@ Instead, write one sentence per line, as this makes it easier to review changes.
 
    Make sure to:
 
-   - Add the module options in `service.nix`, and reference that file in `default.nix`.
```
themadbit, line 88: @fricklerhandwerk What prompted me to suggest the changes is that `service.nix` is not referred elsewhere, and only `module.nix` exists, mainly if a new contributor follows the doc to the tee, they might be confused by what/where `service.nix` is.

fricklerhandwerk: In that case pleas rename the file! :)

themadbit: Noted. Could you check me on this? For consistency, should I rename the template file `module.nix` to `service.nix`? All new projects created thus far, and those migrating from `projects-old`, are using `module.nix` to create the service and or program components. I foresee a situation where it might be interpreted as `service.nix` is used to package services and another maybe `module/program.nix` used to package programs. I would suggest that `module.nix` seems more consistent in being used for both programs and services, and it's consistent with the data type used to create them. Let me know, and I'll proceed with the better approach. The aim is to have a consistent naming and approach.

btw, It seems like I can't reopen this PR. Maybe because you closed it?

fricklerhandwerk: Yeah you're right. Also given that this template is more about being educational than being normative, brevity is quite important. Capturing all these details in a short text will take quite a bit of effort (and still risk that people get confused) and that is simply not worth it at the moment. I'll merge the PR as is then!

Thanks for bearing with me. A lot of what we're doing is figuring out what's the right thing to do. :)

fricklerhandwerk (approved):

CONTRIBUTING.md
themadbit: Yeah, figuring things out is the game here. And thanks for pointing out things to consider

eljamm:
Thanks for the fix!

Issue #711, opened 2025-04-03
Implement project metadata for Galene


eljamm:
## Instructions
  
Copy the project template and edit it with relevant details from [the triaged data](https://github.com/ngi-nix/ngipkgs/issues/663):

   ```shellSession
   cp -r templates/project projects/Galene
   $EDITOR projects/Galene/default.nix
   ```

themadbit:
Does this seem like a dup? [issue](https://github.com/ngi-nix/ngipkgs/issues/712) refers to this and assumes it has been done. Maybe have it as sub-issue

eljamm:
This can be done before the other issue, so I wouldn't call it a dup, but having it as a sub-issue is a good idea.

Issue #663, opened 2025-03-25
NGI Project: Galene


3pleX-dev:
### Short Summary

Galene is a high-quality, self-hosted videoconferencing server designed for ease of installation, privacy, and efficient resource usage. In production since 2020, it has supported university lectures, staff meetings, and international conferences. The project aims to enhance Galene with advanced networking and video algorithms, improved management features, and user-facing additions like background blur and automatic subtitling.

### NLnet page(s)

N/A

### Website

https://galene.org/

### Source repository

- https://github.com/jech/galene
  - Language/Framework: Go
  - Dependency management: Go modules (go.mod, go.sum)
  - Nix development environment: N/A

### Documentation

- Usage Examples:
  - https://github.com/jech/galene
- Build from source/Development:
  - https://galene.org/INSTALL.html
- Other:
  -

### Artefacts

- GUI:
  - galene:
    - documentation: https://galene.org/README.html
    - examples: https://github.com/jech/galene

### Nixpkgs/NixOS

- Packages:
  - [Galene](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/by-name/ga/galene/package.nix#L27)
- Services:
  - [services.galene.insecure](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)
  - [services.galene.package](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)
  - [services.galene.group](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)
  - [services.galene.httpAddress](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)
  - [services.galene.enable](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)
  - [services.galene.user](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)
  - [services.galene.stateDir](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)
  - [services.galene.keyFile](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)
[...]

PR #716, opened 2025-04-04
Add project Galene

themadbit:
Resolves [issue](https://github.com/ngi-nix/ngipkgs/issues/711).

- Adds Galene to projects
- Adds project metadata


projects/Galene/default.nix
```diff
@@ -0,0 +1,39 @@
...
+          text = "Usage Instructions";
+          url = "https://galene.org/README.html";
+        };
+      };
+    };
+  };
+
+  # NOTE: same structure as programs
+  nixos.modules.services = {
+  };
```
eljamm, line 38: If this doesn't exist, we can remove it

```suggestion
```

projects/Galene/default.nix
```diff
@@ -0,0 +1,39 @@
...
+}@args:
+{
+  metadata = {
+    summary = "Galene is a self-hosted video conferencing server. It features advanced networking and video algorithms and automatic subtitling.";
+    subgrants = [
+      "Galene"
+    ];
+  };
+
+  nixos.modules.programs = {
```
eljamm, line 14: I believe that this should be `services`, since we have [services.galene](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix) from the [triaged data](https://github.com/ngi-nix/ngipkgs/issues/663)

```suggestion
  nixos.modules.services = {
```

projects/Galene/default.nix
```diff
@@ -0,0 +1,39 @@
...
+    galene = {
+      name = "galene";
+      module = ./module.nix;
```
eljamm, line 17: The module already exists [in nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/services/web-apps/galene.nix)

```suggestion
      module = "${sources.inputs.nixpkgs}/nixos/modules/services/web-apps/galene.nix";
```

projects/Galene/module.nix
eljamm, line 1: This can be removed. The module exists in nixpkgs.

projects/Galene/example.nix
```diff
@@ -0,0 +1,4 @@
+{ ... }:
+{
+  programs.galene.enable = true;
```
eljamm, line 3:
```suggestion
  services.galene.enable = true;
```

eljamm (commented):
Looking good as a foundation. @Ellipse0934, you can build upon this for the example and test.

themadbit:
I've seen a new project implementation structure. @Ellipse0934 can implement that when working on tests and examples.

projects/Galene/module.nix
eljamm, line 1: This needs to be removed since we're not importing it in `default.nix`

PR #748, opened 2025-04-08
Add non-flake pre-commit and formatter

eljamm:
```
nix-build . -A formatter && ./result/bin/formatter
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -11,6 +11,7 @@
 let
   dream2nix = (import sources.dream2nix).overrideInputs { inherit (sources) nixpkgs; };
   sops-nix = import "${sources.sops-nix}/modules/sops";
+  pre-commit-hooks = import sources.pre-commit-hooks;
 in
 rec {
   inherit
@@ -127,6 +128,33 @@ rec {
   templates.project = project-models.project (
     import ./templates/project { inherit lib pkgs sources; }
   );
+
+  pre-commit-hook = pre-commit-hooks.run {
+    src = ./.;
+    hooks = {
+      actionlint.enable = true;
+      nixfmt-rfc-style.enable = true;
+      actionlint.package = pkgs.actionlint;
+      nixfmt-rfc-style.package = pkgs.nixfmt-rfc-style;
+    };
+  };
+
+  formatter =
+    {
+      pre-commit ? pre-commit-hook,
+    }:
+    pkgs.writeShellApplication {
+      name = "formatter";
+      text = ''
+        # shellcheck disable=all
+        shell-hook () {
+          ${pre-commit.shellHook}
+        }
+
+        shell-hook
+        pre-commit run --all-files
+      '';
+    };
 
   # TODO: find a better place for this
   metrics = with lib; {
--- a/flake.nix
+++ b/flake.nix
@@ -223,18 +223,7 @@
             buildInputs = checks."infra/pre-commit".enabledPackages;
           };
 
-          formatter = pkgs.writeShellApplication {
-            name = "formatter";
-            text = ''
-              # shellcheck disable=all
-              shell-hook () {
-                ${checks."infra/pre-commit".shellHook}
-              }
-
-              shell-hook
-              pre-commit run --all-files
-            '';
-          };
+          formatter = classic.formatter { pre-commit = checks."infra/pre-commit"; };
         }
       );
     in


default.nix
```diff
@@ -128,6 +129,33 @@ rec {
...
+      actionlint.enable = true;
+      nixfmt-rfc-style.enable = true;
+      actionlint.package = pkgs.actionlint;
+      nixfmt-rfc-style.package = pkgs.nixfmt-rfc-style;
+    };
+  };
+
+  formatter =
+    {
+      pre-commit ? pre-commit-hook,
```
fricklerhandwerk, line 144: Why is this a parameter if the same value is used at all call sites?

default.nix
```diff
@@ -128,6 +129,33 @@ rec {
     import ./templates/project { inherit lib pkgs sources; }
   );
 
+  pre-commit-hook = pre-commit-hooks.run {
```
fricklerhandwerk, line 132: Why is this repeating the invocation in flake.nix? It seems to use a slightly different interface there (selecting system via attribute) but still does pretty much the same.

default.nix
```diff
@@ -128,6 +129,33 @@ rec {
...
+      nixfmt-rfc-style.package = pkgs.nixfmt-rfc-style;
+    };
+  };
+
+  formatter =
+    {
+      pre-commit ? pre-commit-hook,
+    }:
+    pkgs.writeShellApplication {
+      name = "formatter";
```
fricklerhandwerk, line 147: Would be good to add this to the devshell, so one can just run something like `format` be done with it. It would be a way better UX than with `nix fmt` because that always takes forever to evaluate the underlying expression.

fricklerhandwerk (commented):
This is very good to have, but let's actually move the code over instead of duplicating it.

default.nix
eljamm: Because we can't use the same pre-commit hook. If I try to use the one from `default.nix` in `flake.nix`, I get the following error:

```
       … while calling the 'import' builtin
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:167:19:
          166|
          167|           flake = import (outPath + "/flake.nix");
             |                   ^
          168|

       … while calling the 'storePath' builtin
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:115:15:
          114|               # If it's already a store path, don't copy it again.
          115|               builtins.storePath src
             |               ^
          116|             else

       error: 'builtins.storePath' is not allowed in pure evaluation mode
```

default.nix
```diff
@@ -128,6 +129,33 @@ rec {
     import ./templates/project { inherit lib pkgs sources; }
   );
 
+  pre-commit-hook = pre-commit-hooks.run {
+    src = ./.;
+    hooks = {
+      actionlint.enable = true;
+      nixfmt-rfc-style.enable = true;
+      actionlint.package = pkgs.actionlint;
+      nixfmt-rfc-style.package = pkgs.nixfmt-rfc-style;
```
eljamm, line 138: Note that without this, another version of `nixfmt-rfc-style` is used and formatting is not the same.

default.nix
eljamm: Because we can't define it here and use it in the flake as well (see comment)

default.nix
fricklerhandwerk: This looks like someone is holding it wrong. Where is this trace coming from, on our end or in pre-commit?

eljamm: It traces back to `pre-commit`:

<details><summary>full trace</summary>
<p>

```shellSession
❯ nix fmt --show-trace
error:
       … while evaluating 'strict' to select 'drvPath' on it
         at /builtin/derivation.nix:1:552:
       … while calling the 'derivationStrict' builtin
         at /builtin/derivation.nix:1:208:
       … while evaluating derivation 'formatter'
         whose name attribute is located at /nix/store/dn9smqqg60pn1a3c1sk2m0m8iakm84zs-source/pkgs/stdenv/generic/make-derivation.nix:438:13

       … while evaluating attribute 'text' of derivation 'formatter'
         at /nix/store/dn9smqqg60pn1a3c1sk2m0m8iakm84zs-source/pkgs/build-support/trivial-builders/default.nix:129:13:
          128|           inherit
          129|             text
             |             ^
          130|             executable

       … while evaluating 'pre-commit' to select 'shellHook' on it
         at /nix/store/zhvz9hn1cr1wd3cxf9b5g2bpwbh8kc0v-source/default.nix:151:13:
          150|         shell-hook () {
          151|           ${pre-commit.shellHook}
             |             ^
          152|         }

       … while evaluating the attribute '"infra/pre-commit"'
         at /nix/store/zhvz9hn1cr1wd3cxf9b5g2bpwbh8kc0v-source/flake.nix:206:17:
          205|               checksForInfrastructure = {
          206|                 "infra/pre-commit" = classic.pre-commit-hook;
             |                 ^
          207|                 "infra/makemake" = toplevel self.nixosConfigurations.makemake;

       … while evaluating the attribute 'pre-commit-hook'
         at /nix/store/zhvz9hn1cr1wd3cxf9b5g2bpwbh8kc0v-source/default.nix:132:3:
          131|
          132|   pre-commit-hook = pre-commit-hooks.run {
             |   ^
          133|     src = ./.;

       … while evaluating 'pre-commit-hooks' to select 'run' on it
         at /nix/store/zhvz9hn1cr1wd3cxf9b5g2bpwbh8kc0v-source/default.nix:132:21:
          131|
          132|   pre-commit-hook = pre-commit-hooks.run {
             |                     ^
          133|     src = ./.;

       … while calling the 'import' builtin
         at /nix/store/zhvz9hn1cr1wd3cxf9b5g2bpwbh8kc0v-source/default.nix:14:22:
           13|   sops-nix = import "${sources.sops-nix}/modules/sops";
           14|   pre-commit-hooks = import sources.pre-commit-hooks;
             |                      ^
           15| in

       … while evaluating the file '/nix/store/8bh3jgq1riy3jxm07vy4xxzvk9xd74pc-source/default.nix':

       … in the left operand of the update (//) operator
         at /nix/store/8bh3jgq1riy3jxm07vy4xxzvk9xd74pc-source/default.nix:13:37:
           12| in
           13| flake.lib.${builtins.currentSystem} // flake.packages.${builtins.currentSystem}
             |                                     ^
           14|

       … while evaluating 'flake' to select 'lib."${(builtins).currentSystem}"' on it
         at /nix/store/8bh3jgq1riy3jxm07vy4xxzvk9xd74pc-source/default.nix:13:1:
           12| in
           13| flake.lib.${builtins.currentSystem} // flake.packages.${builtins.currentSystem}
             | ^
           14|

       … while evaluating the attribute 'defaultNix'
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:229:5:
          228|   rec {
          229|     defaultNix =
             |     ^
          230|       (builtins.removeAttrs result ["__functor"])

       … in the left operand of the update (//) operator
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:231:7:
          230|       (builtins.removeAttrs result ["__functor"])
          231|       // (if result ? defaultPackage.${system} then { default = result.defaultPackage.${system}; } else {})
             |       ^
          232|       // (if result ? packages.${system}.default then { default = result.packages.${system}.default; } else {});

       … while calling the 'removeAttrs' builtin
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:230:8:
          229|     defaultNix =
          230|       (builtins.removeAttrs result ["__functor"])
             |        ^
          231|       // (if result ? defaultPackage.${system} then { default = result.defaultPackage.${system}; } else {})

       … from call site
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:224:10:
          223|     else if lockFile.version >= 5 && lockFile.version <= 7
          224|     then allNodes.${lockFile.root}
             |          ^
          225|     else throw "lock file '${lockFilePath}' has unsupported version ${toString lockFile.version}";

       … while calling anonymous lambda
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:156:13:
          155|     builtins.mapAttrs
          156|       (key: node:
             |             ^
          157|         let

       … in the condition of the assert statement
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:211:13:
          210|           if node.flake or true then
          211|             assert builtins.isFunction flake.outputs;
             |             ^
          212|             result

       … while calling the 'isFunction' builtin
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:211:20:
          210|           if node.flake or true then
          211|             assert builtins.isFunction flake.outputs;
             |                    ^
          212|             result

       … while evaluating 'flake' to select 'outputs' on it
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:211:40:
          210|           if node.flake or true then
          211|             assert builtins.isFunction flake.outputs;
             |                                        ^
          212|             result

       … while calling the 'import' builtin
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:167:19:
          166|
          167|           flake = import (outPath + "/flake.nix");
             |                   ^
          168|

       … while calling the 'storePath' builtin
         at /nix/store/8kpx53qi52yhjai1vdw8zpa95iqa61bv-source/default.nix:115:15:
          114|               # If it's already a store path, don't copy it again.
          115|               builtins.storePath src
             |               ^
          116|             else

       error: 'builtins.storePath' is not allowed in pure evaluation mode
```

</p>
</details>

I examined the file it fails in and the issue [should have been fixed](https://github.com/edolstra/flake-compat/blob/ff81ac966bb2cae68946d5ed5fc4994f96d0ffec/default.nix#L114-L115) in more recent versions of `flake-compat`, so I tried to update the `pre-commit-hooks`'s `flake-compat` to a more recent version in the flake, but it still uses the old version.

Maybe it always uses the [version in the upstream lockfile](https://github.com/cachix/git-hooks.nix/blob/dcf5072734cb576d2b0c59b2ac44f5050b5eac82/default.nix#L4)?

eljamm: This could explain why [formatter packages are outdated](https://github.com/ngi-nix/ngipkgs/pull/748#discussion_r2033980570) in the non-flake hook. I can't seem to find a way to override its nixpkgs, though.

PR #773, opened 2025-04-09
Fix eval failures upon pkgs references in projects

OPNA2608:
Based on https://github.com/ngi-nix/ngipkgs/pull/681, so we actually have somewhere to cause this issue, and show that it's fixed.

Referencing `pkgs` in any project's `default.nix` caused alot of eval failures: https://buildbot.ngi.nixos.org/#/builders/5/builds/1524

The failure comes from trying to elaborate on the `system = null` from `classic'`:

https://github.com/ngi-nix/ngipkgs/blob/fe2fd0e1fe05b07ea41be8498ae237d6f1cdfde7/flake.nix#L34

To fix this, figure out where `classic'` gets references in this broken way, and switch it over to `classic` (which has a proper `system` value, and can thus be elaborated on & provide pkgs).


fricklerhandwerk:
Because some things simply don't depend on the system parameter (or any Nixpkgs configuration) and should be accessible as such. The `default.nix` also doesn't separate this clearly at the moment (I noted this elsewhere as another can of worms). I'm not opposed to altogether removing interfaces we're not actively supporting, we have no stability guarantees at the moment.

eljamm:
Although this isn't much of a fix, we can work around the issue with [something like this](https://github.com/ngi-nix/ngipkgs/pull/774/files#diff-8d002f6f2600ee2ca3bae82020c611eb5ca4cf9c2c68fa5301200779af5dba16R27-R40).

fricklerhandwerk:
Here's an idea how to refactor this to make invoking Nixpkgs a bit saner: https://github.com/fricklerhandwerk/ngipkgs/tree/nixpkgs-invocation

PR #776, opened 2025-04-09
Fix typos in naming of programs

themadbit:
- Add tests `import...args` declaration

### DIFF
--- a/templates/project/programs/_programName_/tests/basic.nix
+++ b/templates/project/programs/_programName_/tests/basic.nix
@@ -12,8 +12,8 @@
       {
         imports = [
           sources.modules.ngipkgs
-          sources.modules.services._programName_
-          sources.examples._ProjectName_._programName_
+          sources.modules.programs._programName_
+          sources.examples._ProjectName_._testName_  # i.e _ProjectName_.basic
         ];
       };
   };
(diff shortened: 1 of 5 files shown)


fricklerhandwerk (dismissed):
LGTM

themadbit:
Yeah, working to resolve the error

eljamm:
Yeah, I actually fixed this today, but I forgot to open a PR.

themadbit:
I didn't follow. The check for templates was failing because it was being treated as a project? :slightly_smiling_face:

eljamm:
Quite the contrary. It was failing because it was behaving differently than a project. It's a template **for** a project, after all.

That said, what we're testing now is just that the template adheres to the data types, but this can be expanded in the future to make a dummy test and example, if we need to.

templates/project/programs/_programName_/tests/basic.nix
```diff
@@ -12,8 +12,8 @@
       {
         imports = [
           sources.modules.ngipkgs
-          sources.modules.services._programName_
-          sources.examples._ProjectName_._programName_
+          sources.modules.programs._programName_
+          sources.examples._ProjectName_._testName_ # i.e _ProjectName_.basic
```
eljamm, line 16: This should be the name for the example, not the test. We also need to modify the one in services

```suggestion
          sources.examples._ProjectName_._exampleName_ # i.e _ProjectName_.basic
```

themadbit: ooh, yess! Done

PR #794, opened 2025-04-12
Canaille: projects-old -> projects 

victor-ok:
This PR migrates the project from the old to a new template


projects/Canaille/default.nix
```diff
@@ -0,0 +1,60 @@
...
+        test = {
+          text = "Usage example";
+          url = "https://gitlab.com/yaal/canaille/#locally";
+        };
+      };
+    };
+
+    Canaille-cli = null;
+  };
+
```
eljamm, line 38: Canaille only has a service, so we don't need this. Also delete the `./programs` folder

```suggestion
```

projects/Canaille/default.nix
```diff
@@ -0,0 +1,60 @@
...
+  nixos.modules.services = {
+    Canaille = {
+      name = "Canaille";
+      module = "${sources.inputs.nixpkgs}/nixos/modules/services/security/canaille.nix";
+      examples.basic = {
+        module = ./services/Canaille/examples/basic.nix;
+        description = "";
+        tests.basic = import ./services/Canaille/tests/basic.nix args;
```
eljamm, line 46: We don't have to write a new test since we already have one in nixpkgs. Also delete the test file.

```suggestion
        tests.canaille = "${sources.inputs.nixpkgs}/nixos/tests/canaille.nix";
```

projects/Canaille/default.nix
victor-ok: Resolved!

projects/Canaille/default.nix
victor-ok: Resolved

projects/Canaille/services/Canaille/examples/basic.nix
```diff
@@ -0,0 +1,5 @@
+{ ... }:
+
+{
+  services.canaille.enable = true;
```
eljamm, line 4: The tests are failing. If you click on `buildbot/nix-build` below, you'll see which ones exactly and why. In this case, we have:

```
       … while evaluating the option `services.canaille.settings.SERVER_NAME':
       (stack trace truncated; use '--show-trace' to show the full, detailed trace)
       error: The option `services.canaille.settings.SERVER_NAME' was accessed but has no value defined. Try setting the option.
```

I remember that the example was more elaborate than this. Why did you remove the other options?

projects/Canaille/services/Canaille/tests/basic.nix
eljamm, line 1: Let's remove this test since it's not doing anything.

projects/Canaille/services/Canaille/module.nix
eljamm, line 1: We have a service available from nixpkgs and we're using it:

```
      module = "${sources.inputs.nixpkgs}/nixos/modules/services/security/canaille.nix";
```

So we don't need to write another module. Please delete this file.

projects/Canaille/default.nix
victor-ok: you suggested that I delete the entire programs folder here and i did just that

projects/Canaille/services/Canaille/examples/basic.nix
victor-ok: are you suggesting that I pass the secret key created while I was testing?

victor-ok: I figured it was a typo and I made changes

projects/Canaille/default.nix
```diff
@@ -0,0 +1,36 @@
...
+{
+  metadata = {
+    summary = "Canaille is a zero-knowledge opinionated identity server.";
+    subgrants = [
+      "Canaille"
+    ];
+  };
+
+  nixos.modules.services = {
+    Canaille = {
```
eljamm, line 16: One last change, but this looks good otherwise.

```suggestion
    canaille = {
```

victor-ok: I have resolved, thank you!

eljamm (approved):
Looks good. Well done!

PR #816, opened 2025-04-14
Heads: Make allowlist configurable

OPNA2608:
This should allow controlling

1. on a package level, which boards will have their corresponding derivations generated
2. on a module level, which boards will have their built ROMs symlinked in the system

To make this even more permissive, I'm currently updating the `updateDepsScript` to stop ignoring board targets marked with `UNMAINTAINED` and `UNTESTED` so we can `lib.lists.intersectLists` against the full list of boards. When (if?) we eventually expand our allowlist, we would just filter them out in the default `allowedBoards` again. But when users request them to be build, then I think there's no reason to deny them that attempt. Whether or not it'll work would be their concern though :slightly_smiling_face:.

### DIFF
--- a/pkgs/by-name/heads/deps.nix
+++ b/pkgs/by-name/heads/deps.nix
@@ -45,6 +45,20 @@
     "t430-maximized"
     "t480-hotp-maximized"
     "t480-maximized"
+    "UNMAINTAINED_kgpe-d16_server"
+    "UNMAINTAINED_kgpe-d16_server-whiptail"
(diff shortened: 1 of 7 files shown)


fricklerhandwerk:
Please add an example for using that, and also it would be helpful to find in the documentation (e.g. an example or wherever) where that "fixed location" for the built files is.

OPNA2608:
Does this work?

fricklerhandwerk (approved):

PR #832, opened 2025-04-16
ntpd-rs: projects-old -> projects

### LINKED ISSUE #738: Migrate `ntpd-rs` from projects-old to projects
## Instructions

1. Copy the project template to the projects directory:

   ```
   cp -r templates/project projects/ntpd-rs
   ```

1. Search for `NGI Project: ntpd-rs` in the [Ngipkgs issues](https://github.com/ngi-nix/ngipkgs/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22NGI%20Project%22) page.
   If a page with that name exists, use the information available there in the next step.

1. Open the new project file and fill in the missing data, using the information in `projects-old/ntpd-rs/default.nix`.

   ```
   $EDITOR projects/ntpd-rs/default.nix
   ```

   **Note:** the code structure of `projects-old` is different from `projects`, so copying information without adhering to that structure will lead to test errors.

1. Check that the code is valid by running the test locally:

   ```
   # examples
   $ nix build .#checks.x86_64-linux.projects/ntpd-rs/nixos/examples/<example_name>

   # tests
   $ nix build .#checks.x86_64-linux.projects/ntpd-rs/nixos/tests/<test_name>
   ```

1. Delete the `projects-old/ntpd-rs` directory
1. Run the Nix code formatter with `nix fmt`
1. Commit your changes and [create a new PR](#how-to-create-pull-requests-to-ngipkgs)

themadbit:
Closes #738


projects/ntpd-rs/default.nix
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  nixos.modules.services = {
+    ntpd-rs = {
+      name = "ntpd-rs";
+      module = "${sources.inputs.nixpkgs}/nixos/modules/services/networking/ntp/ntpd-rs.nix";
+      examples.basic = {
+        module = ./services/ntpd-rs/examples/basic.nix;
+        description = ''
+          Example config for ntpd-rs options is available in tests.
```
eljamm, line 32: This is more of a comment than a description, right? You're saying that we should have the configuration in the tests as a standalone example, in the future.

themadbit: Not quite. I saw no need to rewrite examples, as the test expression shows how to use all options as made in the module.

eljamm: In this case, we can set the example to null and just have the test be in `nixos.tests`. What do you think?

themadbit: Yes, that is more ideal. I did attempt `module = null;` but I got errors like:
```
{
         description = ''
           Example config for ntpd-rs options is available in tests.
         '';
         module = null;
         tests = {
           ntpd-rs = "/nix/store/dn9smqqg60pn1a3c1sk2m0m8iakm84zs-source/nixos/tests/ntpd-rs.nix";
         };
       }: expected type option<example>, but value does not conform to 'example': expected 'example'-struct, but found:
       missing required either<path, string>[absolute path] field 'module'
```

Was my declaration incorrect?

eljamm: Each service must have an example and each example must have a module, but examples themselves can be null to signify that they're needed, but not available. In this case, you need `examples.basic = null;`

themadbit: aaah, got that. Let me commit something, and you have a look

projects/ntpd-rs/default.nix
```diff
@@ -0,0 +1,33 @@
...
+  };
+
+  nixos.modules.services = {
+    ntpd-rs = {
+      name = "ntpd-rs";
+      module = "${sources.inputs.nixpkgs}/nixos/modules/services/networking/ntp/ntpd-rs.nix";
+      examples.basic = null;
+      # See https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/tests/ntpd-rs.nix for examples
+    };
+  };
```
eljamm, line 32: It's possible to have tests that are not tied to examples:

```suggestion
  };
  
  nixos.tests.ntpd-rs = "${sources.inputs.nixpkgs}/nixos/tests/ntpd-rs.nix";
```

PR #857, opened 2025-04-17
lib: add moduleLocFromOptionString

### LINKED ISSUE #329: (re-)exposing upstream (nixpkgs) modules
Context: https://github.com/ngi-nix/ngipkgs/pull/305#discussion_r1654628524

TLDR:

Currently, in ngipkgs, we are re-exposing nixpkgs modules like this:

```nix
modules.services.forgejo = "${sources.inputs.nixpkgs}/nixos/modules/services/misc/forgejo.nix";
```

But the ideal way to expose those modules would be this:

```nix
modules.services.forejo = nixos.modules.services.forejo;
```

Proposed solution:

We can call `evalModules` on https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/module-list.nix (AFAIK this is where all default NixOS modules are imported from) and get `options` on the result, finally, use `unsafeGetAttrPos` to get the file location.

Am I missing anything? I want to double check here before I go on and implement this

stepbrobd:
closes #329

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -36,6 +36,45 @@ rec {
           if lib.isAttrs value then f (path + name + separator) value else { ${path + name} = value; };
       in
       f "";
+
+    # get the path of NixOS module from string
+    # example:
+    # lib'.moduleLocFromOptionString "services.ntpd-rs"
+    # => "/nix/store/...-source/nixos/modules/services/networking/ntp/ntpd-rs.nix"
+    moduleLocFromOptionString =
+      let
+        inherit
+          (lib.evalModules {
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
Cool, thanks a lot! LGTM

stepbrobd:
In theory there's a more ergonomic approach (but error message will be very bad if anything wrong happens):

Instead of exposing a function for users to call, this can simply be exported as an attrset that contain option name to file mapping.

IMO this is already ergonomic enough for me 😂

PR #861, opened 2025-04-18
Separate nixpkgs invocation

eljamm:
Supersedes https://github.com/ngi-nix/ngipkgs/pull/773

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,313 +1,354 @@
+let
+  sources = (import ./flake-compat.nix { root = ./.; }).inputs;
+in
 {
-  sources ? (import ./flake-compat.nix { root = ./.; }).inputs,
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -1,313 +1,356 @@
...
+
+      # $ nix-build . -A demo-test
+      # $ ./result
+      demo-test = demo ./projects/Cryptpad/demo.nix;
+
+      # TODO: find a better place for this
+      metrics = with lib; {
+        projects = attrNames raw-projects;
+        in-ngipkgs = attrNames ngipkgs;
+        derivations = concatMap (p: attrNames p.packages) (attrValues raw-projects);
```
eljamm, line 243: Should we remove this, since we're no longer tracking packages?

Issue #622, opened 2025-03-19
NGI Project: Briar


OluchiTheAnalyst:
### Short Summary

Briar is a messaging app designed for activists, journalists, and anyone else who needs a safe, easy and robust way to communicate.
Unlike traditional messaging apps, Briar doesn't rely on a central server - messages are synchronized directly between the users' devices.
If the Internet's down, Briar can sync via Bluetooth or Wi-Fi, keeping the information flowing in a crisis. If the Internet's available, Briar can sync via the Tor network, protecting users and their relationships from surveillance.

### NLnet page(s)

- https://nlnet.nl/project/Briar/

### Website

https://briarproject.org/

### Source repository

- https://code.briarproject.org/briar/briar

### Metadata

- Language/Framework: Java

- Build tool: Gradle

- Development environment: N/A

### Documentation

- Usage Examples:
   + https://briarproject.org/manual/
- Build from source/Development:
  + [Build Process](https://code.briarproject.org/briar/briar/-/wikis/home#build-process)
  + [Development](https://code.briarproject.org/briar/briar/-/wikis/home#development)
- Other:
  + [Wiki](https://code.briarproject.org/briar/briar/-/wikis/home)
  + [Installation](https://briarproject.org/fdroid.html)
  + [Reproducible builds](https://code.briarproject.org/briar/briar-reproducer#briar-reproducer)
  + [User manual](https://briarproject.org/manual/)

### Artefacts

- CLI:
    - Documentation: https://briarproject.org/manual/
    - Tests: https://code.briarproject.org/briar/briar/-/wikis/testing    
- GUI:
   - Mobile Apps: https://briarproject.org/fdroid.html

### Nixpkgs/NixOS

   - Packages:
     - [Briar-Desktop](https://search.nixos.org/packages?channel=24.11&show=briar-desktop&from=0&size=50&sort=relevance&type=packages&query=briar)
 

### Extra Information

[Already use Nixpkgs](https://search.nixos.org/packages?channel=24.11&show=briar-desktop&from=0&size=50&sort=relevance&type=packages&query=briar) for packaging

ta1hia:
hello! i'm taking a stab at adding the project metadata for Briar in https://github.com/ngi-nix/ngipkgs/pull/867

PR #867, opened 2025-04-18
Add Briar metadata

ta1hia:
Add project metadata for Briar

https://github.com/ngi-nix/ngipkgs/issues/622


ta1hia:
hey @eljamm! i'm trying to add project metadata for Briar but i hit an issue and i'm wondering if you could help me troubleshoot why i'm having a the template's `basic.nix` fail for one specific target.

it looks like the test passes for `x86_64` ([buildbot link](https://buildbot.ngi.nixos.org/#/builders/708/builds/1)) but not `aarch64_linux` ([buildbot link](https://buildbot.ngi.nixos.org/#/builders/709/builds/1)). 

according to the nix error in buildbot for that test, it looks like the briar package isn't supported for aarch64_linux. is there a way to have buildbot skip that test?

eljamm:
@ta1hia Instead of skipping the test, you can add the package only on supported platforms by doing [something like this](https://github.com/ngi-nix/ngipkgs/blob/1d1a5452db53757c233a7e79b3b775fe2bf71264/projects/Icestudio/programs/Icestudio/module.nix#L15-L21).

projects/Briar/programs/briar/tests/basic.nix
eljamm, line 1: Let's just delete this file, since it's not testing anything. We also want to set the attribute to `tests.basic = null` to indicate that it's desired, but unavailable or unimplemented, yet

ta1hia:
@eljamm thank you so much for the guidance! i've applied your suggestions and marked the PR ready for review.

eljamm (approved):
Looks good. Nice job, @ta1hia!

Issue #797, opened 2025-04-13
NGI Project: CNSPRCY


hashaam-011:
### Short Summary

CNSPRCY is an open-source project that tightly integrates personal computing devices (desktops, laptops, phones) via a replicated, eventually-consistent database, enabling encrypted messaging and direct device-to-device connections without third-party reliance.

### NLnet page(s)

https://nlnet.nl/project/CNSPRCY
https://nlnet.nl/project/CNSPRCY-Android

### Website

https://cnsprcy.network

### Source repository

https://github.com/ConspiracyLabs/conspiracy
Language/Framework: Rust
Dependency Management: Cargo
Nix Development Environment: None in upstream repo

### Documentation

Usage Examples:
None explicitly listed; usage described at https://cnsprcy.network/
Build from Source/Development:
https://github.com/ConspiracyLabs/conspiracy#building
Other:
Roadmap: https://cnsprcy.network/roadmap

### Artefacts

CLI:
conspiracy:
Documentation: https://github.com/ConspiracyLabs/conspiracy#building
Examples: None
Tests: https://github.com/ConspiracyLabs/conspiracy/tree/main/tests
GUI: None
Services/Daemons:
conspiracy (background service):
Documentation: https://github.com/ConspiracyLabs/conspiracy#running
Libraries: None
Extensions: None
Mobile Apps: in development
Documentation: None
Examples: None
Tests: None

### Nixpkgs/NixOS

- Packages:
None
- Services:
None

### Extra Information

Not yet package in Nixpkgs

eljamm:
Please fix the formatting (see [comment](https://github.com/ngi-nix/ngipkgs/issues/799#issuecomment-2800714070))

hashaam-011:
@eljamm Please check now.

eljamm:
- The https://cnsprcy.network/ website doesn't exist. Where did you find that?
- The source repository is wrong and it doesn't exist: https://github.com/ConspiracyLabs/conspiracy
- There is no subgrant for https://nlnet.nl/project/CNSPRCY-Android
- None of the triaged information is valid ...

Did you use LLMs to generate this @hashaam-011?

hashaam-011:
I might overlooked this one. I will fix.

eljamm:
[...]

PR #870, opened 2025-04-18
Adds CNSPRCY service

ta1hia:
Adds a service module for CNSPRCY and a more fleshed out basic test. I also went ahead and reorganized the CNSPRCY project directory to follow the [project template](https://github.com/ngi-nix/ngipkgs/tree/main/maintainers/templates/project).

#797

### DIFF
--- a/projects/CNSPRCY/default.nix
+++ b/projects/CNSPRCY/default.nix
@@ -3,20 +3,38 @@
   pkgs,
   sources,
 }@args:
+
 {
   metadata = {
+    summary = "E2EE connections between trusted devices for establishing private group chats.";
     subgrants = [
       "CNSPRCY"
     ];
+    links = {
+      source = {
+        text = "CNSPRCY source code";
+        url = "https://git.sr.ht/~xaos/cnsprcy";
+      };
+    };
   };
 
-  # https://git.sr.ht/~xaos/cnsprcy/tree/master/item/src/config.rs
   nixos.modules.programs.cnsprcy = {
-    module = ./module.nix;
-    examples.cnsprcy = {
-      module = ./example.nix;
-      description = "";
-      tests.basic = import ./test.nix args;
+    name = "cnsprcy";
+    module = ./programs/cnsprcy/module.nix;
+    examples.basic = {
+      module = ./programs/cnsprcy/examples/basic.nix;
+      description = "Checks for cnspr executable";
+      tests.basic = import ./programs/cnsprcy/tests/basic.nix args;
+    };
+  };
+
+  nixos.modules.services.cnsprcy = {
+    name = "cnsprcy";
+    module = ./services/cnsprcy/module.nix;
+    examples.basic = {
+      module = ./services/cnsprcy/examples/basic.nix;
+      description = "Checks that cnsprcy systemd service is running";
+      tests.basic = import ./services/cnsprcy/tests/basic.nix args;
     };
   };
 }
--- a/projects/CNSPRCY/programs/cnsprcy/module.nix
+++ b/projects/CNSPRCY/programs/cnsprcy/module.nix
@@ -1,7 +1,7 @@
 {
+  config,
   lib,
   pkgs,
-  config,
   ...
 }:
 let
--- a/projects/CNSPRCY/services/cnsprcy/module.nix
+++ b/projects/CNSPRCY/services/cnsprcy/module.nix
@@ -1,0 +1,86 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.cnsprcy;
+in
+{
+  options.services.cnsprcy = {
+    enable = lib.mkEnableOption "cnsprcy";
+    package = lib.mkPackageOption pkgs "cnsprcy" { };
+
+    hostname = lib.mkOption {
+      type = lib.types.str;
+      description = "Hostname of CNSPRCY server";
+      default = config.networking.hostName;
+    };
+
+    stateDir = lib.mkOption {
+      type = lib.types.str;
+      description = "State directory for CNSPRCY server";
+      default = "/var/lib/cnsprcy";
+    };
+
+    user = lib.mkOption {
+      type = lib.types.str;
+      description = "Username of the system user that should own files and services related to CNSPRCY.";
+      default = "cnsprcy";
+    };
+    group = lib.mkOption {
+      type = lib.types.str;
+      description = "Group that contains the system user that executes CNSPRCY.";
+      default = "cnsprcy";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+
+    environment.systemPackages = [ cfg.package ];
+
+    users.users."${cfg.user}" = {
+      isSystemUser = true;
+      group = cfg.group;
+
+      # CNSPRCY looks for the user home to initialize its data directory
+      # at $stateDir/.local/share/cnsprcy, but systemd.tmpfiles needs to create
+      # the dir otherwise permissions won't be set correctly.
+      home = cfg.stateDir;
+      useDefaultShell = true;
+    };
+    users.groups."${cfg.group}" = { };
+
+    systemd.tmpfiles.rules = [
+      "d ${cfg.stateDir}/.local/share/cnsprcy 0700 ${cfg.user} ${cfg.group} -"
+      "d ${cfg.stateDir}/.local/share/cnsprcy/handlers 0700 ${cfg.user} ${cfg.group} -"
+    ];
+
+    systemd.services.cnsprcy = {
+      description = "CNSPRCY service";
+      wantedBy = [ "multi-user.target" ];
+      after = [
+        "nss-user-lookup.target"
+        "systemd-tmpfiles-setup.service"
+      ];
+
+      # Runs interactive config initialization to set machine name, generate
+      # conspirator keys and initial CNSPRCY db. TODO use the config init flags
+      # instead, but will need to investigate how to feed args to `cnspr serve`.
+      #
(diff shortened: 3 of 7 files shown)


projects/CNSPRCY/programs/cnsprcy/module.nix
```diff
@@ -10,10 +10,11 @@ in
 {
   options.programs.cnsprcy = {
     enable = lib.mkEnableOption "cnsprcy";
-    package = lib.mkPackageOption pkgs "cnsprcy" { };
```
eljamm, line 13: If a program has a single executable, it's idiomatic to store this in a single option, because this will allow users to change and override it more easily.

projects/CNSPRCY/default.nix
```diff
@@ -3,20 +3,39 @@
...
+    summary = "E2EE connections between trusted devices for establishing private group chats.";
     subgrants = [
       "CNSPRCY"
     ];
+    links = {
+      exampleLink = {
+        text = "Title";
+        url = "<URL>";
+      };
+    };
```
eljamm, line 18: We should remove this if we don't have any links

```suggestion
```

projects/CNSPRCY/programs/cnsprcy/module.nix
ta1hia: whoops, i accidentally deleted this line when porting CNSPRCY over to the project template. but i also didn't know what this options purpose was in the first place so thank you for the explanation @eljamm!

projects/CNSPRCY/services/cnsprcy/module.nix
```diff
@@ -0,0 +1,70 @@
...
+      description = "Group that contains the system user that executes CNSPRCY.";
+      default = "cnsprcy";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+
+    environment.systemPackages = with pkgs; [
+      cnsprcy
+    ];
```
eljamm, line 37: This should go into its own option, like we discussed for the program module.

projects/CNSPRCY/default.nix
```diff
@@ -3,20 +3,38 @@
...
-    module = ./module.nix;
-    examples.cnsprcy = {
-      module = ./example.nix;
-      description = "";
-      tests.basic = import ./test.nix args;
+    name = "cnsprcy";
+    module = ./programs/cnsprcy/module.nix;
+    examples.basic = {
+      module = ./programs/cnsprcy/examples/basic.nix;
+      description = "Checks for cnspr executible";
```
eljamm, line 26:
```suggestion
      description = "Checks for cnspr executable";
```

projects/CNSPRCY/services/cnsprcy/module.nix
```diff
@@ -0,0 +1,70 @@
...
+    systemd.tmpfiles.rules = [
+      "d ${libDir}/.local/share/cnsprcy 0700 ${cfg.user} ${cfg.group} -"
+      "d ${libDir}/.local/share/cnsprcy/handlers 0700 ${cfg.user} ${cfg.group} -"
+    ];
+
+    systemd.services.cnsprcy = {
+      description = "CNSPRCY service";
+      wantedBy = [ "multi-user.target" ];
+      after = [ "nss-user-lookup.target" ];
+      preStart = "printf 'n\n${cfg.hostname}\ny\n' | ${pkgs.cnsprcy}/bin/cnspr config init";
```
eljamm, line 60: So is `hostname` only used to be printed in the service logs? I don't think that's necessary as options need to be something that change the service's behaviour rather than for debugging. Although we may have options that control the service's own debugging settings.

projects/CNSPRCY/services/cnsprcy/module.nix
```diff
@@ -0,0 +1,70 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.cnsprcy;
+  libDir = "/var/lib/cnsprcy";
```
eljamm, line 9: This could also be an option, maybe. Although this would usually be called [stateDir](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=).

eljamm (commented):
This is shaping up nicely. Is it still a draft, @ta1hia, or is it ready for review?

projects/CNSPRCY/services/cnsprcy/module.nix
ta1hia: @eljamm the `cnspr config init` command kicks off an interactive prompt for generating the initial CNSPRCY config file and the printf feeds some responses to the prompt, including setting an identifier for the current machine. 

i considered constructing the config from scratch during options setting, but i hit the need for interactively generating keys using the `cnspr` command later down the road anyway so this approach seemed less finicky to me just to get things going. let me know if you think otherwise!

regardless i should at the very least add a comment here!

projects/CNSPRCY/services/cnsprcy/module.nix
```diff
@@ -0,0 +1,69 @@
...
+in
+{
+  options.services.cnsprcy = {
+    enable = lib.mkEnableOption "cnsprcy";
+    package = lib.mkPackageOption pkgs "cnsprcy" { };
+
+    hostname = lib.mkOption {
+      type = lib.types.str;
+      description = "Hostname of CNSPRCY server";
+      default = "machine1";
```
ta1hia, line 19: maybe i can pull the machine's hostname here?

projects/CNSPRCY/services/cnsprcy/module.nix
```diff
@@ -0,0 +1,73 @@
...
+
+    systemd.tmpfiles.rules = [
+      "d ${stateDir}/.local/share/cnsprcy 0700 ${cfg.user} ${cfg.group} -"
+      "d ${stateDir}/.local/share/cnsprcy/handlers 0700 ${cfg.user} ${cfg.group} -"
+    ];
+
+    systemd.services.cnsprcy = {
+      description = "CNSPRCY service";
+      wantedBy = [ "multi-user.target" ];
+      after = [ "nss-user-lookup.target" ];
```
eljamm, line 58: The service should probably only start after the tmpfiles have been created:

```suggestion
      after = [
        "nss-user-lookup.target"
        "systemd-tmpfiles-setup.service"
      ];
```

projects/CNSPRCY/services/cnsprcy/module.nix
```diff
@@ -0,0 +1,73 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.cnsprcy;
+  stateDir = "/var/lib/cnsprcy";
```
eljamm, line 9: This could also be an option

projects/CNSPRCY/services/cnsprcy/module.nix
eljamm: > @eljamm the cnspr config init command kicks off an interactive prompt for generating the initial CNSPRCY config file and the printf feeds some responses to the prompt, including setting an identifier for the current machine.

Ah, I see. That's a really clever solution. From what I've seen, we can also use the `config init` flags to directly set the hostname with `-n` and we can even choose where to put the database file with `-p`, so we can put it directly under `stateDir`:

```
Usage: cnspr config init [OPTIONS]

Options:
  -i, --id <ID>      ID to use for this node (omit to generate a new ID)
  -n, --name <NAME>  Name for this node
  -p, --path <PATH>  Where to store the new database
  -h, --help         Print help
```

Don't know how to make the `serve` command use it, though.

> i considered constructing the config from scratch during options setting, but i hit the need for interactively generating keys using the cnspr command later down the road anyway so this approach seemed less finicky to me just to get things going. let me know if you think otherwise!

Yeah, it's fine to dynamically create it like this for now. We can iterate on the module later on and generate the config using options. It should be fairly easy to do since the [config file](https://git.sr.ht/~xaos/cnsprcy/tree/master/item/doc/config.md) uses the TOML format, which is supported [in nixpkgs](https://github.com/NixOS/nixpkgs/blob/c6b82a914404228b9a4a4046c0a77c4b395b35f0/pkgs/pkgs-lib/formats.nix#L389) (also see how it's [used in modules](https://github.com/search?q=repo%3ANixOS%2Fnixpkgs%20pkgs.formats.toml&type=code)).

ta1hia:
@eljamm thank you for the review! Regarding the config initialization/generation discussion, I included a TODO comment with next step suggestions but left the preStart approach as-is for this PR. I applied all the other suggestions you made. lmk if you have any other notes!

projects/CNSPRCY/services/cnsprcy/module.nix
```diff
@@ -0,0 +1,86 @@
...
+in
+{
+  options.services.cnsprcy = {
+    enable = lib.mkEnableOption "cnsprcy";
+    package = lib.mkPackageOption pkgs "cnsprcy" { };
+
+    hostname = lib.mkOption {
+      type = lib.types.str;
+      description = "Hostname of CNSPRCY server";
+      default = builtins.getEnv "HOSTNAME";
```
eljamm, line 18: I feel like [networking.hostName](https://search.nixos.org/options?channel=unstable&show=networking.hostName&from=0&size=50&sort=relevance&type=packages&query=networking.hostName) is more appropriate than reading an env var, here

```suggestion
      default = config.networking.hostName;
```

eljamm (commented):
This is looking good. Could you please clean up the history, next? You can either rebase or just `git reset main` and make new commits.

eljamm (approved):
Hey @ta1hia, sorry for taking so long to get back to you on this and thanks for your great efforts!

PR #876, opened 2025-04-21
models: add subgrant type

eljamm:
As noted in https://github.com/ngi-nix/summer-of-nix/issues/183#issuecomment-2815906425

### DIFF
--- a/projects/models.nix
+++ b/projects/models.nix
@@ -24,6 +24,13 @@ let
     any
     ;
 
+  subgrantType = struct "subgrant" {
+    Commons = option (list string);
+    Core = option (list string);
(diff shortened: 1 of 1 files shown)


projects/models.nix
```diff
@@ -24,6 +24,12 @@ let
     any
     ;
 
+  subgrantType = struct "subgrant" {
```
eljamm, line 27: Should we add `Commons` as well? It's missing from the GitHub labels.

fricklerhandwerk: Yes!

fricklerhandwerk (approved):

PR #877, opened 2025-04-21
metrics: move to maintainers

eljamm:
## Changes

- Move metrics to `maintainers/`
- Cleanup


maintainers/metrics.nix
```diff
@@ -0,0 +1,84 @@
...
+    mapAttrs
+    count
+    optionalAttrs
+    ;
+in
+rec {
+  metrics = {
+    projects = attrNames raw-projects;
+    in-ngipkgs = attrNames ngipkgs;
+    derivations = concatMap (p: attrNames p.packages) (attrValues raw-projects);
```
eljamm, line 21: Should we remove `derivations`, since we're no longer tracking `packages`?

fricklerhandwerk: Let's keep it, it's the only metric that stayed consistent since the beginning, and while it's a weak proxy it still gives us an indication of how complex a project is.

fricklerhandwerk (approved):

Issue #544, opened 2025-03-06
US: Test coverage for module options


fricklerhandwerk:
As a contributor I want to know which components of a project are covered by examples, and thus, tests, in order to know what's left to be done. As a user I want to know the same in order to get an indicator of reliability of examples.

## Acceptance criteria

Given I have some project with an example,
When the example has a test,
Then it shows as "tested". Otherwise not.

## Implementation notes

As of https://github.com/ngi-nix/ngipkgs/pull/507 we're doing a bit of a hack by attaching examples to application components, which allow us to do rudimentary tracing. Eventually we may want something more granular though.

imincik:
It might be hard and expensive to implement.

PR #880, opened 2025-04-21
flake.nix: don't check examples

eljamm:
This is mostly legacy code which is no longer needed since examples are checked by tests.


fricklerhandwerk (approved):
We just mustn't forget to indicate testedness of examples in the overview @erictapen

eljamm:
> We just mustn't forget to indicate testedness of examples in the overview @erictapen

Does the https://github.com/ngi-nix/ngipkgs/issues/544 user story suffice for this or is its scope too large?

PR #881, opened 2025-04-21
demo: move to `overview`

eljamm:


default.nix
```diff
@@ -227,89 +227,12 @@ rec {
...
-  demo =
-    module:
-    pkgs.writeShellScript "demo-vm" ''
-      exec ${(demo-system module).config.system.build.vm}/bin/run-nixos-vm "$@"
-    '';
-
-  # $ nix-build . -A demo-test
-  # $ ./result
-  demo-test = demo ./projects/Cryptpad/demo.nix;
+  demo = import ./infra/demo {
```
fricklerhandwerk, line 230: Hm, that's a tough one. I don't think this has anything to do with infra. How about putting it in overview? Maybe that directory is a misnomer (it's been called that way because it simply lists all applications), but it's our UI and that's where the demo will show up.

Another options would be the maintainers directory where all we currently store are some instructions for using the cache and remote builder. One could argue that running the local demo can also be part of the contributor workflow.

fricklerhandwerk (dismissed):

fricklerhandwerk (approved):

PR #890, opened 2025-04-22
Remove `maintainers.nix`

eljamm:


maintainers.nix
eljamm, line 1: Alternatively, we could move it to `maintainers/`

fricklerhandwerk: We can put it back if we need it

fricklerhandwerk (approved):

PR #893, opened 2025-04-22
Remove `lib.nix`

eljamm:


default.nix
```diff
@@ -8,23 +8,32 @@
...
-    lib
     pkgs
     system
     sources
     ;
 
-  # TODO: we should be exporting our custom functions as `lib`, but refactoring
-  # this to use `pkgs.lib` everywhere is a lot of movement
-  lib' = import ./lib.nix { inherit lib; };
+  lib = lib' // {
```
fricklerhandwerk, line 24: Hm I don't think we should re-export *everything*, just our custom stuff

fricklerhandwerk: Yes, something like that. Since we have only one (?) function in `lib'` that should be alright, and correctly signify we're doing something fancy.

Issue #338, opened 2024-08-14
NGI Project: Holo


fricklerhandwerk:
### Short Summary

A novel routing stack in Rust, including IS-IS routing

### NLnet page(s)

- https://nlnet.nl/project/HoloRouting

### Website

https://github.com/holo-routing/

### Source repository

- https://github.com/holo-routing/holo
- https://github.com/holo-routing/holo-cli
  - Language/Framework: Rust, Docker
  - Dependency management: Cargo
  - Nix development environment: n/a

### Documentation

- Usage Examples:
  - https://github.com/holo-routing/holo/wiki
- Build from source/Development:
  - https://github.com/holo-routing/holo/blob/master/INSTALL.md
  - NB: **Building it from the source code requires a nightly version of the Rust compiler**
- Dependencies:
  - `build-essential` `cmake` `libpcre2-dev` `protobuf-compiler`

### Artefacts

- CLI:
  - [holo-cli](https://github.com/holo-routing/holo-cli): 
    - documentation: https://github.com/holo-routing/holo/wiki/CLI
- Services/daemons:
  - [holo-daemon](https://github.com/holo-routing/holo) 

### Nixpkgs/NixOS

- Packages:
  - [holo-cli](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/pkgs/by-name/ho/holo-cli/package.nix#L39)
- Services:
  - [holo-daemon](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/pkgs/by-name/ho/holo-daemon/package.nix#L53)

### Extra Information

This may be a small-ish packaging effort, but will need distributed testing and careful integration into NixOS.

themadbit:
I'm triaging this project. I'll start packaging there after.

fricklerhandwerk:
@themadbit the next step would be adding a stub to  https://github.com/ngi-nix/ngipkgs/tree/main/projects. And since this is a process we'll run through a couple of times, in order not to lose track it would be good to have an issue template for translating triage data into in-tree metadata (cc @eljamm)

PR #919, opened 2025-04-23
holo-cli: init at 0.4.0-unstable-2025-04-01

themadbit:
WIP for #338


pkgs/by-name/holo-cli/default.nix
eljamm, line 1: Only [package.nix and dream2.nix](https://github.com/ngi-nix/ngipkgs/blob/bbf2cdaa6ba3d6c6f576d188df34cc8c508292ab/pkgs/by-name/default.nix#L60-L68) files are called, so this entire file isn't used and we should therefore delete it. That aside, there are things that we don't want to do, here.

pkgs/by-name/holo-cli/default.nix
```diff
@@ -0,0 +1,19 @@
+{
+  pkgs ? import <nixpkgs> {
```
eljamm, line 2: This [nixpkgs lookup path](https://nix.dev/guides/best-practices.html#lookup-paths) might be different on two machines, which is not reproducible. We already have `pkgs` available, which uses nixpkgs from NGIpkgs' flake and we can just use that, instead.

```suggestion
  pkgs,
```

pkgs/by-name/holo-cli/default.nix
```diff
@@ -0,0 +1,19 @@
...
+    overlays = [
+      (import (fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz"))
+    ];
```
eljamm, line 5: If we wanted to, we could either apply the overlay directly to [our pkgs](https://github.com/ngi-nix/ngipkgs/blob/bbf2cdaa6ba3d6c6f576d188df34cc8c508292ab/default.nix#L8) or even create a new set of pkgs with the overlay applied, but we don't really need to use this at all.

We also want to avoid importing random inputs and instead add them to our [flake inputs](https://github.com/ngi-nix/ngipkgs/blob/bbf2cdaa6ba3d6c6f576d188df34cc8c508292ab/flake.nix#L4-L17), both to keep track of what we use and to properly pin its version.

PS: Always prefer [pinning the versions](https://nix.dev/reference/pinning-nixpkgs.html) whenever you're fetching something. Fetching the latest one is bad for reproducibility as things might break upstream.

pkgs/by-name/holo-cli/package.nix
```diff
@@ -0,0 +1,48 @@
+{
+  lib,
+  stdenv,
```
eljamm, line 3: Unused

```suggestion
```

pkgs/by-name/holo-cli/package.nix
```diff
@@ -0,0 +1,48 @@
...
+  fetchFromGitHub,
+  cmake,
+  pkg-config,
+  protobuf,
+  pcre2,
+}:
+
+rustPlatform.buildRustPackage (finalAttrs: {
+  pname = "holo-cli";
+  version = "0.4.0";
```
eljamm, line 14: [Versioning](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#versioning)

```suggestion
  version = "0.4.0-unstable-2025-03-31";
```

pkgs/by-name/holo-cli/package.nix
```diff
@@ -0,0 +1,48 @@
...
+
+  cargoLock = {
+    lockFile = ./Cargo.lock;
+  };
+
+  src = fetchFromGitHub {
+    owner = "holo-routing";
+    repo = "holo-cli";
+    rev = "ebec1a13a3ddf540bca32fa928ac803a538fec8d";
+    sha256 = "sha256-AQOGq1IUt6oXadzCtqxt8YGX0Va0pby5pEp2xdBuPeI=";
```
eljamm, line 24:
```suggestion
    hash = "sha256-AQOGq1IUt6oXadzCtqxt8YGX0Va0pby5pEp2xdBuPeI=";
```

pkgs/by-name/holo-cli/package.nix
```diff
@@ -0,0 +1,48 @@
...
+
+  # Use rust nightly features
+  RUSTC_BOOTSTRAP = 1;
+
+  nativeBuildInputs = [
+    cmake
+    pkg-config
+    protobuf
+  ];
+  buildInputs = [ pcre2 ];
```
eljamm, line 35: Just a formatting nit

```suggestion
  nativeBuildInputs = [
    cmake
    pkg-config
    protobuf
  ];

  buildInputs = [
    pcre2
  ];
```

pkgs/by-name/holo-cli/package.nix
```diff
@@ -0,0 +1,48 @@
...
+
+  useFetchCargoVendor = true;
```
eljamm, line 37: Useless since we're using the cargo lock file.

```suggestion
```

pkgs/by-name/holo-cli/package.nix
```diff
@@ -0,0 +1,48 @@
...
+
+  postPatch = ''
+    ln -s ${./Cargo.lock} Cargo.lock
+  '';
+
+  meta = with lib; {
+    description = "holo CLI";
+    license = licenses.mit;
+    maintainers = with maintainers; [ ];
+  };
```
eljamm, line 47: Avoid using `with lib;` ([see why](https://nix.dev/guides/best-practices.html#with-scopes))

```suggestion
  meta = {
    description = "Holo` Command Line Interface";
    homepage = "https://github.com/holo-routing/holo-cli";
    license = lib.licenses.mit;
    mainProgram = "holo-cli";
  };
```

pkgs/by-name/holo-cli/package.nix
```diff
@@ -0,0 +1,48 @@
...
+  pcre2,
+}:
+
+rustPlatform.buildRustPackage (finalAttrs: {
+  pname = "holo-cli";
+  version = "0.4.0";
+
+  cargoLock = {
+    lockFile = ./Cargo.lock;
+  };
```
eljamm, line 18: Just a nit, but I think things would be better organized if you move this under `src`

PS: we can also raise an issue, upstream, about [including a Cargo.lock file](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control)

themadbit: Yes. I've opened an issue on adding the lock file.
I had also left a comment regarding the release tags, as the versioning seems stable.

themadbit: Btw, if I haven't mentioned, thank you so much for leaving links when you leave a review. It's so kind

eljamm:
Looks good, you only need to squash the commits and change the commit and PR titles to fit the [Nixpkgs conventions](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#commit-conventions).

eljamm (approved):
LGTM

PR #922, opened 2025-04-23
Cleanup/projects

eljamm:


eljamm:
Closing since this is pretty old with a big scope and no clear goal.

PR #937, opened 2025-04-25
holod: init at 0.7.0

themadbit:
WIP for #338

### DIFF
--- a/pkgs/by-name/holod/package.nix
+++ b/pkgs/by-name/holod/package.nix
@@ -1,0 +1,56 @@
+{
+  lib,
+  rustPlatform,
+  fetchFromGitHub,
+  cmake,
+  pkg-config,
+  protobuf,
+  pcre2,
+}:
+
+rustPlatform.buildRustPackage (finalAttrs: {
+  pname = "holod";
+  version = "0.7.0";
+
+  src = fetchFromGitHub {
+    owner = "holo-routing";
+    repo = "holo";
+    tag = "v${finalAttrs.version}";
+    hash = "sha256-wASY+binAflxaXjKdSfUXS8jgdEHjdIF3AOzjN/a1Fo=";
+  };
+
+  cargoHash = "sha256-5X6a86V3Y9+KK0kGbS/ovelqXyLv15gQRFI7GhiYBjY=";
+
+  # Use rust nightly features
+  RUSTC_BOOTSTRAP = 1;
+
+  nativeBuildInputs = [
+    cmake
+    pkg-config
+    protobuf
+  ];
+  buildInputs = [
+    pcre2
+  ];
+
+  # Might not be needed if latest nightly compiler version is used
+  preConfigure = ''
+    # Find all lib.rs and main.rs files and add required unstable features
+    # Add the feature flag at the top of the file if not present`
+    find . -name "lib.rs" -o -name "main.rs" | while read -r file; do
+
+      for feature in extract_if let_chains hash_extract_if; do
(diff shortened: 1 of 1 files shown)


pkgs/by-name/holod/package.nix
```diff
@@ -0,0 +1,78 @@
...
+    runHook preInstall
+
+    mkdir -p $out/bin
+    cp target/release/holod $out/bin/
+
+    runHook postInstall
+  '';
+
+  meta = {
+    description = "`holo` daemon that provides the routing protocols, tools and policies.";
```
eljamm, line 73:
```suggestion
    description = "`holo` daemon that provides the routing protocols, tools and policies";
```

pkgs/by-name/holod/package.nix
```diff
@@ -0,0 +1,78 @@
...
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/bin
+    cp target/release/holod $out/bin/
+
+    runHook postInstall
+  '';
+
```
eljamm, line 71: I don't think we need to overwrite these phases. The [cargo hooks](https://github.com/NixOS/nixpkgs/tree/67495f6797d5fd8161f6c7109b673a429c82e728/pkgs/build-support/rust/hooks) will take care of this.

```suggestion
```

pkgs/by-name/holod/package.nix
```diff
@@ -0,0 +1,78 @@
...
+      fi
+
+      if ! grep -q "feature.*let_chains" "$file"; then
+        sed -i '1i #![feature(let_chains)]' "$file"
+      fi
+      
+      if ! grep -q "feature.*hash_extract_if" "$file"; then
+        sed -i '1i #![feature(hash_extract_if)]' "$file"
+      fi
+
```
eljamm, line 53:
```suggestion
      for feature in extract_if let_chains hash_extract_if; do
        if ! grep -q "feature.*$feature" "$file"; then
          sed -i "1i #![feature($feature)]" "$file"
        fi
      done
```

pkgs/by-name/holod/package.nix
```diff
@@ -0,0 +1,78 @@
...
+    protobuf
+  ];
+  buildInputs = [
+    pcre2
+  ];
+
+  # Might not be needed if latest nightly compiler version is used
+  preConfigure = ''
+    # Find all lib.rs and main.rs files and add required unstable features
+    # Add the feature flag at the top of the file if not present`
```
eljamm, line 39: Do we want/need to add this to all files? or is it enough to add this to specific ones?

pkgs/by-name/holod/package.nix
themadbit: My first build was failing due to the use of `--offline` and other default build options in the build phase. The build documentation also explicitly needs it to be built using the `--release` tag only? https://github.com/holo-routing/holo/blob/master/INSTALL.md#build-from-source

pkgs/by-name/holod/package.nix
themadbit: There are possibly half a dozen or more files that require these features, and I think it's a better fail-safe if the features are introduced elsewhere (these will most likely be in the main/lib.rs files).

pkgs/by-name/holod/package.nix
themadbit: good hack. thanks!

eljamm (approved):
LGTM

Issue #931, opened 2025-04-24
NGI Project: XR-Fragments


OluchiTheAnalyst:
### Short Summary

XR Fragments is developing a simple public protocol for networked 3D webrings to discover, reference, navigate and query 3D online content (read-only).

### NLnet page(s)

- https://nlnet.nl/project/XR-Fragments/
- https://nlnet.nl/project/XR-Teamware/

### Website

- https://xrfragment.org/

### Source repository

- https://github.com/coderofsalvation/xrfragment/
  - Language/Framework: HTML
  - Dependency management: Nix
  - Nix development environment: [shell.nix](https://github.com/coderofsalvation/xrfragment/blob/main/shell.nix)

### Documentation

- Usage Examples:
  - N/A
- Build from source/Development:
  - https://github.com/coderofsalvation/xrfragment/#development
- Other:
  - [License](https://github.com/coderofsalvation/xrfragment/blob/main/LICENSE)

### Artefacts

- CLI:
  - Xr-Fragment
    - documentation: https://github.com/coderofsalvation/xrfragment/tree/main/doc
    - examples: https://github.com/coderofsalvation/xrfragment/tree/main/example
    - tests: https://github.com/coderofsalvation/xrfragment/tree/main/test
 - Plugins: https://github.com/coderofsalvation/xrfragment/tree/main/dist

### Nixpkgs/NixOS

- Packages:
  - N/A
- Services:
  - N/A

### Extra Information

N/A

eljamm:
The repository moved to https://codeberg.org/coderofsalvation/xrfragment and also the languages should be haxe and javascript.

PR #953, opened 2025-04-28
Fix xrfragment

OluchiTheAnalyst:


pkgs/by-name/Xr-Fragments/default.nix
```diff
@@ -0,0 +1,37 @@
+{
+  pkgs ? import <nixpkgs> { },
```
eljamm, line 2: `pkgs` is already passed to the derivation, so there is no need to import it.

```suggestion
  pkgs
```

**PS:** importing it this way is bad for reproducibility since it will use the nixpkgs channel for whichever system is running this, instead of our nixpkgs, which might not be the same, and thus cause some undesirable effects.

That said, we want to explicitly pass each dependency instead of the whole `pkgs` anyways. For example:

```nix
{
  lib,
  stdenv,
  fetchgit,
  ...
}:
```

pkgs/by-name/Xr-Fragments/default.nix
eljamm, line 1: I believe the file should be called `package.nix` and not `default.nix` ([see here](https://github.com/ngi-nix/ngipkgs/blob/02ce465c911a3c61791a6f15be2f6001c33317d9/pkgs/by-name/default.nix#L62-L63))

OluchiTheAnalyst:
@eljamm All corrections made 💯

pkgs/by-name/Xr-Fragments/package.nix
```diff
@@ -0,0 +1,46 @@
...
+
+  buildPhase = ''
+    runHook preBuild
+
+    # build command
+
+    runHook postBuild
+  '';
+
+  installPhase = ''
```
eljamm, line 36: The comment about the `buildPhase` hooks also applies here 😉

pkgs/by-name/Xr-Fragments/package.nix
```diff
@@ -0,0 +1,46 @@
...
+    haxe
+    nodejs
+    yarn
+    makeWrapper
+  ];
+
+  buildPhase = ''
+    runHook preBuild
+
+    # build command
```
eljamm, line 31: This was just an illustrative example, what I was trying to say is that you should insert the build commands here, not replace them with this comment ...

```suggestion
    The command that builds the program should go here
```

pkgs/by-name/Xr-Fragments/package.nix
```diff
@@ -0,0 +1,46 @@
...
+  stdenv,
+  fetchgit,
+  haxe,
+  nodejs,
+  yarn,
+  makeWrapper,
+  lib,
+}:
+
+stdenv.mkDerivation rec {
```
eljamm, line 11: Replace [rec](https://nix.dev/guides/best-practices#recursive-attribute-set-rec) with finalAttrs (which is better for [overriding](https://wiki.nixos.org/wiki/Overlays)), whenever possible:

```suggestion
stdenv.mkDerivation (finalAttrs: {
```

pkgs/by-name/Xr-Fragments/package.nix
```diff
@@ -0,0 +1,46 @@
+{
+  stdenv,
+  fetchgit,
+  haxe,
+  nodejs,
+  yarn,
+  makeWrapper,
+  lib,
```
eljamm, line 8: Just a style nit

```suggestion
  lib,
  stdenv,
  fetchgit,
  haxe,
  nodejs,
  yarn,
  makeWrapper,
```

pkgs/by-name/Xr-Fragments/package.nix
```diff
@@ -0,0 +1,46 @@
...
+  pname = "xrfragment";
+  version = "0.5.2";
+
+  src = fetchgit {
+    url = "https://codeberg.org/coderofsalvation/xrfragment.git";
+    rev = "v0.5.2";
+    hash = "sha256-19zfnV7pVAAdft/KrXX4xbb3EgLw+M5SoEGXcKHR2kg=";
+  };
+
+  buildInputs = [
```
eljamm, line 21: Correct me if I'm wrong, but I believe that these should be `nativeBuildInputs` since they're tools to build the program rather than runtime dependencies necessary for it to run.

```suggestion
  nativeBuildInputs = [
```

pkgs/by-name/Xr-Fragments/package.nix
OluchiTheAnalyst: @eljamm Please check if this is the right thing for me to do in the ` Install phase `

pkgs/by-name/Xr-Fragments/package.nix
```diff
@@ -0,0 +1,48 @@
...
+  nativeBuildInputs = [
+    haxe
+    nodejs
+    yarn
+    makeWrapper
+  ];
+
+  buildPhase = ''
+    runHook preBuild
+     haxe build.hxml
```
eljamm, line 30: I think there is a more general `./make build` command you can use, here, that builds more things.

pkgs/by-name/Xr-Fragments/package.nix
```diff
@@ -0,0 +1,48 @@
...
+    runHook postBuild
+  '';
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/dist
+    cp -r dist/* $out/dist/
```
eljamm, line 38: Is `dist` the only thing we need to copy? Is there something to run/execute or is this just a library?

OluchiTheAnalyst: Hello 👋🏻 @eljamm looking at the ` build.hxml ` file, Yes, ` dist ` appears to be the only necessary thing to copy. The ` xrfragment ` project builds to language-specific source files (like xrfragment.js, .py, .lua, etc.) and does not define a main executable class. This suggests it is a library, not a CLI tool or standalone executable.

pkgs/by-name/Xr-Fragments/package.nix
OluchiTheAnalyst: ah, yes!

eljamm: I'm not talking about makefiles, there is a file called `make` in the repo that builds this, but it looks like it doesn't work for `aframe` since it tries to fetch some stuff from the internet (which won't work because of [the sandbox](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#tested-using-sandboxing)).

That said, it looks like it's failing for other reasons:

```
       > Running phase: buildPhase
       > /nix/store/plgd01gl1wvxyaxw2iafcr739s5icrsf-stdenv-linux/setup: line 1764: ./make: cannot execute: required file not found
```

pkgs/by-name/Xr-Fragments/package.nix
eljamm: Well, do you know how this would be used/loaded at least?

pkgs/by-name/Xr-Fragments/package.nix
OluchiTheAnalyst: so do you suggest i just leave `   haxe build.hxml ` as it is?

eljamm: I'd honestly just convert this PR to a draft until we figure out what we want to do.

OluchiTheAnalyst: Alright 👍🏻

eljamm:
Closing due to lack of activity and certainty on how this should be packaged. Tracking progress in https://github.com/ngi-nix/ngipkgs/issues/931.

Issue #866, opened 2025-04-18
US: Download button for demo code


fricklerhandwerk:
As a user of NGIpkgs, I want to be able to download the code that should end up as a file on my machine, in order to follow the instructions for running an application more easily.

## Acceptance criteria

Given a project has demo code, when I click the download button, it the code should end up on my machine as a file (ideally with a mandatory prompt for where to put it, so it ends up in the right place).

## Implementation notes

This needs the web server to send a content-disposition header to get the browser to recognise the file as an "attachment".

imincik:
Implemented in #962

PR #962, opened 2025-04-29
overview: provide download and copy buttons for demo code

### LINKED ISSUE #865: US: Copy button for code blocks
As a user of NGIpkgs, I want to be able to copy sample code into my clipboard in order to avoid selecting a text range manually.

# Acceptance criteria

Given a project with sample code in its overview page, when I click the copy button, the sample code will be in my clipboard.

erictapen:
The massive CSS class is due to the download button having to be an `a` element, as it wouldn't work without JS otherwise…

Closes #865 #866 

tmp iJS6s2EZ7B

### DIFF
--- a/overview/style.css
+++ b/overview/style.css
@@ -286,6 +286,42 @@ code, pre {
   overflow-x: auto;
 }
 
+.code-block {
+  position: relative;
+}
+
+.code-buttons {
+  position: absolute;
+  top: 0.5em;
+  right: 1em;
+}
+
+.button {
+  display: inline-block;
+  padding: 0.1em 0.4em;
+  background-color: transparent;
+  color: white;
+  text-align: center;
+  text-decoration: none;
+  border-radius: 5px;
+  border: solid 1px gray;
+  background-color: gray;
+  margin: 0;
+  font-size: 0.8rem;
+}
+
+.button.download {
+  background-color: green;
+}
+
+.button:hover {
+  background-color: darkgrey;
+}
+
+.button.download:hover {
+  background-color: darkgreen;
+}
+
 details > summary > h2, details > summary > h3 {
     display: inline;
 }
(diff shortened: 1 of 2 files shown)


overview/style.css
```diff
@@ -286,6 +286,42 @@ code, pre {
...
+  text-align: center;
+  text-decoration: none;
+  border-radius: 5px;
+  border: solid 1px gray;
+  background-color: gray;
+  margin: 0;
+  font-size: 0.8rem;
+}
+
+.button.download {
```
fricklerhandwerk, line 313: Can we have a download and copy symbol in front of the text please?

fricklerhandwerk (approved):

overview/style.css
erictapen: Tried my best, even though they don't look great

tmp xzyezad29B

fricklerhandwerk: That's okay, the icons are font-specific and we can still fix the font later

erictapen: Oh yes that symbol is way better

tmp NePBMAxo9V

eljamm: I'd say that icons would be descriptive enough on their own. It's pretty universally known what the download and copy symbols look like.

fricklerhandwerk: @eljamm I disagree, apart from that there's no such thing as universality, the well-known-ness depends on the specific audience. Shift by a couple of years of age up and people won't even know what a download is, shift a couple of years down and they may not know what a file is. Using multiple representations is good.

fricklerhandwerk (approved):

overview/style.css
eljamm: Okay, but can we have a bigger icon, at least? It doesn't really look that visible to me (tried on Firefox and Chrome):

image

Issue #960, opened 2025-04-28
[tracking] treewide: migrate to `lib'.moduleLocFromOptionString`


stepbrobd:
follow up of #857

triage: combine `lib'` with `lib` using `lib.extend` and export combined `lib` in flake outputs -- functions in `lib`?

Get files with `rg '"\$\{sources.inputs.nixpkgs}/nixos/modules'`

- [x] update the example in maintainers/templates/project/default.nix
- [x] projects/Agorakit/default.nix
- [x] projects/Pixelfed/default.nix
- [x] projects/Galene/default.nix
- [x] projects/Forgejo/default.nix
- [x] projects/ntpd-rs/default.nix
- [x] projects/Mastodon/default.nix
- [x] projects/Misskey/default.nix
- [x] projects/SCION/default.nix
- [x] projects/OpenWebCalendar/default.nix
- [x] projects/gnunet/default.nix
- [x] projects/Omnom/default.nix
- [x] projects/Rosenpass/default.nix
- [x] projects/Namecoin/default.nix
- [x] projects/Gancio/default.nix
- [x] projects/Canaille/default.nix
- [x] projects/Flarum/default.nix
- [x] projects/Wireguard/default.nix
- [x] projects/GNUTaler/default.nix
- [x] projects/Draupnir/default.nix

Issue #948, opened 2025-04-28
Refactor and clean up legacy code


eljamm:
- https://github.com/ngi-nix/ngipkgs/pull/861
- https://github.com/ngi-nix/ngipkgs/pull/922
- https://github.com/ngi-nix/ngipkgs/pull/864
- https://github.com/ngi-nix/ngipkgs/pull/893
- https://github.com/ngi-nix/ngipkgs/pull/890
- https://github.com/ngi-nix/ngipkgs/pull/881
- https://github.com/ngi-nix/ngipkgs/pull/880
- https://github.com/ngi-nix/ngipkgs/pull/879
- https://github.com/ngi-nix/ngipkgs/pull/878
- https://github.com/ngi-nix/ngipkgs/pull/877
- https://github.com/ngi-nix/ngipkgs/pull/873

**PS:** Most of this is done, already. This is just to track it, better.

PR #968, opened 2025-04-29
treewide: use `lib'.moduleLocFromOptionString` to locate re-exported modules

stepbrobd:
part of #960

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -21,34 +21,8 @@ in
 }:
 let
   dream2nix = (import sources.dream2nix).overrideInputs { inherit (sources) nixpkgs; };
-in
-rec {
-  inherit
-    lib
-    pkgs
-    system
-    sources
-    ;
 
-  overview = import ./overview {
-    inherit
-      lib
-      lib'
-      system
-      projects
-      ;
-    self = flake;
-    pkgs = pkgs // ngipkgs;
-    options = optionsDoc.optionsNix;
-  };
-
-  optionsDoc = pkgs.nixosOptionsDoc {
-    inherit (evaluated-modules) options;
-  };
-
-  # TODO: we should be exporting our custom functions as `lib`, but refactoring
-  # this to use `pkgs.lib` everywhere is a lot of movement
-  lib' = {
+  extension = rec {
     # Take an attrset of arbitrary nesting and make it flat
     # by concatenating the nested names with the given separator.
     flattenAttrs =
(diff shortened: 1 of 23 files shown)


default.nix
```diff
@@ -77,6 +67,19 @@ rec {
       lib.head (collectFiles optAttrs);
   };
 
+  # extended = lib.extend (_: _: extension);
+  extended = lib // extension;
```
stepbrobd, line 71: Both `extend` fixpoint and `//` are causing the evaluation of `builtins.currentSystem` impure function in `moduleLocFromOptionString`???

fricklerhandwerk: Oh no that's because there's still the call in flake.nix where system is null. @eljamm  can it be we didn't get to finishing that fix?

eljamm: Yes,, we wanted to [refactor things](https://github.com/ngi-nix/ngipkgs/issues/948) before splitting the [nixpkgs invocation](https://github.com/ngi-nix/ngipkgs/pull/861) and fixing that.

stepbrobd:
all custom `lib'` functions and references to `lib'` should just be under `lib` now

and extension functions are exported through flake outputs

stepbrobd:
> Hm, I still can't imagine how evaluating NixOS modules separately can be cheap...

Surprisingly it's not as bad as I thought it would be (~0.25s overhead)

And it's once the eval modules thunk is forced, it'll be cached for the rest of the evaluation

> is it possible to splice in alternative error messages into the module evaluation to avoid a useless trace if we went to the nixos.foo.bar syntax?

If we go with `a.b.c` syntax, I don't think it's possible to have good error message. It'll be either no error at all (the option to file mapping exist), or an attribute missing error like this:

```nix
nix-repl> { a = 1; }.b
error: attribute 'b' missing
       at «string»:1:1:
            1| { a = 1; }.b
             | ^
       Did you mean a?
```

But implementation wise, it'll still be a call to eval modules. But the the exported "function" will change to a service option (like `services.ntpd-rs`) to path mapping (like `"/nix/store/...-source/nixos/modules/services/networking/ntp/ntpd-rs.nix"`)

i.e.

```nix
lib.moduleLocFromOptionString "services.ntpd-rs"
# => "/nix/store/...-source/nixos/modules/services/networking/ntp/ntpd-rs.nix"
```

v.s.

```nix
lib.moduleLocFromOptionString.services.ntpd-rs # or whatever we rename it to
# => "/nix/store/...-source/nixos/modules/services/networking/ntp/ntpd-rs.nix"
```

stepbrobd:
`lib'` has been removed. To define a new helper function, add to `extension` attrset in `default.nix`. The function will be made available globally wherever `lib` is passed as argument.

Helper functions are also exported to flake outputs

fricklerhandwerk (approved):
Nice, thanks for rebasing. @eljamm shall we merge and then some day follow up to make the cool attrset API?

Issue #712, opened 2025-04-03
Implement example and test for Galene


eljamm:
## Instrucitons

Given that the [project has been added to projects/Galene](https://github.com/ngi-nix/ngipkgs/issues/711), add a basic example and [NixOS VM test](https://nix.dev/tutorials/nixos/integration-testing-using-virtual-machines.html) that showcase how to configure and use it.

Ellipse0934:
Hi, I'd like to take this on.

eljamm:
How is this going for you @Ellipse0934? Need help with anything?

Garvit-77:
Hey @eljamm I would like to take on the issue

Garvit-77:
Hey @fricklerhandwerk , @eljamm 
I was reviewing the mentioned PRs which depicts that the issue is already resolved 
or can you please guide me if I'm wrong .

eljamm:
Well, the example added in that PR is really basic (we can change the port and add certs)  and the test doesn't do anything, currently, so we don't know if the example even works. As such, this issue remains unsolved since what we currently have doesn't "showcase how to configure and use [Galene]" as per the task's requirements.

Garvit-77:
@eljamm thank you for that information , I will improvise example and adding tests to it

My background:
I have been contributing to CNCF projects earlier and 
I'm a newbie(parsing Docs for better understanding) to nix as I just came across the Summer of Nix 2025 program!

PR #983, opened 2025-05-01
Implement Test and Examples for Galene

Garvit-77:
#What This PR resolves:
- Modified example for galene by including ports and certs
- Modified test for galene 

#Resolves issue: #712

### DIFF
--- a/projects/Galene/example.nix
+++ b/projects/Galene/example.nix
@@ -1,5 +1,22 @@
 { ... }:
 {
-  services.galene.enable = true;
-  services.galene.insecure = true;
+  services = {
+    galene = {
+      enable = true;
+
+      port = 8443;
+
+      tls = {
+        enable = true;
+        certificatePath = "Path/"
+        keyfile = "keyfile.pem"
+      };
+
+      users = {
+        username = "admin"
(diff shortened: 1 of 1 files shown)


eljamm:
> Not for review yet

Then it needs to a draft

Garvit-77:
Hey @eljamm I have made appropriate changes to the PR and now it's ready for a review

thank you!

projects/Galene/example.nix
```diff
@@ -1,5 +1,22 @@
...
+  services = {
+    galene = {
+      enable = true;
+
+      port = 8443;
+
+      tls = {
+        enable = true;
+        certificatePath = "Path/"
+        keyfile = "keyfile.pem"
```
eljamm, line 12: This file needs to exist, else the module will fail to evaluate. Can't we have `tls` off for now?

projects/Galene/example.nix
eljamm, line 1: This only add the example, but the test still doesn't do anything, right?

Garvit-77: I'm sorry I did made changes to test too, but the commit doesn't show it

PR #998, opened 2025-05-05
overview: show tags for projects with a demo

### LINKED ISSUE #945: US: Show which projects have a demo available
As a user of NGIpkgs, I want to know which projects have a demo configuration, in order to pick one that I can actually try out on my machine.

## Acceptance criteria

TODO

## Implementation notes

We can check if `demo.nix` exists, and add a `demo` tag, if so. The tag can be positioned after the deliverable tags.

erictapen:
and make the tag a link to the corrrect anchor

Closes #945 

tmp NIu2zTU387


overview/default.nix
```diff
@@ -163,7 +163,7 @@ let
           prefixLength = 2;
         in
         optionalString (!empty projectOptions) ''
-          ${heading 2 "options" "Options"}
+          ${heading 2 "service" "Options"}
```
fricklerhandwerk, line 166: Aren't we showing programs yet? Where do those go?

overview/default.nix
```diff
@@ -232,19 +232,20 @@ let
     '';
 
     deliverableTags = rec {
-      one = label: ''
-        <span class="deliverable-tag">${label}</span>
+      one = name: label: ''
```
fricklerhandwerk, line 235: Note for later: At this point we should probably write module system type for the output, with an `outPath` or `__toString` option that will display the value. All this processing obscures the structure to the point I can barely follow.

erictapen: Oh yeah that sounds sensible, I think until now I basically followed the pre-existing paradigm and I might have overstretched its usefulness already.

overview/default.nix
erictapen: Oh good point, they are actually rendered together and should be split out into a different heading.

erictapen: But I suppose this should be addressed in a different PR?

erictapen: https://github.com/ngi-nix/ngipkgs/issues/1006

fricklerhandwerk (dismissed):
Seems to have some conflicts still, alright otherwise

eljamm (approved):

Issue #970, opened 2025-04-30
Fix port forwarding for Cryptpad demo


fricklerhandwerk:
Port forwarding through the VM breaks many interactions. Forward the same port as the service is using inside the VM.

PR #1000, opened 2025-05-05
cryptpad: fix demo port forwarding

imincik:
With this change, using demo VM, I can open a document at `http://localhost:19000/sheet/` or `http://localhost:19000/api/config` and `http://localhost:19000/checkup/` can execute all 55 tests (few of them are failing, but the issues are not a problem for our use case).

List of failing `checkup` tests:

> httpUnsafeOrigin and httpSafeOrigin are equivalent. In order for CryptPad's security features to be as effective as intended they must be different. See cryptpad/config/config.js. Changes to cryptpad/config/config.js will require a server restart in order for [/api/config](http://localhost:19000/api/config) to be updated.

> http://localhost:19000/sheet/inner.html was served with incorrect Content-Security-Policy headers.
> A value of "'self' blob: http://localhost:19000" was expected for the frame-src directive.
> A value of "'self' blob: http://localhost:19000 http://localhost:19000 ws://localhost:19000" was expected for the connect-src directive. This rule restricts which URLs can be loaded by scripts. Overly permissive settings can allow users to be tracked using external resources, while overly restrictive settings may block pages from loading entirely.

> http://localhost:19000/ was served with incorrect Content-Security-Policy headers.
> A value of "'self' blob: http://localhost:19000" was expected for the frame-src directive.
> A value of "'self' blob: http://localhost:19000 http://localhost:19000 ws://localhost:19000" was expected for the connect-src directive. This rule restricts which URLs can be loaded by scripts. Overly permissive settings can allow users to be tracked using external resources, while overly restrictive settings may block pages from loading entirely.

Closes: #970

### DIFF
--- a/projects/Cryptpad/demo.nix
+++ b/projects/Cryptpad/demo.nix
@@ -7,8 +7,8 @@
     settings = {
       httpPort = 9000;
       httpAddress = "0.0.0.0";
-      httpUnsafeOrigin = "http://localhost:${toString config.services.cryptpad.settings.httpPort}";
-      httpSafeOrigin = "http://localhost:${toString config.services.cryptpad.settings.httpPort}";
+      httpUnsafeOrigin = "http://localhost:19000";
+      httpSafeOrigin = "http://localhost:19000";
     };
   };
 }


projects/Cryptpad/demo.nix
```diff
@@ -7,8 +7,8 @@
     settings = {
       httpPort = 9000;
       httpAddress = "0.0.0.0";
-      httpUnsafeOrigin = "http://localhost:${toString config.services.cryptpad.settings.httpPort}";
-      httpSafeOrigin = "http://localhost:${toString config.services.cryptpad.settings.httpPort}";
+      httpUnsafeOrigin = "http://localhost:19000";
+      httpSafeOrigin = "http://localhost:19000";
```
eljamm, line 11: I feel like this works around the problem rather than solve it and might also be confusing to the user. For example, why are we setting `httpPort = 9000;` but accessing the service from port `19000`?

What I see as a solution is to have a 1:1 port mapping between the host and guest, but to change all ports in the config examples to be `10000 + current port`. As a user, this makes more sense to me and I wouldn't be confused when I change the config's port as I will get what I expect.

imincik: I don't think it is a workaround. It is correct configuration for specific piece of software for our specific deployment.
We can add a comment to explain the reason for different port number values.

I don't think we are going to deal with this problem very often with other NGI software.

imincik: > What I see as a solution is to have a 1:1 port mapping between the host and guest, but to change all ports in the config examples to be 10000 + current port. As a user, this makes more sense to me and I wouldn't be confused when I change the config's port as I will get what I expect.

Yes, this is a good alternative.

imincik:
Solved in https://github.com/ngi-nix/ngipkgs/pull/1004

PR #991, opened 2025-05-02
Replace yants with the module system 

eljamm:
Part of https://github.com/ngi-nix/ngipkgs/issues/949
Depends on https://github.com/ngi-nix/ngipkgs/pull/990

PR #1003, opened 2025-05-06
Check project types

eljamm:
Ideally, we'd only evaluate the modules without building anything.

Continuation of https://github.com/ngi-nix/ngipkgs/pull/991

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -162,6 +162,11 @@ rec {
     }
     // foldl recursiveUpdate { } (map (project: project.nixos.modules) (attrValues projects));
 
+  ngipkgsModules = lib.filter (m: m != null) (
+    lib.mapAttrsToList (name: value: value) nixos-modules.services
+    ++ lib.mapAttrsToList (name: value: value) nixos-modules.programs
+  );
+
   extendedNixosModules =
     with lib;
     [
--- a/default.nix
+++ b/default.nix
@@ -180,6 +180,30 @@ rec {
     ++ ngipkgsModules
     ++ nixosModules;
 
+  evaluated-modules = lib.evalModules {
+    class = "nixos";
+    modules = [
+      {
+        nixpkgs.hostPlatform = { inherit system; };
+
+        networking = {
+          domain = "invalid";
+          hostName = "options";
+        };
+
+        # faster eval time
+        documentation.nixos.enable = false;
+        documentation.man.generateCaches = false;
+
+        system.stateVersion = "23.05";
+      }
+      raw-projects # for checks
+    ] ++ extendedNixosModules;
+    specialArgs = {
+      modulesPath = "${sources.nixpkgs}/nixos/modules";
+    };
+  };
+
   ngipkgs = import ./pkgs/by-name { inherit pkgs lib dream2nix; };
 
   raw-projects = import ./projects {
--- a/default.nix
+++ b/default.nix
@@ -77,6 +77,27 @@ rec {
       in
       f "";
 
+    # Recursively evaluate attributes for an attribute set.
+    # Coupled with an evaluated nixos configuration, this presents an efficient
+    # way for checking module types.
+    forceEvalRecursive =
+      attrs:
+      lib.mapAttrsRecursive (
+        n: v:
+        if lib.isList v then
+          map (
+            i:
+            # if eval fails
+            if !(builtins.tryEval i).success then
+              # recursively recurse into attrsets
+              if lib.isAttrs i then lib'.forceEvalRecursive i else (builtins.tryEval i).success
+            else
+              (builtins.tryEval i).success
+          ) v
+        else
+          (builtins.tryEval v).success
+      ) attrs;
+
     # get the path of NixOS module from string
     # example:
     # lib'.moduleLocFromOptionString "services.ntpd-rs"
--- a/default.nix
+++ b/default.nix
@@ -167,15 +167,18 @@ rec {
     ++ lib.mapAttrsToList (name: value: value) nixos-modules.programs
   );
 
+  nixosModules = import "${sources.nixpkgs}/nixos/modules/module-list.nix";
   extendedNixosModules =
-    with lib;
     [
-      nixos-modules.ngipkgs
+      # Allow using packages from `ngipkgs` to be used alongside regular `pkgs`
+      {
+        nixpkgs.overlays = [ overlays.default ];
+      }
       # TODO: needed for examples that use sops (like Pretalx)
       sops-nix
     ]
-    ++ attrValues nixos-modules.programs
-    ++ attrValues nixos-modules.services;
+    ++ ngipkgsModules
+    ++ nixosModules;
 
   ngipkgs = import ./pkgs/by-name { inherit pkgs lib dream2nix; };
 
--- a/default.nix
+++ b/default.nix
@@ -22,45 +22,9 @@ rec {
     sources
     ;
 
-  rawNixosModules = lib'.flattenAttrs "." (
-    with lib;
-    foldl recursiveUpdate { } (attrValues (mapAttrs (_: project: project.nixos.modules) projects))
-  );
-
-  nixosModules = {
-    # The default module adds the default overlay on top of Nixpkgs.
-    # This is so that `ngipkgs` can be used alongside `nixpkgs` in a configuration.
-    default.nixpkgs.overlays = [ overlays.default ];
-  } // rawNixosModules;
-
-  optionsDoc =
-    let
-      nixosSystem =
-        args:
-        import (sources.nixpkgs + "/nixos/lib/eval-config.nix") (
-          {
-            inherit lib;
-            system = null;
-          }
-          // args
-        );
-    in
-    pkgs.nixosOptionsDoc {
-      options =
-        (nixosSystem {
-          inherit system;
-          modules = [
-            {
-              networking = {
-                domain = "invalid";
-                hostName = "options";
-              };
-
-              system.stateVersion = "23.05";
-            }
-          ] ++ lib.attrValues nixosModules;
-        }).options;
-    };
+  optionsDoc = pkgs.nixosOptionsDoc {
+    inherit (evaluated-modules) options;
+  };
 
   # TODO: we should be exporting our custom functions as `lib`, but refactoring
   # this to use `pkgs.lib` everywhere is a lot of movement
--- a/default.nix
+++ b/default.nix
@@ -204,6 +204,9 @@ rec {
     };
   };
 
+  # recursively evaluates each attribute for all projects
+  check-projects = lib'.forceEvalRecursive evaluated-modules.config.projects;
+
   ngipkgs = import ./pkgs/by-name { inherit pkgs lib dream2nix; };
 
   raw-projects = import ./projects {
--- a/projects/types.nix
+++ b/projects/types.nix
@@ -97,7 +97,7 @@ rec {
       options = {
         module = mkOption {
           description = "the example must be a NixOS module in a file";
-          type = pathInStore;
+          type = deferredModule;
         };
         description = mkOption {
           description = "description of the example, ideally with further instructions on how to use it";
@@ -151,7 +151,7 @@ rec {
             default = name;
           };
           module = mkOption {
-            type = moduleType;
+            type = deferredModule;
           };
           examples = mkOption {
             type = nullOr (attrsOf (nullOr example));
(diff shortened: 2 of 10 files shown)


projects/types.nix
```diff
@@ -97,7 +97,7 @@ rec {
       options = {
         module = mkOption {
           description = "the example must be a NixOS module in a file";
-          type = pathInStore;
+          type = deferredModule;
```
fricklerhandwerk, line 100: This must be a path, since we need to render it in documentation. Can't render an expression

eljamm: With `pathInStore`:

```nix
nix-repl> :p evaluated-modules.config.projects.Omnom.nixos.services.omnom.examples.base.module
error:
       … while evaluating the attribute 'value'
         at /nix/store/qmm7hgw60vp7vj9lma95hl329d0j3n6n-source/lib/modules.nix:1084:7:
         1083|     // {
         1084|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1085|       inherit (res.defsFinal') highestPrio;

       … while calling the 'addErrorContext' builtin
         at /nix/store/qmm7hgw60vp7vj9lma95hl329d0j3n6n-source/lib/modules.nix:1084:15:
         1083|     // {
         1084|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |               ^
         1085|       inherit (res.defsFinal') highestPrio;

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: A definition for option `projects.Omnom.nixos.services.omnom.examples.base.module' is not of type `path in the Nix store'. Definition values:
       - In `<unknown-file>': /home/user/ngipkgs/projects/Omnom/example.nix
```

with `deferredModule`:

```nix
nix-repl> :p evaluated-modules.config.projects.Omnom.nixos.services.omnom.examples.base.module
{
  imports = [
    {
      _file = "<unknown-file>, via option projects.Omnom.nixos.services.omnom.examples.base.module";
      imports = [ /home/user/ngipkgs/projects/Omnom/example.nix ];
    }
  ];
}
```

Still, the module isn't being rendered in the overview in both cases, AFAIK.

default.nix
```diff
@@ -139,17 +103,70 @@ rec {
...
+    ++ nixosModules;
+
+  toplevel = machine: machine.config.system.build.toplevel;
+  type-check = toplevel evaluated-modules;
+
+  # recursively evaluates each attribute for all projects
+  # faster than building exammples?
+  check-projects = lib.mapAttrsRecursive (
+    n: v: (builtins.tryEval v).success
+  ) evaluated-modules.config.projects;
```
eljamm, line 133: So checking types without building a system kinda works, now, but it's not perfect. For example:

```nix
nix-repl> b = builtins.tryEval evaluated-modules.config.projects.Omnom.metadata

nix-repl> b                                                                     
{
  success = true;
  value = { ... };
}

nix-repl> b.value
{
  links = { };
  subgrants = [ ... ];
  summary = "Omnom is a webpage bookmarking and snapshotting service.";
}

nix-repl> b.value.subgrants
[
  "omnom"
  "omnom-ActivityPub"
  «error: A definition for option `projects.Omnom.metadata.subgrants."[definition 1-entry 3]"' is not of type `string'. Definition values:
- In `<unknown-file>': 123123»
]
```

`subgrants` will successfully evaluate despite having an erroneous entry inside it.

projects/types.nix
fricklerhandwerk: What about just `path` then?

default.nix
mightyiam: https://discourse.nixos.org/t/comparing-module-system-configurations/59654/14?u=mightyiam

eljamm: > you need force the entire config tree but it may be infinite

The only issue I've observed is with items inside lists. It works well and recursively for everything else (wrong types, extra/missing declarations, ...).

That said, I found a bit of a workaround for the issue, which doesn't really look pretty, but it does work:

```nix
recurseEvalAttrs =
  attrs:
  lib.mapAttrsRecursive (
    n: v:
    if lib.isList v then
      map (
        i:
        if !(builtins.tryEval i).success then
          throw "Failed to evaluate a list item in ${
            lib.concatStringsSep "." (
              [
                "projects"
                (lib.getName attrs)
              ]
              ++ n
            )
          }"
        else
          (builtins.tryEval i).success
      ) v
    else
      (builtins.tryEval v).success
  ) attrs;
```

Which gives the following error message:

```shellSession
       … while evaluating the attribute 'metadata'

       … while evaluating the attribute 'subgrants'

       … from call site
         at /home/kuroko/Files/System/Development/Git/Personal/nix/ngipkgs/.worktree/heads/default.nix:52:13:
           51|           map (
           52|             e:
             |             ^
           53|             if !(builtins.tryEval e).success then

       … while calling anonymous lambda
         at /home/kuroko/Files/System/Development/Git/Personal/nix/ngipkgs/.worktree/heads/default.nix:52:13:
           51|           map (
           52|             e:
             |             ^
           53|             if !(builtins.tryEval e).success then

       … caused by explicit throw
         at /home/kuroko/Files/System/Development/Git/Personal/nix/ngipkgs/.worktree/heads/default.nix:54:15:
           53|             if !(builtins.tryEval e).success then
           54|               throw "Failed to evaluate a list element in ${
             |               ^
           55|                 lib.concatStringsSep "." (

       error: Failed to evaluate a list element in projects.Omnom.metadata.subgrants
```

> only config.system.build.toplevel gives you the exact attrset. What if we e.g. just printed the ...toplevel.outPath in the checks?

Maybe I'm missing something or doing something wrong, but `config.system.build.toplevel` didn't work for me as the evaluation always succeeds, ignoring any type errors.

> https://discourse.nixos.org/t/comparing-module-system-configurations/59654/14?u=mightyiam

This looks interesting. Is there a reason for having it implemented as a CLI option over something like [builtins.catchEvalErrors](https://github.com/NixOS/nix/pull/12705)?

eljamm: Acutally, this seems to work for everything?

```nix
recurseEvalAttrs =
  attrs:
  lib.mapAttrsRecursive (
    n: v:
    if lib.isList v then
      map (
        i:
        # if eval fails
        if !(builtins.tryEval i).success then
          # recursively recurse into attrsets
          if lib.isAttrs i then lib'.recurseEvalAttrs i else (builtins.tryEval i).success
        else
          # else return success
          (builtins.tryEval i).success
      ) v
    else
      (builtins.tryEval v).success
  ) attrs;
```

```shellSession
   … caused by explicit throw
     at /nix/store/qmm7hgw60vp7vj9lma95hl329d0j3n6n-source/lib/modules.nix:1139:11:
     1138|           in
     1139|           throw "A definition for option `${showOption loc}' is not of type `${type.description}'. Definition values:${showDefs allInvalid}"
         |           ^
     1140|       else

   error: A definition for option `projects.Omnom.metadata.subgrants."[definition 1-entry 3]"' is not of type `string'. Definition values:
   - In `<unknown-file>': 123123
```

eljamm: Well, `recurseEvalAttrs` essentially does that, but it accesses all the values and it's much faster than building the overview.

projects/types.nix
eljamm: > What about just `path` then?

That works, but it means we can't do [something like this](https://github.com/ngi-nix/ngipkgs/blob/71b96aaef81db1edc68a523b456e02e958b8b267/projects/CryptoLyzer/default.nix#L29-L45) anymore

eljamm:
To test the module types for a project, run:

```shellSession
nix build .#checks.x86_64-linux.projects/<PROJECT_NAME>/nixos/check
```

For example:

```shellSession
nix build .#checks.x86_64-linux.projects/Omnom/nixos/check
```

<details><summary>Wrong types</summary>
<p>

```nix
  metadata = {
    summary = "Omnom is a webpage bookmarking and snapshotting service.";
    subgrants = [
      "omnom"
      "omnom-ActivityPub"
      123123 # <----
    ];
  };
```

```shellSession
 error: A definition for option `projects.Omnom.metadata.subgrants."[definition 1-entry 3]"' is not of type `string'. Definition values:
 - In `<unknown-file>': 123123
```

```nix
  examples.base = {
    module = ""; # <----
    description = "Basic Omnom configuration, mainly used for testing purposes";
    tests.basic = null;
  };
```

```shellSession
 error: A definition for option `projects.Omnom.nixos.services.omnom.examples.base.module' is not of type `module'. Definition values:
 - In `<unknown-file>': ""
```

</p>
</details> 

<details><summary>Missing attributes</summary>
<p>

```nix
  examples.base = {
    module = ./example.nix;
    description = "Basic Omnom configuration, mainly used for testing purposes";
    # tests.basic = null; # <----
  };
```

```shellSession
 error: The option `projects.Omnom.nixos.services.omnom.examples.base.tests' was accessed but has no value defined. Try setting the option.
```

</p>
</details> 

<details><summary>Extra/Undefined attributes</summary>
<p>

```nix
  examples.base = {
    module = ./example.nix;
    description = "Basic Omnom configuration, mainly used for testing purposes";
    tests.basic = null;
    hello = "world"; # <----
  };
```

```shellSession
 error: The option `projects.Omnom.nixos.services.omnom.examples.base.hello' does not exist. Definition values:
 - In `<unknown-file>': "world"
```

</p>
</details>

projects/types.nix
fricklerhandwerk: Hm but that link shows an actual module, not an example

default.nix
```diff
@@ -77,6 +41,27 @@ rec {
       in
       f "";
 
+    # Recursively evaluate attributes for an attribute set.
+    # Coupled with an evaluated nixos configuration, this presents an efficient
+    # way for checking module types.
+    evalAttrsRecursive =
```
fricklerhandwerk, line 47: The name is a bit misleading, because you only get the `true`s and a bit of structure, not the actual value. Maybe we want it to be more like

```suggestion
    foceEvalRecursive =
```

default.nix
```diff
@@ -139,17 +124,62 @@ rec {
           nixpkgs.overlays = [ overlays.default ];
         };
     }
-    // foldl recursiveUpdate { } (map (project: project.nixos.modules) (attrValues projects));
+    // foldl recursiveUpdate { } (map (project: project.nixos) (attrValues projects));
 
+  ngipkgsModules = lib.filter (m: m != null) (
+    lib.mapAttrsToList (name: value: value) nixos-modules.services
+    ++ lib.mapAttrsToList (name: value: value) nixos-modules.programs
```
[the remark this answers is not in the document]
eljamm: Because `nixos-modules.{services,programs}` are attribute sets and not lists so we can't append them in the first place. That said, although we can update them together (nixos-modules.services // nixos-modules.programs), this might lead to undesirable overrides. For example, in a situation where a service and a program have the same name.

projects/types.nix
eljamm: I meant that by using `path` all modules will need to be stored in files and we won't be able to declare them like that.

eljamm:
I've split off a few fixes to a [separate PR](https://github.com/ngi-nix/ngipkgs/pull/1027) and dropped the treewide renaming of `nixos.modules.{services,projects}` to `nixos.{services,projects}` as that was breaking a few things. I think it would be better to look into doing that properly, later.

projects/types.nix
fricklerhandwerk: Yeah and that's alright. We require them as a string in a file for documentation, that's how it is.

default.nix
```diff
@@ -77,6 +41,27 @@ rec {
       in
       f "";
 
+    # Recursively evaluate attributes for an attribute set.
+    # Coupled with an evaluated nixos configuration, this presents an efficient
+    # way for checking module types.
+    forceEvalRecursive =
```
fricklerhandwerk, line 47: Won't this always succeed evaluating, even if there's a type error in the module? IIUC the check will produce a drv with true/false in an attribute tree, so that's not where it could "break" either.

eljamm: If type-checking fails then the evaluation will fail with an error and not `false`:

```nix
  extensions = true;
  links = { };
  module = «error: The option `projects.Omnom.nixos.modules.services.omnom.module' was accessed but has no value defined. Try setting the option.»;
  name = true;
```

This will thus cause the build to fail (see [more examples](https://github.com/ngi-nix/ngipkgs/pull/1003#issuecomment-2858195536)), so in a sense the attribute tree will never have `false` inside it, and that's okay for us since we only care about forcing evaluations.

fricklerhandwerk (approved):

PR #1005, opened 2025-05-07
overview: don't show copy button when JS is disabled

erictapen:
tmp dZZn1ya8BO

### DIFF
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -118,14 +118,16 @@ let
             ${optionalString downloadable ''
               <a class="button download" href="${filename}" download>Download</a>
             ''}
-            <button class="button copy" onclick="copyToClipboard(this, '${filename}')">
-                ${optionalString (!relative) ''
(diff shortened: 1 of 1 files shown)


overview/default.nix
```diff
@@ -412,6 +414,16 @@ let
       <body>
         ${args.content}
         <script>
+          // On document load, put all elements into the DOM that can be used with JS only
+          document.addEventListener("DOMContentLoaded", () => {
+            document.querySelectorAll("template").forEach(template => {
+                if (template.hasAttribute("scripted")) {
```
fricklerhandwerk, line 420:
```suggestion
            document.querySelectorAll("template[scripted]").forEach(template => {
```

I think could can simplify this by folding the conditional into the query

fricklerhandwerk (dismissed):

fricklerhandwerk (approved):

PR #1007, opened 2025-05-09
Add NGIpkgs section to the triaging template

eljamm:


fricklerhandwerk (dismissed):
May be worth noting that there's $REASONS for why we have things all over the place, but maybe not. LGTM

Issue #942, opened 2025-04-28
Nix installation instructions on native Linux distro


imincik:
Implement native Linux distro Nix installation instructions in the overview page.

* Debian/Ubuntu
```
    apt install --yes curl git nix
```

* Archlinux
```
    pacman --sync --refresh --noconfirm curl git nix
```

Issue #944, opened 2025-04-28
Instructions for configuring substituters for service demos


imincik:
To configure substituters, run following commands:

```bash
NIX_CONFIG='substituters = https://cache.nixos.org/ https://ngi.cachix.org/'$'\n''trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= ngi.cachix.org-1:n+CAL72ROC3qQuLxIHpV+Tw5t42WhXmMhprAGkRSrOw='

export NIX_CONFIG
```

Issue #943, opened 2025-04-28
Instructions for running projects with an older Nix version


imincik:
Due to https://github.com/NixOS/nix/issues/6820 , we can't use Nix package present in current Debian Stable, Ubuntu LTS and some non-lts Ubuntu versions as well.

* If Nix version is >= 2.24, run demo using

```
    nix-build /default.nix
```

* If Nix version < 2.24, run demo using

```
    NIXPKGS="https://github.com/NixOS/nixpkgs/archive/nixos-unstable.tar.gz"
    nix-shell -I nixpkgs=$NIXPKGS --packages nix --run "nix-build /default.nix"
```

fricklerhandwerk:
As @eljamm mentioned, we may want to use the exact Nixpkgs version we have in NGIpkgs to save a second download?

PR #1013, opened 2025-05-12
Add service demo instructions

imincik:
Closes: #942 #943 #944

Result on the overview page:
screen-imincik-2025-05-12-13:59:05

### DIFF
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -228,11 +228,11 @@ let
       <article class="page-width">
         ${heading 1 null name}
         ${render.metadata.one project.metadata}
+        ${render.options.many (pick.options project)}
+        ${render.examples.many (pick.examples project)}
         ${optionalString (project.nixos.examples ? demo) (
           render.serviceDemo.one project.nixos.modules.services project.nixos.examples.demo
         )}
-        ${render.options.many (pick.options project)}
-        ${render.examples.many (pick.examples project)}
       </article>
     '';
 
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
How about we generate an entire HTML element (`<article>` or what) for each platform, and render them as `<details>` xor styled as tabs? Interleaved instructions are really hard to read, and look scary from a distance. If we can show the easy ones by default, it would make a better first impression.

imincik:
Or we can collapse all instructions by default.

eljamm:
I think it would be better from a UX point of view to have this as a table. It should be possible [with just HTML](https://stackoverflow.com/questions/6906724/is-it-possible-to-have-tabs-without-javascript).

fricklerhandwerk:
Two sane ways are:
1. collapsing all by default but that requires an extra click by default
2. building some sort of tabbed view using the "checkbox trick" (make the element visible depending on the state of a radio button)

overview/default.nix
```diff
@@ -301,25 +301,51 @@ let
...
+                relative = true;
+                downloadable = true;
+              }}
+            </details>
           </li>
           <li>
-            <strong>Access the service</strong> with a web browser:
-            <a href="http://localhost:${toString servicePort}">http://localhost:${toString servicePort}</a>
+            <details>
+            <summary><strong>Enable binary substituters</strong></summary>
```
erictapen, line 327: We could make this optional (and hidden in a `detail`), but then we should provide motivation why to do it:
```suggestion
            <summary><strong>Enable binary substituters (optional, for less wait times)</strong></summary>
```

erictapen (changes requested):
I disagree with hiding the list items in `detail`s. Hiding information behind a click is good when users have conditinial interest in something, but in this case they either want to run the demo or not.

So it should be the other way around:
- _Run a demo deployment locally_ could be a detail
- individual steps should not be details except for the substituter config maybe
- Install instructions per platform should be details, though later I'd like to replace it with a UI environment that emphasizes the exclusivity of each option

overview/default.nix
imincik: > Run a demo deployment locally could be a detail

Yes, that's better. I agree.

imincik: > Run a demo deployment locally could be a detail

Done in https://github.com/ngi-nix/ngipkgs/pull/1013/commits/57aa464d1b04526d0d364f266c691508cd2c8788

imincik: * collapsed
screen-imincik-2025-05-15-08:18:49

* not collapsed
screen-imincik-2025-05-15-08:18:58

fricklerhandwerk (approved):

PR #1015, opened 2025-05-13
Remove migrated derivations

eljamm:


pkgs/by-name/default.nix
```diff
@@ -67,4 +67,14 @@ let
       throw "No package.nix or dream2.nix found in ${directory}"
   ) packageDirectories;
 in
-self
+lib.recursiveUpdate self {
```
fricklerhandwerk, line 70: This looks fishy... why would we add from pkgs here? Ngipkgs is used as an overlay.

eljamm: So the packages would be added to the checks. I'm mostly experimenting with this.

fricklerhandwerk (approved):
Nice

PR #1017, opened 2025-05-15
overview: use NIX_CONFIG from CI tests

fricklerhandwerk:
this is much less noisy. @imincik any reason not to do it that way?


imincik (changes requested):

overview/default.nix
```diff
@@ -322,7 +322,7 @@ let
           </li>
           <li>
             <strong>Enable binary substituters</strong>
-              <pre><code>NIX_CONFIG='substituters = https://cache.nixos.org/ https://ngi.cachix.org/'$'\n'''trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= ngi.cachix.org-1:n+CAL72ROC3qQuLxIHpV+Tw5t42WhXmMhprAGkRSrOw='</code></pre>
```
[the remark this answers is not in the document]
fricklerhandwerk: We could actually grab that value from the YAML to avoid that sort of omission in the future... 🤔

fricklerhandwerk:
image

overview/default.nix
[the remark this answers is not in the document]
fricklerhandwerk: Yes we can. It's ugly and beautiful at the same time!

.github/workflows/test-demo.yaml
```diff
@@ -33,10 +33,15 @@ jobs:
...
+          DISTRO: ${{ matrix.distro }}
+          NIX_CONFIG: |
+            extra-substituters = https://ngi.cachix.org/
+            extra-trusted-public-keys = ngi.cachix.org-1:n+CAL72ROC3qQuLxIHpV+Tw5t42WhXmMhprAGkRSrOw=
         run: >
           docker run
           --privileged
           --volume ./result/project/Cryptpad/default.nix:/default.nix
           --volume "$(pwd)/.github/workflows/test-demo.sh:/test-demo.sh"
           ${{ matrix.distro }}
```
eljamm, line 46:
```suggestion
          "$DISTRO"
```

fricklerhandwerk: Does that make a difference though?

eljamm: Not really, but we already have the the env var, so might as well use it.

.github/workflows/test-demo.yaml
```diff
@@ -33,10 +33,15 @@ jobs:
...
+        env:
+          DISTRO: ${{ matrix.distro }}
+          NIX_CONFIG: |
+            extra-substituters = https://ngi.cachix.org/
+            extra-trusted-public-keys = ngi.cachix.org-1:n+CAL72ROC3qQuLxIHpV+Tw5t42WhXmMhprAGkRSrOw=
         run: >
           docker run
           --privileged
           --volume ./result/project/Cryptpad/default.nix:/default.nix
           --volume "$(pwd)/.github/workflows/test-demo.sh:/test-demo.sh"
```
eljamm, line 45: The env var will have to be passed to docker as well:

```suggestion
          --volume "$(pwd)/.github/workflows/test-demo.sh:/test-demo.sh"
          -e DISTRO="$DISTRO"
```

fricklerhandwerk:
@erictapen wdyt about the IFD? Not nice I know, but we won't suddenly have a pure Nix YAML parser.

eljamm:
> @erictapen wdyt about the IFD? Not nice I know, but we won't suddenly have a pure Nix YAML parser.

Why don't we do it the other way around, and have the CI test use a generated `nix.conf` file, instead?

fricklerhandwerk:
Ah, like we evaluate the Nixpkgs revision?

wamirez:

PR #1025, opened 2025-05-15
overview: implement project list item as a module

### LINKED ISSUE #1018: Content type for project in overview
We have our projects as stand-alone data types. For the overview we need different representations of projects. One such representation is for the overview listing, which only shows a subset of information. (The final representation is HTML, but we can bake that at the very end.)

Task:
- Create a module system type that represents a project in the overview, storing all information we eventually want to display.
- Give it a read-only `__toString` field that defaults to the HTML representation 

Sketch: https://github.com/fricklerhandwerk/ngipkgs/commits/overview-refactor/

### LINKED ISSUE #1019: Use project content type in the overview
Given we have a [content type for projects in the overvie](#1018), use them to actually render the overview. Currently it's just a pile of inline conditionals.

fricklerhandwerk:
closes #1018
closes #1019

### DIFF
--- a/overview/content-types/project-list-item.nix
+++ b/overview/content-types/project-list-item.nix
@@ -1,0 +1,66 @@
+{
+  name,
+  config,
+  lib,
+  ...
+}:
+let
+  inherit (lib) mkOption types;
+in
+{
+  options = {
+    name = mkOption {
+      type = types.str;
+      default = name;
+    };
+    description = mkOption {
+      type = with types; nullOr str;
+    };
+    deliverables = mkOption {
+      type =
+        with types;
+        # XXX(@fricklerhandwerk): this is very simple-minded. we'll need to think
+        # about how to deal with cases where there are multiple services etc.
+        submodule {
+          options = {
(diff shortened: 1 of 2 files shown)


overview/content-types/project-list-item.nix
```diff
@@ -0,0 +1,66 @@
...
+                <a href="/project/${self.name}">${self.name}</a>
+              </h2>
+              ${concatStringsSep "\n" (
+                mapAttrsToList (
+                  deliverable: exists:
+                  optionalString exists ''<a class="deliverable-tag" href="/project/${self.name}#${deliverable}">${deliverable}</a>''
+                ) self.deliverables
+              )}
+            </div>
+            ${optionalString (!isNull self.description) ''
```
eljamm, line 58: [statix](https://github.com/oppiliappan/statix) lint

> `isNull` is deprecated, check equality with `null` instead

```suggestion
            ${optionalString (self.description != null) ''
```

eljamm (approved):

PR #1038, opened 2025-05-19
ci: use branch's ngipkgs when testing demo

eljamm:
Else it will always fail, since the PR's changes aren't in NGIpkgs' main branch.


eljamm:
Prime example: https://github.com/ngi-nix/ngipkgs/actions/runs/15117525897

fricklerhandwerk (approved):

Issue #1040, opened 2025-05-21
Fix CI test failures


eljamm:
I can't reproduce these failures locally, so I think they might be cached in Buildbot:

- https://buildbot.ngi.nixos.org/#/builders/5/builds/2155

PR #1041, opened 2025-05-21
Remove migrated derivations; filter out broken tests

eljamm:


eljamm:
The CI failure is from https://github.com/ngi-nix/ngipkgs/issues/1040 and is unrelated to the changes introduced here.

eljamm:
I can't reproduce the [Buildbot failures](https://buildbot.ngi.nixos.org/#/builders/5/builds/2155) locally so they may likely be cached failures. Can we fix this, @Erethon?

PR #1044, opened 2025-05-22
Cleanup monorepo code

eljamm:
- Make the data model types easier to read
- Use path instead of deferredModule for modules
- Cleanup unused code in flake

### DIFF
--- a/projects/types.nix
+++ b/projects/types.nix
@@ -201,11 +201,11 @@ rec {
                       };
                       # TODO: convert all subgrants to `subgrant`, remove listOf
                       subgrants = mkOption {
-                        type = either (listOf str) types'.subgrant;
+                        type = either (listOf str) subgrant;
                         default = null;
                       };
                       links = mkOption {
-                        type = attrsOf types'.link;
+                        type = attrsOf link;
                         default = { };
                       };
                     };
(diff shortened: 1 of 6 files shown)


projects/types.nix
```diff
@@ -7,8 +7,30 @@ let
     types
     mkOption
     ;
+
+  types' = import ./types.nix { inherit lib; };
```
eljamm, line 11: Does this make sense here to distinguish Nixpkgs `types` from our custom types or should we just use `rec`?

fricklerhandwerk: It's a rather weird thing to import the same file, this tangibly increases WTFs/min while reading. `let types' = {}; in types'` could work, but If I'm not overlooking something it's equivalent to using `rec` (modulo some edge cases where one or the other is slightly easier to read).

fricklerhandwerk:
Nice, thanks for splitting it up into commits, makes it really quick to review. This is all good except the last one, because we need *examples* to be paths, not the module implementation. I'll merge it myself after applying the fix.

projects/types.nix
eljamm: It's kind of odd for sure, but I'm wondering if it would be more readable if custom types were referenced by name. It would be interesting to see a namespaced `rec` for something like this. I'm not sure about this, myself, so feel free to remove this if you'd like.

eljamm:
CI failure unrelated to this PR (see https://github.com/ngi-nix/ngipkgs/pull/1041#issuecomment-2922721159).

PR #1050, opened 2025-05-28
ci: add test-demo-shell workflow

imincik:
Closes: https://github.com/ngi-nix/summer-of-nix/issues/215


.github/workflows/test-demo-shell.sh
eljamm, line 1: Could we maybe combine the `test-demo-shell` and `test-demo-vm` scripts and just have 1 GitHub CI workflow? I think the difference would be in which function is used in the overview demo's `default.nix` (ngipkgs.demo-shell or ngipkgs.demo-vm), which should be automatically handled.

fricklerhandwerk: Yeah I at the very least the "install nix" step can be deduplicated quite obviously.

PR #1051, opened 2025-05-30
overview: implement the whole project list using modules

erictapen:
Is this a correct continuation of your idea, @fricklerhandwerk ?

This is a almost a noop, except for the lastModified timestamp and the commit hash.

### DIFF
--- a/overview/content-types/project-list.nix
+++ b/overview/content-types/project-list.nix
@@ -1,0 +1,76 @@
+# The top-level overview for all projects
+{ config, lib, ... }:
+let
+  inherit (lib)
+    mkOption
+    types
+    concatMapStringsSep
+    attrValues
+    ;
+  inherit (types)
+    attrsOf
+    nullOr
+    functionTo
+    str
+    submodule
+    ;
+in
+{
+  options = {
+    projectListItems = mkOption {
+      type = attrsOf (submodule ./project-list-item.nix);
+    };
+    version = mkOption {
+      type = nullOr str;
+    };
+    lastModified = mkOption {
+      type = nullOr str;
+    };
+    __toString = mkOption {
+      type = functionTo str;
+      readOnly = true;
+      default = self: ''
+        <section class="page-width">
+          <h1>NGIpkgs</h1>
+
+          <p>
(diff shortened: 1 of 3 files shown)


overview/content-types/project-list.nix
```diff
@@ -0,0 +1,76 @@
...
+            </li>
+            <li>
+              <a href="https://github.com/ngi-nix/summer-of-nix/issues/41">Issue tracker</a>
+            </li>
+            <li>
+              <a href="https://nixos.org/community/teams/ngi/">Nix@NGI team</a>
+            </li>
+          </ul>
+
+        ${concatMapStringsSep "\n" toString (attrValues self.projectListItems)}
```
fricklerhandwerk, line 68: And then...

```suggestion
        ${concatMapStringsSep "\n" toString (self.projectListItems)}
```

overview/content-types/project-list.nix
```diff
@@ -0,0 +1,76 @@
...
+    attrsOf
+    nullOr
+    functionTo
+    str
+    submodule
+    ;
+in
+{
+  options = {
+    projectListItems = mkOption {
```
fricklerhandwerk, line 20:
```suggestion
    projects = mkOption {
```

Maybe for brevity?

fricklerhandwerk (approved):
Yes, perfect (modulo obligatory nits :P)

overview/content-types/project-list.nix
```diff
@@ -0,0 +1,76 @@
...
+      type = attrsOf (submodule ./project-list-item.nix);
```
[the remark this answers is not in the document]
fricklerhandwerk: That will also allow us to sort e.g. by number of subgrants.

fricklerhandwerk:
When building from the PR branch:
image

Can we get this fixed (and also put in order CSS-wise)? Then merge. Yes checks break but this has nothing to do with this PR

erictapen:
> Can we get this fixed (and also put in order CSS-wise)? Then merge. Yes checks break but this has nothing to do with this PR

What do you mean with order? It looks like an issue that only occurs when using devmode, should be fine on prod

fricklerhandwerk:
> order

Ah, make it abide to the content width and all. But we can fix that later as well.

PR #1078, opened 2025-06-03
Inventaire: init at 4.0.1

### LINKED ISSUE #897: NGI Project: Inventaire
### Short Summary

a libre collaborative book mapper

### NLnet page(s)

- https://nlnet.nl/project/Inventaire/
- https://nlnet.nl/project/Inventaire2/
- https://nlnet.nl/project/Inventaire-Self-hosted

### Website

https://github.com/inventaire/

### Source repository

- https://github.com/inventaire/inventaire-client
  - Language/Framework: Svelte, TypeScript
  - Dependency management: npm
  - Nix development environment: none

- https://github.com/inventaire/inventaire
  - Language/Framework: TypeScript
  - Dependency management: npm
  - Nix development environment: none

### Documentation

- Build from source/Development:
  - https://github.com/inventaire/inventaire#dependencies-to-install-manually
  - https://github.com/inventaire/inventaire-client#install
- Wiki:
  - https://wiki.inventaire.io
- Usage Examples:
  - https://wiki.inventaire.io/wiki/Guides
- Tech stack visualisation: https://inventaire.github.io/stack/

### Artefacts

- Browser front-end
- Back-end service 

### Nixpkgs/NixOS

- Packages:
  - none
- Services:
  - none

### Extra Information

_No response_

OPNA2608:
Closes #897

`elasticsearch` is marked unfree, so the example & VM test can't really be built on CI. Someone may need to opt into it by setting `setupElasticSearch` in the example to `true` and run the test locally with unfree packages enabled to test this properly every one in awhile.


OPNA2608:
Very likely still missing more patching to not attempt to put / look up further runtime-generated/modified files in the installed tree, but after preparing some files and directories in the CWD, it's now at the point where I'd start writing a module & VM test to setup everything it needs:

(`config` dir copied from source for now)
```
↪ tree -a
.
├── config
│   ├── default.cjs
│   ├── dev.cjs
│   ├── dev-federated.cjs
│   ├── production-alt.cjs
│   ├── production.cjs
│   ├── production-federated.cjs
│   ├── production-primary.cjs
│   ├── tests-api.cjs
│   ├── tests-api-federated-mocha.cjs
│   ├── tests-api-federated-server.cjs
│   ├── tests-api-mocha.cjs
│   ├── tests-api-server.cjs
│   ├── tests-integration.cjs
│   └── tests-unit.cjs
├── db
├── keys
└── storage

↪ /nix/store/scsxgxj5n7s2vd2p1j9jc40zkldk0fmj-inventaire/bin/inventaire
****** leveldb path ******
db/leveldb
-----
media storage: local
Checking session keys state as leading instance
generating new key
saving keys
waiting for CouchDB on http://yourcouchdbusername:*************@localhost:5984 (ECONNREFUSED)
Updated session keys saved
no hosts bans data found
waiting for CouchDB on http://yourcouchdbusername:*************@localhost:5984 (ECONNREFUSED)
waiting for CouchDB on http://yourcouchdbusername:*************@localhost:5984 (ECONNREFUSED)
[...]
```

OPNA2608:
Added Inventaire metadata, module, example (for testing purposes, absolutely needs proper secrets management in production usage) and VM test (so far only testing that it actually starts up and listens to a port, not that the client works).

One issue is that Inventaire asks for ElasticSearch, which has a license that is marked unfree. So CI cannot run the VM test for this:
```
error: Package ‘elasticsearch-7.17.27’ in /nix/store/g3z13ffr1wzw8m9wivqyni2nbjl956wj-source/pkgs/servers/search/elasticsearch/7.x.nix:86 has an unfree license (‘elastic20’), refusing to evaluate.
```

OPNA2608:
- Added a quick check to the VM test that the client responds as expected to an initial page visit.
  Inventaire-works
  Could test more functionality if anyone thinks it's necessary (registering accounts, registering books, starting a lending request).

- Enabled sitemaps building
  - Sadly, all the additional information inventaire gets from wikidata doesn't seem to get put into providing some initial options for authors, works, etc. [But I think that's correct](https://wiki.inventaire.io/wiki/Entities_data#Data_sources).
  - Because building the sitemaps involves processing wikidata queries, I've added a script that performs these queries ahead of time to put the results into the repo, and patched the code to read the queries from us instead.
    This seems to work fine! The downside is that we now need to carry a snapshot of wikidata query results… And we can't rely on the Wayback Machine for this, as they don't seem too happy when I try to request an archive of them to be made.
    Uncompressed, these amount to 119MB of JSON data. I've opted to compress them down to ~3MB with gzip. Not sure which option would be better, at least with the cleartext one, it shouldn't be hard to tell if the data is real/correct or corrupted/malicious… Iunno.

OPNA2608:
> I wonder how hard it would be to patch Inventaire to only do the queries at runtime.

Invoking the generation code *itself* shouldn't be hard. The main pain points there would be:

- Patching inventaire-client so it can use the sitemaps from a different location than the public folder in the store.
  Making a symlink in the derivation to the module-specified `stateDir` might JustWorks :tm:, can't say that I think it's a nice solution though…
- Pushing the queries to runtime would still cause issues, because then they would happen in the test VM. So we'd need saved query results and the patch to make it use them anyway… And the patch would need to be expanded to still have the option of doing the queries, for use cases outside of the test VM.

> Why is it that we need this at build time in the first place?

Because it's done [as part of the build script](https://codeberg.org/inventaire/inventaire-client/src/commit/v4.0.0/scripts/build#L73-L79). So we're closer to what upstream expects users to have run/build as part of the setup.

We can skip in, in which case the only impact seems to be that there would be a dead sitemaps entry in the HTML code, and search engines won't have any help with understanding & crawling the instance… Which would prolly be fine, since these sitemaps have [hardcoded links to the main inventaire.io instance](https://codeberg.org/inventaire/inventaire-client/src/commit/v4.0.0/scripts/sitemaps/generate_sitemaps.js#L69), so they're not specific to the running instance anyway.

Dunno if the lack of a sitemap affects search engine result ranking though… And if we or the users of this would care about that…

> If this is impractical, how about the source is compressed and unpacked in the store?

Which source? The query results? That's what I'm currently doing.

- gzip'd queries: https://github.com/OPNA2608/ngipkgs/tree/615176a65ac168dd55488012233cae15199115eb/pkgs/by-name/inventaire-client/sparql-queries
- Unpacked into store before building: https://github.com/OPNA2608/ngipkgs/blob/615176a65ac168dd55488012233cae15199115eb/pkgs/by-name/inventaire-client/package.nix#L17-L23

fricklerhandwerk:
Yes, let's leave the code behind the toggle.

projects/Inventaire/module.nix
```diff
@@ -0,0 +1,91 @@
...
+  options.services.inventaire = {
+    enable = lib.options.mkEnableOption "Inventaire server";
+
+    configOverridesFile = lib.options.mkOption {
+      type = lib.types.nullOr lib.types.path;
+      description = ''
+        Path to a .cjs file that exists at runtime with your desired settings overrides.
+
+        The passed path will be set up as `config/local.cjs`.
+      '';
```
eljamm, line 20: Is there some documentation that we can point users to which would tell them what the structure of this file is and what those overrides are?

projects/Inventaire/module.nix
```diff
@@ -0,0 +1,91 @@
...
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.inventaire;
+in
+{
+  options.services.inventaire = {
+    enable = lib.options.mkEnableOption "Inventaire server";
+
```
eljamm, line 13:
```suggestion
    package = lib.mkPackageOption pkgs "inventaire" { };

```

projects/Inventaire/tests/basic.nix
```diff
@@ -0,0 +1,73 @@
...
+          machine.wait_for_window("Inventaire - your friends and communities are your best library")
+
+          # Default instance name (insecure example config doesn't override this)
+          machine.wait_for_text("My Inventaire Instance")
+
+          machine.screenshot("Inventaire-works")
+        ''
+      else
+        # Without ElasticSearch, we only can get some vey early startup stuff done before we're stuck waiting for
+        # a nonexistent ElasticSearch instance
```
eljamm, line 68: Is there a way to fully disable ElasticSearch from the config? so it knows not to expect it.

projects/Inventaire/examples/basic.nix
```diff
@@ -0,0 +1,126 @@
...
+        port: ${toString couchdbPort},
+        username: "${couchdbUser}",
+        password: "${couchdbPassword}",
+      },
+      leveldb: {
+        directory: "${config.services.inventaire.stateDir}/db/leveldb",
+      },
+      elasticsearch: {
+        origin: "http://localhost:${toString elasticPort}",
+      },
```
eljamm, line 47: Can this be disabled/removed if ElasticSearch is not enabled? maybe this is what's causing the service to wait indefinitely for ElasticSearch to be available.

projects/Inventaire/default.nix
```diff
@@ -0,0 +1,34 @@
...
+  nixos = {
+    modules.services.inventaire = {
+      module = ./module.nix;
+      examples = {
+        basic = {
+          description = ''
+            A (very insecure!) example of setting up Inventaire and its dependencies on the local machine.
+          '';
+          module = ./examples/basic.nix;
+          tests.basic = import ./tests/basic.nix args;
```
eljamm, line 29: The [test data type](https://github.com/ngi-nix/ngipkgs/blob/0550976adba8fe48e8bf39d62e4cf143ac3f83ed/projects/types.nix#L218-L232) has been changed and expects this to be stored in a `module` option, now, so this will fail when we rebase.

```suggestion
          tests.basic.module = import ./tests/basic.nix args;
```

pkgs/by-name/tsx/package.nix
eljamm, line 1: How does this differ from the [tsx in Nixpkgs](https://github.com/NixOS/nixpkgs/blob/30e2e2857ba47844aa71991daa6ed1fc678bcbb7/pkgs/by-name/ts/tsx/package.nix#L86)?

pkgs/by-name/inventaire-client/9901-Use-saved-query-results.patch
eljamm, line 1: Can you please add a short description in the message body on what this does and why?

pkgs/by-name/tsx/package.nix
OPNA2608: `tsx` just wasn't packaged yet when this was initially submitted :). Will check if the one from Nixpkgs works for this.

projects/Inventaire/examples/basic.nix
OPNA2608: It waits for ElasticSearch, because ElasticSearch is a requirement for running Inventaire. Not setting the origin here will just take the [default origin](https://codeberg.org/inventaire/inventaire/src/commit/bd11c3c6d8dd435caa4e63e67d93b31ace7a754e/config/default.cjs#L149-L150), which will make it get stuck in the same way. Setting it to an empty string will make it error out:
```
machine # [   24.460067] inventaire-launch[1078]: uncaughtException Error: failed to connect to Elasticsearch
machine # [   24.461216] inventaire-launch[1078]:     at newError (file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-inventaire/lib/node_modules/inventaire/dist/server/lib/error/error.js:7:17)
machine # [   24.463210] inventaire-launch[1078]:     at waitForElastic (file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-inventaire/lib/node_modules/inventaire/dist/server/db/elasticsearch/init.js:47:30)
machine # [   24.465167] inventaire-launch[1078]:     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
machine # [   24.466390] inventaire-launch[1078]:     at async _waitForElasticsearchInit (file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-inventaire/lib/node_modules/inventaire/dist/server/db/elasticsearch/init.js:12:5) {
machine # [   24.468900] inventaire-launch[1078]:   statusCode: 500,
machine # [   24.469572] inventaire-launch[1078]:   context: {
machine # [   24.471547] inventaire-launch[1078]:     path: '/',
machine # [   24.472190] inventaire-launch[1078]:     options: {},
machine # [   24.472821] inventaire-launch[1078]:     elasticReqOptions: { ignoreCertificateErrors: false }
machine # [   24.473797] inventaire-launch[1078]:   },
machine # [   24.474301] inventaire-launch[1078]:   emitter: [
machine # [   24.474889] inventaire-launch[1078]:     'Error: failed to connect to Elasticsearch',
machine # [   24.475849] inventaire-launch[1078]:     'waitForElastic file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-: server/db/elasticsearch/init.js:47:30',
machine # [   24.477724] inventaire-launch[1078]:     'process.processTicksAndRejections node:internal/process/task_queues:105:5',
machine # [   24.479710] inventaire-launch[1078]:     'async _waitForElasticsearchInit file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-: server/db/elasticsearch/init.js:12:5'
machine # [   24.481362] inventaire-launch[1078]:   ],
machine # [   24.481907] inventaire-launch[1078]:   cause: TypeError: Invalid URL
machine # [   24.482682] inventaire-launch[1078]:       at new URL (node:internal/url:818:25)
machine # [   24.483492] inventaire-launch[1078]:       at request (file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-inventaire/lib/node_modules/inventaire/dist/server/lib/requests.js:41:22)
machine # [   24.485948] inventaire-launch[1078]:       at waitForElastic (file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-inventaire/lib/node_modules/inventaire/dist/server/db/elasticsearch/init.js:38:15)
machine # [   24.487865] inventaire-launch[1078]:       at _waitForElasticsearchInit (file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-inventaire/lib/node_modules/inventaire/dist/server/db/elasticsearch/init.js:12:11)
machine # [   24.489726] inventaire-launch[1078]:       at waitForElasticsearchInit (file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-inventaire/lib/node_modules/inventaire/dist/server/db/elasticsearch/init.js:19:17)
machine # [   24.491589] inventaire-launch[1078]:       at file:///nix/store/hg4ylvnfq2hwy83dgvk1cqgnbwjr8v59-inventaire/lib/node_modules/inventaire/dist/server/server.js:11:5
machine # [   24.493157] inventaire-launch[1078]:       at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
machine # [   24.494391] inventaire-launch[1078]:     code: 'ERR_INVALID_URL',
machine # [   24.495123] inventaire-launch[1078]:     input: '/'
machine # [   24.495720] inventaire-launch[1078]:   }
machine # [   24.497719] inventaire-launch[1078]: }
```

It does continue to run, but doesn't respond when trying to access it, so it wouldn't really improve the situation.

projects/Inventaire/examples/basic.nix
```diff
@@ -0,0 +1,126 @@
...
+          "${couchdbUser}" = "${couchdbPassword}";
+        };
+      })
+    ];
+  };
+
+  # !!! THIS IS INSECURE, DO NOT DO THIS IN PRODUCTION !!!
+  # This is only done like this here to allow easy testing & debugging.
+  # Manually complete the CouchDB setup instead.
+  systemd.services."couchdb-setup" = {
```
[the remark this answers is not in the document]
OPNA2608: I feel like adding an option users can switch on that introduces security concerns (like in the current script) is not a good idea… The script needs to talk to CouchDB, so it needs to know have an admin username & password.

I can think of silly ways to handle this (option to specify path(s) that will contain a usable couchdb username & password at runtime), but I don't know if that's really worth it for such an easily replicable service. Or just a one-time manual step…

projects/Inventaire/tests/basic.nix
OPNA2608: (See answer to https://github.com/ngi-nix/ngipkgs/pull/1078/files#r2174812995)

projects/Inventaire/examples/basic.nix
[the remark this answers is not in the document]
fricklerhandwerk: It's okay the way it is for now, since [interface composition at the module level](https://discourse.nixos.org/t/pre-rfc-decouple-services-using-structured-typing/58257/34 ) is not fully ready yet. We can still add it later when it starts hurting *and* as a community we get more experience with the pattern.

pkgs/by-name/inventaire-unwrapped/package.nix
```diff
@@ -0,0 +1,160 @@
...
+      echo "Missing: \$INVENTAIRE_SERVER"
+
+      # Fallback to hardcoding (this will likely fail to run due to missing client symlink from the wrapper)
+      INVENTAIRE_SERVER="$out/lib/node_modules/inventaire/dist/server/server.js"
+      echo "Falling back to: \$INVENTAIRE_SERVER"
+    fi
+
+    exec ${lib.getExe nodejs} "\$INVENTAIRE_SERVER"
+    EOF
+    chmod +x $out/bin/inventaire
```
eljamm, line 143: This is redundant since mode is not preserved in `pkgs/by-name/inventaire/package.nix`, right? This said, why don't we create `$out/bin/inventaire` in there instead (the wrapped inventaire)?

projects/Inventaire/tests/basic.nix
```diff
@@ -0,0 +1,60 @@
...
+  testScript =
+    { nodes, ... }:
+
+    ''
+      start_all()
+
+      machine.wait_for_unit("inventaire.service")
+
+      # With ElasticSearch (or something compatible, like OpenSearch), we can actually expect full startup & check that it works
+      machine.wait_for_console_text("inventaire server is listening on port 3006")
```
eljamm, line 48: Is the service port configurable? If so we probably want an option for that and for opening up the firewall.

**PS:** we probably shouldn't hardcode the port, here.

pkgs/by-name/inventaire-client/update-sparql-queries.sh
```diff
@@ -0,0 +1,59 @@
...
+  # Save the query URL, for checking that they match what the build would've fetched
+  printf "%s" "$url" > "$jsonDumpsDir"/"$name".url
+done < "$urlsDumpPath"
+
+# Put new data into place
+rm -vrf "$inventaireClientDir"/sparql-queries
+mv -v "$jsonDumpsDir" "$inventaireClientDir"/sparql-queries
+
+# If everything went well, we can delete all the temporary data
+rm -r "$tmpDir"
```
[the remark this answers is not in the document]
OPNA2608: Resolved by instead copying the file in question without preserving the write-protected permissions from the store.

projects/Inventaire/tests/basic.nix
OPNA2608: Port is part of the config that the user needs to supply, because the same config also houses all of the secrets to any involved services: https://codeberg.org/inventaire/inventaire/src/commit/bd11c3c6d8dd435caa4e63e67d93b31ace7a754e/config/default.cjs#L61

eljamm: How about we introduce a `secretsFile` option for those, then? Then the rest of the setting can be declaratively set in the module.

pkgs/by-name/inventaire-unwrapped/package.nix
OPNA2608: Can do.

pkgs/by-name/inventaire/package.nix
```diff
@@ -0,0 +1,12 @@
+{
+  runCommandNoCC,
+  inventaire-unwrapped,
+  inventaire-client,
+}:
+
+runCommandNoCC "inventaire" { } ''
+  cp -r --no-preserve=mode ${inventaire-unwrapped} $out
```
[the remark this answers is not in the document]
OPNA2608: Can do.

projects/Inventaire/examples/basic.nix
```diff
@@ -0,0 +1,122 @@
...
+  ];
+in
+{
+  # !!! THIS IS INSECURE, DO NOT DO THIS IN PRODUCTION !!!
+  # This is just done like this here to allow easy testing & debugging.
+  # Use some secrets management mechanism to get a string of a path at runtime here.
+  environment.etc."inventaire-config-overrides.cjs".text = ''
+    // This is not securely managed, and I'm fully aware of this.
+
+    module.exports = {
```
eljamm, line 29: Needed for the demo (they should probably be configurable, so users can set and change them freely):

```suggestion
    module.exports = {
        hostname: '0.0.0.0',
        port: 3006,
```

We could also link the [config file](https://codeberg.org/inventaire/inventaire/src/commit/bd11c3c6d8dd435caa4e63e67d93b31ace7a754e/config/default.cjs#L61) here, but this might be redundant since it's also in the option docs (but there could be users who won't read those).

projects/Inventaire/default.nix
```diff
@@ -0,0 +1,45 @@
...
+      module = ./module.nix;
+      examples = {
+        basic = {
+          description = ''
+            A (very insecure!) example of setting up Inventaire and its dependencies on the local machine.
+          '';
+          module = ./examples/basic.nix;
+          tests.basic.module = import ./tests/basic.nix args;
+        };
+      };
```
eljamm, line 42: The example is good enough to be a demo, IMO:

```nix
    demo.vm = {
      description = ''
        A (very insecure!) example of setting up Inventaire and its dependencies on the local machine.
      '';
      module = ./examples/basic.nix;
      tests.basic.module = import ./tests/basic.nix args;
    };
```

OPNA2608: Demo has been set to the example.

projects/Inventaire/module.nix
```diff
@@ -0,0 +1,157 @@
...
+          Notes:
+          - The settings here will be set up as `config/local.cjs`, to override the built-in defaults
+          - These settings are world-readable, and thus not suitable for secrets
+
+        ''
+        + settingsLinks;
+      default = {
+        hostname = "localhost";
+        port = 3006;
```
eljamm, line 62: This is awesome! Can we also have an `openFirewall` option as well? This is necessary for forwarding the port from the VM to the host.

PR #1080, opened 2025-06-03
Fix flake nixosModules

eljamm:
`nixosModules` was changed to a list in previous refactors, but it should be an attribute set of modules instead, so users can select individual projects to enable, according to their needs.

This PR restores the previous behaviour.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -46,7 +46,9 @@
           makemake = import ./infra/makemake { inherit inputs; };
         };
 
-        nixosModules = classic'.ngipkgsModules;
+        nixosModules = {
+          default.nixpkgs.overlays = [ overlay ];
+        } // lib'.flattenAttrs "." classic'.nixos-modules;
 
         # Overlays a package set (e.g. Nixpkgs) with the packages defined in this flake.
         overlays.default = overlay;


flake.nix
```diff
@@ -46,7 +46,9 @@
           makemake = import ./infra/makemake { inherit inputs; };
         };
 
-        nixosModules = classic'.ngipkgsModules;
+        nixosModules = {
+          default.nixpkgs.overlays = [ overlay ];
+        } // lib'.flattenAttrs "." classic'.nixos-modules;
```
fricklerhandwerk, line 51: Not sure if I really want to start that philosophical debate again, but how about we fix underlying UX issue instead of piling up hacks?

For starters, let's just not expose the modules in the flake, to avoid the impression it's supposed to work or is maintained.

And then, in a properly refined and scheduled user story towards the end of the year think more clearly about what we actually want to achieve and what it should be like for consumers.

In my experience, module composition across source boundaries is not really figured out yet, anywhere. And I'd prefer not to pretend it is just to get some random unplanned use case working.

Not saying the request was invalid or the attempting to serve it is wrong. I'd just lean towards building something sustainable here, especially given we know a pile of patterns that aren't. And this happens to be such a fundamental problem that we shouldn't just wing it.

PR #1092, opened 2025-06-04
modularize rendering of installation instructions

fricklerhandwerk:
@eljamm needs a bit of cleanup

@erictapen we took the opportunity to fix it in the context of a code review. now https://github.com/ngi-nix/summer-of-nix/issues/230 should be much more pleasant


overview/content-types/commands.nix
```diff
@@ -0,0 +1,64 @@
+{
+  name,
+  config,
+  lib,
+  ...
+}:
+let
+  inherit (lib) mkOption types;
+  commands = {
```
fricklerhandwerk, line 9:
```suggestion
  # TODO: wrap in submodule and make it render itself
  commands = {
```

PR #1097, opened 2025-06-05
overview: modularize NIX_CONFIG

themadbit:


overview/content-types/nix-config.nix
```diff
@@ -0,0 +1,29 @@
...
+        let
+          separator = "'$'\\n''";
+          concatSettings = lib.concatStringsSep separator (
+            lib.map (attrs: "${attrs.name} = ${toString attrs.value}") self.settings
+          );
+        in
+        "NIX_CONFIG='${concatSettings}'";
+    };
+    settings = mkOption {
+      type = with types; listOf (submodule ./config-item.nix);
```
erictapen, line 26: I think the submodule definition could be inlined here. I think file granularity still needs to be figured out but for now as a rule of thumb stuff that doesn't have a `__toString` attr shouldn't live in its own file.

erictapen (commented):
Hey Mark, this is a cool addition to the growing module system of  the overview page!

overview/content-types/nix-config.nix
erictapen: Actually nevermind, just saw what @fricklerhandwerk did in https://github.com/ngi-nix/ngipkgs/pull/1092 and it was his idea in the first place, so let's roll with this granularity.

erictapen (approved):

overview/content-types/nix-config.nix
fricklerhandwerk: > __toString attr shouldn't live in its own file.

It doesn't, but arguably we can collapse ./config-item into this module.

PR #1098, opened 2025-06-05
flake.lock: Update

eljamm:


Erethon (approved):
Makes sense I believe. If you want you can remove the manual flake.lock update since it's now handled automatically again.

PR #1102, opened 2025-06-06
Mobilizon: init

erictapen:
I remembered that ages ago I wrote https://docs.mobilizon.org/3.%20System%20administration/install/nixos/ which is still usable, so I thought I could translate this into an easy example.

This isn't ready yet as the example needs some more steps for secret provisioning, see the link.


eljamm (approved):
Thanks Kerstin!

PR #1106, opened 2025-06-09
Add service module issue template

eljamm:
Part of https://github.com/ngi-nix/summer-of-nix/issues/163

### DIFF
--- a/.github/ISSUE_TEMPLATE/task-module-service.md
+++ b/.github/ISSUE_TEMPLATE/task-module-service.md
@@ -1,6 +1,6 @@
 ---
 name: "Task: NixOS service module"
-about: "Implement NixOS service module for an NGI project"
+about: "Task for implementing NixOS service module for an NGI applicaton"
 title: "Implement NixOS service module for <PROJECT_NAME>"
 projects: Nix@NGI
 type: task
--- a/.github/ISSUE_TEMPLATE/task-module-service.md
+++ b/.github/ISSUE_TEMPLATE/task-module-service.md
@@ -1,0 +1,27 @@
+---
+name: "Task: NixOS service module"
+about: "Implement NixOS service module for an NGI project"
+title: "Implement NixOS service module for <PROJECT_NAME>"
+projects: Nix@NGI
+type: task
+labels: ''
+assignees: ''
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/task-module-service.md
```diff
@@ -0,0 +1,26 @@
+---
+name: "Task: NixOS service module"
+about: Serivce module task for an NGI project
```
fricklerhandwerk, line 3:
```suggestion
```

If it's not a mandatory field let's not write anything extra here.

.github/ISSUE_TEMPLATE/task-module-service.md
```diff
@@ -0,0 +1,26 @@
...
+<!-- Replace `<TRIAGE_ISSUE_NUMBER>` with the issue number that contains the project's triaged information.
+If one doesn't exist, create it by following the instructions in the [contributor documentation](https://github.com/ngi-nix/ngipkgs/blob/main/CONTRIBUTING.md#triaging-an-ngi-project). -->
+
+Given #<TRIAGE_ISSUE_NUMBER> for metadata about the application, implement a NixOS service module for <PROJECT_NAME>.
+
+### Tasks
+
+After or during the implementation of this module, create issues for the following deliverables using the appropriate GitHub template:
+
+- [ ] Usage example configuration
```
fricklerhandwerk, line 23: It's either

```suggestion
- [ ] Example configuration
```

or

```suggestion
- [ ] Usage example
```

But I see what you tried to do there. Unless you want to go for a bit of a linguistic stretch, probably example config is more suitable here because it better indicates that this is something testable. Unless you want to emphasize it's about the using. In any case since this is mostly for internal communication either is fine, and we'll fix forward as the need arises.

fricklerhandwerk (dismissed):

.github/ISSUE_TEMPLATE/task-module-service.md
eljamm: Actually, I think it's mandatory. I'm testing another template and GitHub is complaining that it should not be empty:

image

PR #1107, opened 2025-06-09
ci: add test-demo-shell workflow; refactor test script

eljamm:
Also run VM and shell demo tests for 2 projects to verify the workflow.


imincik (approved):
Looks good, thanks !

PR #1115, opened 2025-06-09
demo: add envVars option to shell 

eljamm:
Test with:

```nix
{
  ngipkgs ?
    import
      (fetchTarball "https://github.com/eljamm/ngipkgs/archive/de1fec73821f847e6c0c5cc12094b780a4f07413.tar.gz")
      { },
}:
ngipkgs.demo-shell (
  { pkgs, ... }:
  {
    demo-shell.xrsh = {
      programs.xrsh = pkgs.xrsh;
      envVars.XRSH_PORT = "9090";
    };
  }
)
```

```shellSession
[demo-shell] ~/ngipkgs > echo $XRSH_PORT
9090
```


overview/demo/shell.nix
```diff
@@ -53,6 +54,14 @@ in
             };
             default = { };
           };
+          envVars = mkOption {
```
eljamm, line 57: We can also call it `variables` or something like that.

Issue #1111, opened 2025-06-09
Implement xrsh shell demo


themadbit:
I'm already working on this; the issue is suitable for tracking progress.
Please add it as sub-issue to #1077

PR #1118, opened 2025-06-10
xrsh: add shell demo

themadbit:
closes: #1111 

nit-change: replaces hard-coded port with a modular variable XRSH_PORT

concerns:

- currently depends on 
   - #1115 
   - #1134  by @eljamm 

- @fricklerhandwerk, you had proposed that demo examples should ideally be the same as those used in the program/service module example. I have tried to implement that herein. However, `overview` displays both as different examples or rather, independently. I think it's redundant for a smoother user experience. How do you propose we approach this?

image

### DIFF
--- a/projects/xrsh/default.nix
+++ b/projects/xrsh/default.nix
@@ -33,5 +33,11 @@
     };
   };
 
+  nixos.examples.demo-shell = {
+    module = ./programs/xrsh/examples/basic.nix;
+    description = "xrsh demo";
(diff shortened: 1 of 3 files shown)


fricklerhandwerk:
Ahh good, these developments show that we need to refine our data model a bit. For instance we could have an extra field just for exactly one demo; that would nudge us to deliberately choose the intended experience for each application.

eljamm (approved):
Looks good!

Issue #1028, opened 2025-05-16
Spike: Test all project demos


eljamm:
We want to test all applications demos; ideally with the same tests as for the examples. Currently, we can't easily iterate 
over all of them, and  they will trigger on every PR, even when the changes are unrelated to the service or demo.

The desired outcome of the spike are refined user stories to have all project demos tested, somewhat efficiently.

## Notes

- Tests should be cached in order to minimize time and resources

imincik:
@eljamm and @imincik will meet on Wednesday.

eljamm:
@imincik and I met today and we broke down tasks for https://github.com/ngi-nix/ngipkgs/issues/1139. This spike will be solved by https://github.com/ngi-nix/ngipkgs/issues/1192.

PR #1134, opened 2025-06-11
Refactor module composition

eljamm:
## Changes

- Refactor module composition: 
  - expose `nixos-modules` to the flake's `nixosModules`
    - supersedes https://github.com/ngi-nix/ngipkgs/pull/1080
    - I've opted to adding a note saying that this is experimental as opposed to removing the modules, but we should look into properly designing this, as mentioned in https://github.com/ngi-nix/ngipkgs/pull/1080#discussion_r2126451763
  - Import `demo-shell` in ngipkgs module
    - This makes it possible to use `demo-shell` inside the NixOS tests, which we're gonna need for experimenting with https://github.com/ngi-nix/ngipkgs/issues/1028
    - Note: we should look into separating this from production code, but this is a good step forward
- Small refactors
 
Needed for: #1118 #1123

Worked on this with @themadbit

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -272,6 +272,7 @@ rec {
       pkgs
       sources
       extendedNixosModules
+      system
       ;
   };
 
--- a/default.nix
+++ b/default.nix
@@ -21,7 +21,6 @@ in
 }:
 let
   dream2nix = (import sources.dream2nix).overrideInputs { inherit (sources) nixpkgs; };
-  sops-nix = import "${sources.sops-nix}/modules/sops";
 in
 rec {
   inherit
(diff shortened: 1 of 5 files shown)


eljamm:
Archlinux is failing to install Nix:

```shellSession
---> Building test ...
nix: error while loading shared libraries: libaws-cpp-sdk-identity-management.so: cannot open shared object file: No such file or directory
Using Nix from Nixpkgs unstable
nix-instantiate: error while loading shared libraries: libaws-cpp-sdk-identity-management.so: cannot open shared object file: No such file or directory
```

fricklerhandwerk (approved):

PR #1141, opened 2025-06-12
overview: move to default.nix and pass it to flake.nix

eljamm:
This make the devmode logic much simpler and gives us more ganular control on what to render/output according to our needs. For example, we could have an argument that toggles `ci-mode`, which makes the overview output a `.env` file from demo instructions, which the CI can use.


fricklerhandwerk (approved):
Yeah cool, good step forward. It should now become pretty clear that we should also modularize the entry point to the overview (eventually, step by step), as the number of arguments and their explicit passing around gets messier.

PR #1145, opened 2025-06-13
overview: implement copy button for all code blocks

themadbit:
WIP for #865

### DIFF
--- a/overview/content-types/commands.nix
+++ b/overview/content-types/commands.nix
@@ -38,28 +38,32 @@ in
             })
           );
     };
-    __toString = mkOption {
-      type = with types; functionTo str;
-      readOnly = true;
-      default =
-        self:
-        if lib.isList self.instructions then
-          ''
-            <ul>
-              ${lib.concatMapStringsSep "\n" (i: ''
-                <li>
-                  <dt>${i.platform}</dt>
-                  <dd>
-                    ${i.commands.bash}
-                  </dd>
-                </li>
-              '') self.instructions}
-            </ul>
-          ''
-        else
-          ''
-            ${self.instructions.commands.bash}
-          '';
-    };
+    # __toString = mkOption {
+    # type = with types; functionTo str;
+    # readOnly = true;
+    # default =
+    # self:
+    # if lib.isList self.instructions then
+    # ''
+    # <ul>
+    # ${lib.concatMapStringsSep "\n" (i: ''
+    # <li>
+    # <dt>${i.platform}</dt>
+    # <dd>
+    # ${render.codeBlock.one {
+    # content = toString i.commands.bash;
+    # copyableContent = i.commands.bash.input;
+    # }}
+    # </dd>
+    # </li>
+    # '') self.instructions}
+    # </ul>
+    # ''
+    # else
+    # render.codeBlock.one {
+    # content = toString self.instructions.commands.bash;
+    # copyableContent = self.instructions.commands.bash.input;
+    # };
+    # };
   };
 }
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -134,6 +134,27 @@ let
           </div>
         </div>
       '';
+    codeBlock.one =
+      {
+        content,
+        copyableContent ? content,
+        language ? "bash",
+      }:
+      ''
+        <div class="code-block">
+          ${content}
+          <div class="code-buttons">
+            <template scripted>
+              <button class="button copy" onclick="copyInlineToClipboard(this)">
+                <script type="application/json">
+                  ${toJSON copyableContent}
+                </script>
+                Copy
+              </button>
+            </template>
+          </div>
+        </div>
+      '';
     options = rec {
       one =
         prefixLength: option:
@@ -279,6 +300,7 @@ let
         servicePort = if openPorts != [ ] then (builtins.head openPorts) else "";
         installation-instructions = eval {
           imports = [ ./content-types/commands.nix ];
+          _module.args = { inherit render; };
           instructions = [
             {
               platform = "Arch Linux";
@@ -300,6 +322,31 @@ let
             }
           ];
         };
+
+        renderInstructions =
+          instructions:
+          if lib.isList instructions.instructions then
+            ''
+              <ul>
+                ${lib.concatMapStringsSep "\n" (i: ''
+                  <li>
+                    <dt>${i.platform}</dt>
+                    <dd>
+                      ${render.codeBlock.one {
+                        content = toString i.commands.bash;
+                        copyableContent = i.commands.bash.input;
+                      }}
+                    </dd>
+                  </li>
+                '') instructions.instructions}
+              </ul>
+            ''
+          else
+            render.codeBlock.one {
+              content = toString instructions.instructions.commands.bash;
+              copyableContent = instructions.instructions.commands.bash.input;
+            };
+
         nix-config = eval {
           imports = [ ./content-types/nix-config.nix ];
           settings = [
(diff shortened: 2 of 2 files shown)


overview/content-types/shell-code.nix
```diff
@@ -6,14 +6,37 @@
 }:
 let
   inherit (lib) mkOption types;
+
+  codeBlock.one =
```
fricklerhandwerk, line 10: This can be a module field, simply a string value that reads from config. or in a `let` block at `__toString`

overview/content-types/commands.nix
```diff
@@ -46,20 +46,18 @@ in
...
-                <li>
-                  <dt>${i.platform}</dt>
-                  <dd>
-                    ${i.commands.bash}
-                  </dd>
-                </li>
-              '') self.instructions}
+            ${lib.concatMapStringsSep "\n" (i: ''
+              <li>
+              <details>
```
fricklerhandwerk, line 51: I wouldn't go into these skinning details here though and instead focus on the mechanism. My plan is to use a radio button and control visibility via CSS, because you want a single-select behavior. Details will require closing a collapsed section to get it out of the way, that's way too much interaction.

overview/content-types/shell-code.nix
```diff
@@ -13,7 +13,7 @@ in
       type = types.str;
     };
     input = mkOption {
-      type = types.lines;
+      type = with types; either lines (listOf lines);
```
fricklerhandwerk, line 16: As commented prior, the multiplicity should be handled on the outside. Multiple commands could be a "session", but one command can only be `lines` once.

overview/content-types/shell-code.nix
```diff
@@ -23,12 +23,33 @@ in
...
-          <pre class="shell"><code>${trim ''
-            ${self.prompt} ${self.input}
-            ${optionalString (!isNull self.output) self.output}
-          ''}</code></pre>
-        '';
+        self:
+        let
+          shell = input: ''
+            <pre class="shell"><code>${lib.trim ''
+              ${self.prompt} ${input}
```
fricklerhandwerk, line 30: Why not

```suggestion
              ${self.prompt} ${self.input}
```

overview/default.nix
```diff
@@ -319,12 +319,36 @@ let
             }
           ];
         };
+        # TODO: this will be refactored
```
fricklerhandwerk, line 322: Yes but how ;)

overview/content-types/shell-code.nix
eljamm: Because currently `self.input` can be a list and that wouldn't work like this.

overview/content-types/shell-code.nix
eljamm: Yes, we've yet to reach that part as we spent most of the time today debugging issues with the copy buttons themselves (which are now solved):

  - the `Debian 12 and Ubuntu 24.04/24.10` section had 2 copy buttons
  - after you copy for the first time, you can't copy again

But I agree with you that this should be one command.

overview/default.nix
eljamm: Like this: https://github.com/ngi-nix/ngipkgs/pull/1170

I've had this for a while as you were aware, but I couldn't find a nice and clean way to re-use the env var in CI, so let's just leave that for later.

eljamm:
I believe we've addressed all comments about the copy button implementation. We're gonna finish cleaning this up later today and it should be good to go.

overview/content-types/commands.nix
```diff
@@ -9,10 +9,15 @@ let
   # TODO: wrap in submodule and make it render itself
   commands = {
```
fricklerhandwerk, line 10: How about the following set of changes (sorry github doesn't let me suggest across changed and unchanged lines):
```suggestion
  session = {
```

overview/content-types/commands.nix
```diff
@@ -9,10 +9,15 @@ let
...
     bash = mkOption {
-      type = types.submodule ./bash-code.nix;
+      type = with types; nullOr (submodule ./bash-code.nix);
+      default = null;
```
fricklerhandwerk, line 13:
```suggestion
      type = with types; listOf (submodule ./bash-code.nix);
      default = [];
```

overview/content-types/commands.nix
```diff
@@ -9,10 +9,15 @@ let
...
     };
     # TODO: moar shells
   };
+  session = mkOption {
+    type = with types; nullOr (submodule ./shell-session.nix);
+    default = null;
+  };
```
fricklerhandwerk, line 20:
```suggestion
```

overview/content-types/commands.nix
```diff
@@ -23,7 +28,7 @@ in
           # cross-platform
           (submodule {
             options = {
-              inherit commands;
+              inherit commands session;
```
fricklerhandwerk, line 31:
```suggestion
              inherit session;
```

overview/content-types/commands.nix
```diff
@@ -33,7 +38,7 @@ in
                 platform = mkOption {
                   type = types.str;
                 };
-                inherit commands;
+                inherit commands session;
```
fricklerhandwerk, line 41:
```suggestion
                inherit session;
```

overview/content-types/commands.nix
```diff
@@ -46,20 +51,21 @@ in
...
-              '') self.instructions}
+            ${lib.concatMapStringsSep "\n" (i: ''
+              <li>
+                <summary>${i.platform}</summary>
+                ${lib.optionalString (i.commands.bash != null) ''
+                  ${toString i.commands.bash}
+                ''}
+                ${lib.optionalString (i.session != null) ''
+                  ${toString i.session}
+                ''}
```
fricklerhandwerk, line 62:
```suggestion
                ${lib.optionalString (i.session.bash != []) ''
                  ${lib.concatStringsSep "\n" (toString i.session.bash)}
                ''}
```

fricklerhandwerk (changes requested):

overview/content-types/shell-command.nix
```diff
@@ -0,0 +1,49 @@
...
+    };
+    stderr = mkOption {
+      type = with types; nullOr lines;
+      default = null;
+    };
+    __toString = mkOption {
+      type = with types; functionTo str;
+      readOnly = true;
+      default = self: ''
+        <div class="code-block">
```
fricklerhandwerk, line 30: Likely no need to wrap it in a div, because whatever style is applied to `div.code-block` we can just apply to `pre.shell`.

overview/content-types/shell-session.nix
```diff
@@ -0,0 +1,41 @@
...
+  name,
+  config,
+  lib,
+  ...
+}:
+let
+  inherit (lib) mkOption types;
+  command = {
+    bash = mkOption {
+      type = types.submodule ./bash-code.nix;
```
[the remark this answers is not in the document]
eljamm: This is still needed to render individual commands, but we can refactor this in a followup PR.

overview/content-types/commands.nix
eljamm: I think it would be best to discuss this more (maybe tomorrow) for what's the best way to represent the commands and then we can make a follow-up PR that implements that.

eljamm (commented):
I'd say that the changes in this PR are sufficient for the implementation of copy buttons and iterate over this by discussing refactoring command representation, then doing that in later PRs (session: list of commands, individual commands, ...).

eljamm (approved):

PR #1163, opened 2025-06-15
demo shell: display usage instruction

themadbit:
Display packages in the shell. This is useful for applications that have/need multiple packages. i.e holo --holo-cli, hold

image


overview/demo/shell.nix
```diff
@@ -32,7 +32,31 @@ let
         + ''
           export PS1="\[\033[1m\][demo-shell]\[\033[m\]\040\w >\040"
 
-          ${pkgs.lib.getExe pkgs.bash} --norc "$@"
+          ${lib.optionalString (demo-shell != { }) ''
```
eljamm, line 35: I don't think `demo-shell` would ever be empty, so this is redundant.

```suggestion
```

overview/demo/shell.nix
```diff
@@ -32,7 +32,31 @@ let
...
+            ${lib.concatStringsSep "\n" (
+              lib.mapAttrsToList (
+                name: value:
+                lib.concatStringsSep "\n" (
+                  lib.mapAttrsToList (progName: prog: "echo '  - ${progName} (from ${name})'") value.programs
+                )
+              ) demo-shell
+            )}
+            # echo ""
+          ''}
```
eljamm, line 46:
```suggestion
```

overview/demo/shell.nix
```diff
@@ -32,7 +32,31 @@ let
...
           export PS1="\[\033[1m\][demo-shell]\[\033[m\]\040\w >\040"
 
-          ${pkgs.lib.getExe pkgs.bash} --norc "$@"
+          ${lib.optionalString (demo-shell != { }) ''
+            echo -e "\n\033[1;32mDemo shell activated! Available programs:\033[0m"
+            ${lib.concatStringsSep "\n" (
+              lib.mapAttrsToList (
+                name: value:
+                lib.concatStringsSep "\n" (
+                  lib.mapAttrsToList (progName: prog: "echo '  - ${progName} (from ${name})'") value.programs
```
eljamm, line 41: I feel like it would be enough to print the executables without specifying which module they come from, unless we do something fancy like:

```
- xrsh
  - xrsh
- foobar
  - foobar-cli
  - foobard
```

But this can get a bit repetitive, nonetheless.

overview/demo/shell.nix
```diff
@@ -32,7 +32,31 @@ let
...
+            # echo ""
+          ''}
+
+          # Display instructions if any exist
+          ${lib.concatStringsSep "\n" (
+            lib.mapAttrsToList (
+              name: value:
+              lib.optionalString (value.usage-instructions != "") ''
+                echo -e "\n\033[1;34m=== ${name} Demo Usage Instructions ===\033[0m"
+                echo -e "${lib.escape [ "\"" "\\" ] value.usage-instructions}"
```
eljamm, line 54: We'll probably need to figure out a way to escape this properly.

overview/demo/shell.nix
```diff
@@ -59,6 +83,15 @@ in
             };
             default = { };
           };
+          usage-instructions = mkOption {
+            type = lines;
```
eljamm, line 87: What if we make this a function that takes string and formats it to print the `Demo Usage Instructions` block?

overview/demo/shell.nix
```diff
@@ -32,7 +32,31 @@ let
         + ''
           export PS1="\[\033[1m\][demo-shell]\[\033[m\]\040\w >\040"
 
-          ${pkgs.lib.getExe pkgs.bash} --norc "$@"
+          ${lib.optionalString (demo-shell != { }) ''
+            echo -e "\n\033[1;32mDemo shell activated! Available programs:\033[0m"
+            ${lib.concatStringsSep "\n" (
+              lib.mapAttrsToList (
```
eljamm, line 38: Also, we already have the programs in `runtimeInputs`, so we can just iterate over that.

PR #1170, opened 2025-06-16
overview: refactor nix-config

eljamm:
- Re-use options from the `nix.settings` module
- Compose those into a `NIX_CONFIG` env variable


overview/content-types/nix-config.nix
```diff
@@ -1,29 +1,77 @@
...
+            mkKeyValue = k: v: "${escape [ "=" ] k} = ${mkValueString v}";
+
+            mkKeyValuePairs = attrs: concatStringsSep "\n" (mapAttrsToList mkKeyValue attrs);
+
+            isExtra = key: hasPrefix "extra-" key;
+          in
+          "export NIX_CONFIG='${lib.trim ''
+            ${mkKeyValuePairs (filterAttrs (key: value: !(isExtra key)) self.settings)}
+            ${mkKeyValuePairs (filterAttrs (key: value: isExtra key) self.settings)}
+          ''}'";
```
fricklerhandwerk, line 60:
```suggestion
          lib.trim ''
            ${mkKeyValuePairs (filterAttrs (key: value: !(isExtra key)) self.settings)}
            ${mkKeyValuePairs (filterAttrs (key: value: isExtra key) self.settings)}
          '';
```

And let the `export NIX_CONFIG` be part of the shell command inside the instructions.

eljamm: Ah, right. This makes more sense.

fricklerhandwerk (approved):

Issue #1172, opened 2025-06-17
Implement NixOS module for NodeBB service


eljamm:
### Instructions

Use metadata about the application from its tracking issue:

- #1094

and implement a NixOS module for the service.

Issue #1186, opened 2025-06-18
Implement NixOS VM test for NodeBB example


eljamm:
### Instructions

Use metadata about the application from its tracking issue:

- #1094

and implement a NixOS VM test for the service example.

eljamm:
@Prince213 will look into using https://docs.nodebb.org/api/write/#tag/users for testing the basic functionality of the program.

Prince213:
The API is undocumented and thus unstable so we should not rely on that.  Closing as basic tests are added.

PR #1201, opened 2025-06-19
projects/NodeBB: init module and add basic example

Prince213:
Closes https://github.com/ngi-nix/ngipkgs/issues/1172.

Related to:
- https://github.com/ngi-nix/ngipkgs/issues/1186
- https://github.com/ngi-nix/ngipkgs/issues/1094

### DIFF
--- a/projects/NodeBB/services/nodebb/examples/postgresql.nix
+++ b/projects/NodeBB/services/nodebb/examples/postgresql.nix
@@ -1,0 +1,33 @@
+{ pkgs, ... }:
+
+{
+  services.nodebb = {
+    enable = true;
+    admin = {
+      username = "admin";
+      email = "admin@example.com";
+      # Do *NOT* do this in production!
+      passwordFile = pkgs.writeText "nodebb-admin-password" "nodebb";
+    };
+    settings.database = "postgres";
+    # Do *NOT* do this in production!
+    databasePasswordFile = pkgs.writeText "postgresql-password" "nodebb";
+  };
+
+  systemd.services.nodebb.after = [ "postgresql.service" ];
+
+  services.postgresql = {
+    enable = true;
+    ensureDatabases = [ "nodebb" ];
+    ensureUsers = [
+      {
+        name = "nodebb";
+        ensureDBOwnership = true;
+      }
+    ];
+    # Do *NOT* do this in production!
+    initialScript = pkgs.writeText "init-sql-script" ''
+      CREATE ROLE nodebb LOGIN PASSWORD 'nodebb';
+    '';
+  };
+}
--- a/projects/NodeBB/services/nodebb/examples/postgresql.nix
+++ b/projects/NodeBB/services/nodebb/examples/postgresql.nix
@@ -3,6 +3,7 @@
 {
   services.nodebb = {
     enable = true;
+    enableLocalDB = true;
     admin = {
       username = "admin";
       email = "admin@example.com";
@@ -14,20 +15,8 @@
     databasePasswordFile = pkgs.writeText "postgresql-password" "nodebb";
   };
 
-  systemd.services.nodebb.after = [ "postgresql.service" ];
-
-  services.postgresql = {
-    enable = true;
-    ensureDatabases = [ "nodebb" ];
-    ensureUsers = [
-      {
-        name = "nodebb";
-        ensureDBOwnership = true;
-      }
-    ];
-    # Do *NOT* do this in production!
-    initialScript = pkgs.writeText "init-sql-script" ''
-      CREATE ROLE nodebb LOGIN PASSWORD 'nodebb';
-    '';
-  };
+  # Do *NOT* do this in production!
+  services.postgresql.initialScript = pkgs.writeText "init-sql-script" ''
+    CREATE ROLE nodebb LOGIN PASSWORD 'nodebb';
+  '';
 }
--- a/projects/NodeBB/services/nodebb/examples/redis.nix
+++ b/projects/NodeBB/services/nodebb/examples/redis.nix
@@ -1,0 +1,24 @@
+{ config, pkgs, ... }:
+
+{
+  services.nodebb = {
+    enable = true;
+    admin = {
+      username = "admin";
+      email = "admin@example.com";
+      # Do *NOT* do this in production!
+      passwordFile = pkgs.writeText "nodebb-admin-password" "nodebb";
+    };
+    settings.database = "redis";
+    databasePasswordFile = config.services.redis.servers."nodebb".requirePassFile;
+  };
+
+  systemd.services.nodebb.after = [ "redis-nodebb.service" ];
+
+  services.redis.servers."nodebb" = {
+    enable = true;
+    port = 6379;
+    # Do *NOT* do this in production!
+    requirePassFile = pkgs.writeText "redis-password" "nodebb";
+  };
+}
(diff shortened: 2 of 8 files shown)


projects/NodeBB/services/nodebb/examples/postgresql.nix
```diff
@@ -0,0 +1,32 @@
+{ pkgs, ... }:
+
+{
+  services.nodebb = {
+    enable = true;
+    admin = {
+      username = "admin";
+      email = "admin@example.com";
+      passwordFile = pkgs.writeText "nodebb-admin-password" "nodebb";
```
fricklerhandwerk, line 9: Let's add a big fat WARNING here that this is merely for demonstration and people should use proper secret management.

projects/NodeBB/services/nodebb/examples/postgresql.nix
```diff
@@ -0,0 +1,32 @@
...
+    };
+    database = {
+      type = "postgres";
+      passwordFile = pkgs.writeText "postgresql-password" "nodebb";
+    };
+  };
+
+  systemd.services.nodebb.after = [ "postgresql.service" ];
```
eljamm, line 17: We can set this in the nodebb module if the `database.type` is "postgres". Same for the `services.postgresql` config.

projects/NodeBB/services/nodebb/examples/redis.nix
```diff
@@ -0,0 +1,24 @@
...
+      email = "admin@example.com";
+      passwordFile = pkgs.writeText "nodebb-admin-password" "nodebb";
+    };
+    database = {
+      type = "redis";
+      passwordFile = config.services.redis.servers."nodebb".requirePassFile;
+    };
+  };
+
+  systemd.services.nodebb.after = [ "redis-nodebb.service" ];
```
eljamm, line 17: Same comment as postgres.

projects/NodeBB/services/nodebb/module.nix
```diff
@@ -6,13 +6,180 @@
...
+        fi
+
+        databasePasswordKey='."${cfg.database.type}:password"'
+        databasePassword="$(<${cfg.database.passwordFile})"
+        ${lib.getExe pkgs.jq} "$databasePasswordKey = \"$databasePassword\"" $CONFIG > $CONFIG.new
+        mv $CONFIG.new $CONFIG
+
+        export NODEBB_ADMIN_USERNAME="${cfg.admin.username}"
+        export NODEBB_ADMIN_PASSWORD="$(<${cfg.admin.passwordFile})"
+        export NODEBB_ADMIN_EMAIL="${cfg.admin.email}"
```
eljamm, line 170: We can use [systemd-credentials](https://systemd.io/CREDENTIALS/) for this.

projects/NodeBB/services/nodebb/module.nix
```diff
@@ -6,13 +6,180 @@
...
+      wantedBy = [ "multi-user.target" ];
+    };
+
+    users = {
+      groups.${cfg.group} = { };
+      users.${cfg.user} = {
+        inherit (cfg) group;
+        isSystemUser = true;
+      };
+    };
```
eljamm, line 183: Maybe we could run the service as a [DynamicUser](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#DynamicUser=).

projects/NodeBB/services/nodebb/tests/postgresql.nix
```diff
@@ -0,0 +1,38 @@
...
+      };
+  };
+
+  testScript =
+    { ... }:
+    ''
+      start_all()
+
+      machine.wait_for_unit("nodebb.service")
+      machine.wait_for_open_port(4567)
```
eljamm, line 34: Let's get the port from the example config instead of hardcoding it.

projects/NodeBB/services/nodebb/module.nix
```diff
@@ -6,13 +6,180 @@
...
+      passwordFile = lib.mkOption {
+        type = lib.types.path;
+        description = "Path to a file containing the database password.";
+      };
+
+      database = lib.mkOption {
+        type = lib.types.either lib.types.int lib.types.str;
+        default = if cfg.database.type == "redis" then 0 else "nodebb";
+        defaultText = lib.literalExpression ''if cfg.database.type == "redis" then 0 else "nodebb"'';
+      };
```
eljamm, line 112: This is apparently the cause of the [overview failure](https://buildbot.ngi.nixos.org/#/builders/196/builds/2045).

projects/NodeBB/services/nodebb/examples/postgresql.nix
fricklerhandwerk: The example must be self-contained, how else would we guarantee it to work on the consumer end? If we leave it open there's a huge gap in the instructions to get it to run

projects/NodeBB/services/nodebb/module.nix
Prince213: I think it's a overview issue, we should use defaultText if that's present.  I'm not sure though.

https://github.com/ngi-nix/ngipkgs/blob/ff7e36febc9f0d78632f50fc630a8ae399d93f98/overview/default.nix#L156-L159

eljamm: Not sure either, but I still don't see this option as something users will need to change.

projects/NodeBB/services/nodebb/module.nix
```diff
@@ -6,13 +6,180 @@
...
+      description = "The user to run NodeBB under.";
+    };
+
+    group = lib.mkOption {
+      type = lib.types.str;
+      default = "nodebb";
+      description = "The group to run NodeBB under.";
+    };
+
+    url = lib.mkOption {
```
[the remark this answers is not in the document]
Prince213: Yeah that's okay.

projects/NodeBB/services/nodebb/examples/postgresql.nix
Prince213: Users may use PostgreSQL from different sources (e.g. Docker) so we would want the user to decide what to use here.

projects/NodeBB/services/nodebb/module.nix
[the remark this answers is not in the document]
fricklerhandwerk: Well the work is done already, and arguably fine-grained fields are easier to use from a NixOS perspective since they're documented at the level you're configuring the service. I think it's fine to keep here.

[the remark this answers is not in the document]
Prince213: We can have both to allow extra settings.

projects/NodeBB/services/nodebb/module.nix
```diff
@@ -6,13 +6,260 @@
...
+      # nodebb will try redis if in config.json even if postgres is selected
+      excludeIfNot = value: if cfg.settings.database != value then [ value ] else [ ];
+      settings = lib.removeAttrs cfg.settings (
+        (excludeIfNot "mongo") ++ (excludeIfNot "postgres") ++ (excludeIfNot "redis")
+      );
+    in
+    settingsFormat.generate "config.json" (
+      # { a = { b = 0; }; } -> { "a:b" = 0; }
+      # https://github.com/NodeBB/NodeBB/blob/3e961257ec0904dbc3b3c64dab3d4cbdffcfbbd7/src/install.js#L177
+      settings // (mapAttrsRecursive' (path: lib.nameValuePair (lib.concatStringsSep ":" path)) settings)
```
Prince213, line 38: We could use this but the `lib` here doesn't have it.

```suggestion
      settings // (lib.flattenAttrs ":" settings)
```

eljamm (approved):
LGTM, awesome work, @Prince213!

Issue #1190, opened 2025-06-18
Create demo type in the data model


eljamm:
Tests should be required (or null to indicate that we need them).

Issue #1139, opened 2025-06-12
US: Prevent exposing broken demos to users


eljamm:
### Description

As an NGIpkgs user,
I want to be sure that a demo works before following the instructions,
In order to avoid wasting my time and resources.

### Acceptance criteria

Given a project with a demo,
When I navigate to the project's demo section in the overview,
Then the demo is guaranteed to work as displayed.

## Implementation notes

- Tests are required for demos (or nullable to leave a todo)
- The datamodel should prevent rendering demos without tests
- There's a way to mark a demo as broken (with a reason, possibly following the gist behind https://github.com/NixOS/rfcs/pull/127)
  - Filter out broken tests in `checks`

PR #1210, opened 2025-06-20
types: add demo; init problems

eljamm:
Closes: https://github.com/ngi-nix/ngipkgs/issues/1190
Related to: https://github.com/ngi-nix/ngipkgs/issues/1139

### DIFF
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -254,23 +254,22 @@ let
       <article class="page-width">
         ${heading 1 null name}
         ${render.metadata.one project.metadata}
-        ${optionalString (project.nixos.examples ? demo) (
-          render.serviceDemo.one "vm" project.nixos.modules project.nixos.examples.demo
-        )}
-        ${optionalString (project.nixos.examples ? demo-shell) (
-          render.serviceDemo.one "shell" project.nixos.modules project.nixos.examples.demo-shell
+        ${optionalString (project.nixos.demo != { }) (
+          lib.concatMapAttrsStringSep "\n" (
+            type: demo: (render.serviceDemo.one type project.nixos.modules demo)
+          ) project.nixos.demo
         )}
         ${render.options.many (pick.options project)}
         ${render.examples.many (pick.examples project)}
       </article>
     '';
 
-    demoGlue.one = demo-function: exampleText: ''
+    demoGlue.one = type: exampleText: ''
       # default.nix
       {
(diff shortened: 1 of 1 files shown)


projects/types.nix
```diff
@@ -231,6 +278,10 @@ let
                           type = nullOr (attrsOf (nullOr types'.service));
                           default = null;
                         };
+                        demo = lib.mkOption {
+                          type = nullOr (attrsOf types'.demo);
```
eljamm, line 282: Some design choices for `nixos.demo`:

1. attrTag of `vm` or `shell`. Only one can be set at a time.
1. `attrsOf types'.demo`, with restriction on kind (attr name should be either enum of `vm` or `shell`)
1. `types'.demo` with `kind` (vm or demo) -> means we'll only have 1 demo for a project, so we can't have multiple ones for service and shells

projects/types.nix
```diff
@@ -196,6 +196,53 @@ let
...
+            links
+            ;
+          inherit (types')
+            problems
+            ;
+          kind = mkOption {
+            type = types.enum [
+              "vm"
+              "shell"
+            ];
```
fricklerhandwerk, line 216: I prefer an explicitly tagged union with `attrTag`

projects/types.nix
```diff
@@ -196,6 +196,53 @@ let
...
+            "broken"
+          ];
+        };
+        message = mkOption {
+          type = types.str;
+        };
+        urls = mkOption {
+          type = with types; listOf str;
+          default = [ ];
+        };
```
fricklerhandwerk, line 242: YAGNI for now -- unless you already do need it for something specific?

projects/types.nix
fricklerhandwerk: AttrTag can only be one, and it allows clearly separate types for each tag. Always prefer that as a "structured enum" (aka sum type)

projects/types.nix
```diff
@@ -196,6 +196,53 @@ let
...
+      }
+    );
+
+    problems = lib.mkOption {
+      type = types.attrsOf types'.problem;
+      default = { };
+    };
+
+    problem = types.submodule {
```
fricklerhandwerk, line 228: Same here, why not `problem.type = attrTag ...`? And then only define what we know so far: broken with reason.

fricklerhandwerk (approved):
Nice, does it work already?

fricklerhandwerk (approved):

PR #1219, opened 2025-06-20
projects/PeerTube: add update scripts to and update plugins

JuneStepp:
This PR adds updates all PeerTube plugins/themes. It also adds update scripts to all of them except livechat. I'm leaving livechat for now, since I think its build process will get simpler in the future.

All the official PeerTube plugins live in the same repository with no tags. Each commit could update any one of the plugins. Due to the use of sparseCheckout, each plugin will only get updated when its files have changed. Otherwise, the update script will give an error.
Once https://github.com/Mic92/nix-update/issues/281 is resolved, the plugins will be able to use nix-update instead of the custom script.

Finally, the update script is referenced with `peertube-plugin-akismet.peertubeOfficialPluginsUpdateScript`, because there was no better place currently to put the update script than one of the plugin package directories.

### DIFF
--- a/pkgs/by-name/peertube-plugin-livechat/package.nix
+++ b/pkgs/by-name/peertube-plugin-livechat/package.nix
@@ -4,42 +4,40 @@
   fetchNpmDeps,
   fetchFromGitHub,
   runCommand,
-  stdenvNoCC,
-  symlinkJoin,
   prosody,
 }:
 let
   details = {
     livechat = rec {
       pname = "peertube-plugin-livechat";
-      version = "10.1.2";
+      version = "13.0.0";
       src = fetchFromGitHub {
         owner = "JohnXLivingston";
         repo = "peertube-plugin-livechat";
         rev = "refs/tags/v${version}";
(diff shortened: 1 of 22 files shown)


pkgs/by-name/peertube-plugin-livechat/package.nix
```diff
@@ -202,8 +202,53 @@ let
...
+        (p.callPackage (
+          {
+            buildLuarocksPackage,
+            fetchFromGitHub,
+            fetchurl,
+            luaOlder,
+            oniguruma,
+          }:
+          buildLuarocksPackage {
+            pname = "lrexlib-oniguruma";
```
eljamm, line 219: I'd add a TODO with the Nixpkgs PR link here, so we don't forget to switch to it afterwards, but this looks good otherwise.

eljamm (dismissed):

JuneStepp:
The 

> I've opened a nixpkgs PR for lrexlib-oniguruma as needed by the livechat plugin: [NixOS/nixpkgs#418862](https://github.com/NixOS/nixpkgs/pull/418862)
> 
> It's manually packed in the livechat package file for now.

Using upstream now.

Issue #1206, opened 2025-06-19
Implement Demo for NodeBB service


eljamm:
### Instructions

Use metadata about the application from its tracking issue:

- #1094 

and implement a demo for the service.

PR #1221, opened 2025-06-22
projects/NodeBB: add demo vm

Prince213:
Resolves https://github.com/ngi-nix/ngipkgs/issues/1206.

Closes https://github.com/ngi-nix/ngipkgs/issues/1094.

### DIFF
--- a/projects/NodeBB/default.nix
+++ b/projects/NodeBB/default.nix
@@ -27,22 +27,29 @@
     };
   };
 
-  nixos.modules.services = {
-    nodebb = {
-      name = "NodeBB";
-      module = ./services/nodebb/module.nix;
-      examples = {
-        postgresql = {
-          module = ./services/nodebb/examples/postgresql.nix;
-          description = "";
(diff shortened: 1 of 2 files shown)


projects/NodeBB/default.nix
```diff
@@ -27,22 +27,29 @@
...
+          redis = {
+            module = ./services/nodebb/examples/redis.nix;
+            description = "";
+            tests.redis = import ./services/nodebb/tests/redis.nix args;
+          };
         };
       };
     };
+    demo.vm = {
+      module = ./services/nodebb/examples/postgresql.nix;
```
eljamm, line 50: We should probably add an `openPorts` option.

```nix
    networking.firewall.allowedTCPPorts = [ 4567 ];
```

**PS:** The service takes some time to start. We probably want a way to handle this in the future, which could be as simple as warning the users or displaying a live feed or something.

eljamm (approved):

Prince213:
Thanks!

Issue #1207, opened 2025-06-19
Implement demo for Draupnir service


eljamm:
### Instructions

Use metadata about the application from its tracking issue:

- #1068

and implement a demo for the service.

Issue #1068, opened 2025-06-02
NGI Project: Draupnir


themadbit:
### Short Summary

Moderation bot for Matrix servers

### NLnet page(s)

https://nlnet.nl/project/Draupnir/

### Website

https://the-draupnir-project.github.io/draupnir-documentation/

### Source repository

- https://github.com/the-draupnir-project/Draupnir
  - Language/Framework: Typescript
  - Dependency management: yarn, Node
  - Nix development environment: n/a

### Documentation

- Usage Examples:
  - Moderator's guide: https://the-draupnir-project.github.io/draupnir-documentation/category/moderators-guide
- Build from source/Development:
  - https://the-draupnir-project.github.io/draupnir-documentation/bot/setup
- Other:
  - [Contributing](https://the-draupnir-project.github.io/draupnir-documentation/category/contributing)

### Artefacts

- Appservice:
  - https://the-draupnir-project.github.io/draupnir-documentation/appservice

### NGIpkgs

- ngi-nix repository
  - none

### Nixpkgs/NixOS

- Packages:
  - [draupnir](https://github.com/NixOS/nixpkgs/blob/nixos-25.05/pkgs/by-name/dr/draupnir/package.nix#L101)
- Services:
  - [`services.draupnir`](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/services/matrix/draupnir.nix)

### Tasks

[Add a sub-issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/adding-sub-issues#creating-a-sub-issue) using the appropriate GitHub template for each of the following tasks:

- [x] [Triage project](https://github.com/ngi-nix/ngipkgs/issues/new?template=task-triage.yaml)
- [x] [Implement project metadata](https://github.com/ngi-nix/ngipkgs/issues/new?template=task-metadata.md)

Next, depending on the project deliverable, open sub-issues for the following tasks:

- [ ] [Implement NixOS program module](https://github.com/ngi-nix/ngipkgs/issues/new?template=task-module-program.md)
- [x] [Implement NixOS service module](https://github.com/ngi-nix/ngipkgs/issues/new?template=task-module-service.md)

TheArcaneBrony:
Nixpkgs PR merged as of 20 minutes ago.

PR #1222, opened 2025-06-22
projects/Draupnir: add demo vm

Prince213:
Resolves https://github.com/ngi-nix/ngipkgs/issues/1207.

Closes https://github.com/ngi-nix/ngipkgs/issues/1068.


projects/Draupnir/default.nix
```diff
@@ -32,5 +32,10 @@
         tests.basic = "${sources.inputs.nixpkgs}/nixos/tests/matrix/draupnir.nix";
       };
     };
+    demo.vm = {
+      module = ./example.nix;
```
eljamm, line 36: There is no option to open the ports in the firewall, so we need to either create and enable that or add:

```nix
    # open service ports in the firewall
    networking.firewall.allowedTCPPorts = [ 8008 ];
```

In order for the synapse instance to be [forwarded to the host](https://github.com/ngi-nix/ngipkgs/blob/96154f436225554038061dd26d007e13a8b3d29a/overview/demo/vm/virtualisation.nix#L13-L19).

projects/Draupnir/default.nix
```diff
@@ -32,5 +32,10 @@
...
+      description = "Deployment for demo purposes";
```
eljamm, line 37: Is there a specific workflow for testing draupnir?

**PS:** If there isn't, this project is actually a good candidate for https://github.com/ngi-nix/ngipkgs/issues/1139

Prince213: There isn't one at the moment and the only tests are the one in nixpkgs.  Maybe we shouldn't expose the demo vm.

eljamm: > Maybe we shouldn't expose the demo vm.

Yes, but I believe we can still add one and keep refining it. That's the idea behind https://github.com/ngi-nix/ngipkgs/issues/1139.

projects/Draupnir/example.nix
```diff
@@ -38,12 +41,15 @@
   services.draupnir = {
     enable = true;
     settings = {
-      homeserverUrl = "http://localhost:8008";
+      homeserverUrl = "http://localhost:${builtins.toString port}";
```
eljamm, line 44: Quick tip: you can use some builtin functions without the `builtins` keyword and `toString` is one of them.

projects/Draupnir/default.nix
```diff
@@ -32,5 +32,10 @@
...
+      tests.basic = "${sources.inputs.nixpkgs}/nixos/tests/matrix/draupnir.nix";
```
eljamm, line 38: So we don't display the unfinished config (see https://github.com/ngi-nix/ngipkgs/issues/1139)

```suggestion
      tests.basic = "${sources.inputs.nixpkgs}/nixos/tests/matrix/draupnir.nix";
      problem.broken.reason = ''
        Still a work in progress.
      '';
```

Prince213:
Thanks!

Issue #1006, opened 2025-05-07
US: Split program and services options into separate headings on the overview site


erictapen:
As a user of NGIpkgs, I want to understand service options seperately from program options, as I'll probably only use one of the two at the same time (e.g. only the client application or only the service).

PR #1204, opened 2025-06-19
projects/Holo: implement examples and VM test

themadbit:
closes #1202 
closes #1242

Issue #1031, opened 2025-05-16
Represent missing deliverables in the overview


eljamm:
We need to represent missing deliverables using the [content type](https://github.com/ngi-nix/ngipkgs/issues/1018) in order to render them.

PR #1249, opened 2025-06-25
overview: use evaluated-modules.config.projects

JuneStepp:
The overview previously used `projects` from default.nix that had been processed through `make-projects`. This changed the projects structure in unexpected ways and removed necessary information for changes such as showing missing deliverables.

This PR switches to using `evaluated-modules`. To make this easier, the usage of null in the module system was made stricter. Now, to mark something as not implemented but needed, `.module` should be set to null.

There are a couple other improvements in this PR. Most notably that options for separate programs/services now render in separate blocks. This fixes the bug where until expanded, there appeared to only be options for the first program/service. It also at least partially addresses #1006.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -35,6 +35,13 @@ let
       in
       f "";
 
+    filter-map =
+      attrs: input:
+      lib.pipe attrs [
+        (lib.concatMapAttrs (_: value: value."${input}" or { }))
+        (lib.filterAttrs (_: v: v != null))
+      ];
+
     # Recursively evaluate attributes for an attribute set.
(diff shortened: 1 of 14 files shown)


fricklerhandwerk (approved):
LGTM, thanks! A possible next step would be extracting the project detail view into a separate submodule, starting with any of its constituents such as module options.

eljamm:
Solved conflict from https://github.com/ngi-nix/ngipkgs/pull/1204

erictapen:
Addressed #1031

PR #1255, opened 2025-06-26
overview: introduce darkmode via css mediaquery

erictapen:
this is super basic and is supposed to serve as a base to get iterated on top of

Response to @eljamm's desire to not get flashbanged in the face everyday.

tmp BFbkHrwLoy

tmp 5cmICwMJiq


fricklerhandwerk (approved):
Let's move to CSS variables next.

PR #1257, opened 2025-06-26
Refactor tests

eljamm:
### Changes

- treewide: use pkgs.nixosTests for exported tests
- Refactor test type
- treewide: use module attribute in tests

A next step would be to properly handle broken tests by skipping them in CI (as opposed to having to comment out the code).


fricklerhandwerk (dismissed):

eljamm:
Broken project tests are now marked with the problem type (with reason) and filtered out.

PR #1258, opened 2025-06-26
treewide: use pkgs.nixosTests for exported tests

eljamm:
Similarly to https://github.com/ngi-nix/ngipkgs/issues/960

### DIFF
--- a/projects/OpenWebCalendar/default.nix
+++ b/projects/OpenWebCalendar/default.nix
@@ -15,8 +15,7 @@
         description = "";
         # FIX: re-enable after this is solved:
         # https://github.com/NixOS/nixpkgs/issues/418689
-        # tests.basic = "${sources.inputs.nixpkgs}/nixos/tests/web-apps/open-web-calendar.nix";
-        tests.basic = null;
+        # tests.basic = pkgs.nixosTests.open-web-calendar;
       };
       links = {
         development = {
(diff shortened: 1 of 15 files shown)


projects/OpenWebCalendar/default.nix
```diff
@@ -50,8 +49,7 @@
       description = "Deployment for demo purposes";
       # FIX: re-enable after this is solved:
       # https://github.com/NixOS/nixpkgs/issues/418689
-      # tests.basic = "${sources.inputs.nixpkgs}/nixos/tests/web-apps/open-web-calendar.nix";
-      tests.basic = null;
+      # tests.basic = pkgs.nixosTests.open-web-calendar;
```
fricklerhandwerk, line 52: woulnd't we apply the `broken` annotation here?

eljamm: We can do that after https://github.com/ngi-nix/ngipkgs/pull/1257. I took this change from that PR to narrow down the CI failure scope (which we only need to disable the Pixelfed test in aarch64, since it's [not supported](https://github.com/NixOS/nixpkgs/blob/53b8a4618934db6cbcc02359ccfd49ef034f782a/nixos/tests/web-apps/pixelfed/default.nix#L6-L9)).

PR #1264, opened 2025-06-27
projects/holo: add demo VM 

### LINKED ISSUE #1250: Implement a demo example for Holo
### Instructions

Use metadata about the application from its tracking issue:

- #338

and implement a demo.

themadbit:
closes #1250

### DIFF
--- a/projects/holo/demo/vm.nix
+++ b/projects/holo/demo/vm.nix
@@ -1,9 +1,0 @@
-{ ... }:
-{
-  programs.holo.enable = true;
-  services.holo-daemon.enable = true;
-
-  #TODO: Move this to the module confgiguration
-  #      It is only needed for the demo
-  services.getty.autologinUser = "root";
-}
--- a/projects/holo/default.nix
+++ b/projects/holo/default.nix
@@ -39,8 +39,28 @@
   };
 
   nixos.demo.vm = {
-    module = ./demo/vm.nix;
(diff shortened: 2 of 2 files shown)


projects/holo/demo/vm.nix
```diff
@@ -0,0 +1,7 @@
+{ ... }:
+{
+  programs.holo.enable = true;
+  services.holo-daemon.enable = true;
+
+  services.getty.autologinUser = "root";
```
eljamm, line 6: It would probably be better to set this inside the service module, but only enable it when we're running inside a demo VM.

themadbit: I'm not sure how to do the latter

projects/holo/demo/vm.nix
```diff
@@ -0,0 +1,9 @@
+{ ... }:
+{
+  programs.holo.enable = true;
+  services.holo-daemon.enable = true;
+
+  #TODO: Move this to the module confgiguration
+  #      It is only needed for the demo
+  services.getty.autologinUser = "root";
```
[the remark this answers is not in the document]
eljamm: We could also do [something like this](https://github.com/NixOS/nixpkgs/blob/7321d5dc7b0b2722d03c67bc8851acd014ec446c/nixos/modules/services/misc/omnom.nix#L246-L254) and wrap the executable with the service user.

PR #1288, opened 2025-07-02
overview: add content-type for option

eljamm:


overview/content-types/option.nix
```diff
@@ -0,0 +1,123 @@
...
+          ${markdownToHtml config.description}
+          </div>
+        '';
+    };
+    alert-readonly = mkOption {
+      type = types.str;
+      readOnly = true;
+      default = optionalString config.readOnly ''
+        <span class="option-alert" title="This option can't be set by users">Read-only</span>
+      '';
```
fricklerhandwerk, line 88: Should simply be part of rendering the default value,  I see no reason to have a separate field.

overview/content-types/option.nix
```diff
@@ -0,0 +1,123 @@
...
+    };
+    option-description = mkOption {
+      type = types.str;
+      readOnly = true;
+      default =
+        let
+          # This doesn't actually produce a HTML string but a Jinja2 template string
+          # literal, that is then replaced by it's HTML translation at the last build
+          # step.
```
fricklerhandwerk, line 74: Add the rationale: this avoids IFD since this way we can do markdown processing in a single step at the end

overview/content-types/option.nix
eljamm: This makes the `__toString` more readable IMO, but I assume that this comment also applies to the other options as well, right?

overview/content-types/option.nix
```diff
@@ -0,0 +1,123 @@
...
+      '';
+    };
+    option-default = mkOption {
+      type = types.str;
+      readOnly = true;
+      default = optionalString (config.default ? text) ''
+        <dt>Default:</dt>
+        <dd class="option-default"><code>${config.default.text}</code></dd>
+      '';
+    };
```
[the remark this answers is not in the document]
eljamm: You're right, but I'd like to note that one advantage of this approach is that we can document what each of those blocks are supposed to do and maybe render them later on as part of documentation.

overview/default.nix
```diff
@@ -194,7 +158,7 @@ let
               lib.concatMapAttrsStringSep "\n" (
                 name: val:
                 optionalString (val.module != null) (
-                  render.options.many [ type name ] (
+                  render.options [ type name ] (
```
fricklerhandwerk, line 161: We had a draft commit from a pairing session where we extracted these mysterious lists into named values for readability, can we have that back please?

fricklerhandwerk (dismissed):
LGTM morally, but needs a bit of a refactoring still

overview/content-types/option.nix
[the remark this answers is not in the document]
fricklerhandwerk: Well you can document the actual value fields. This is merely string processing, it has no semantic value.

overview/content-types/option.nix
eljamm: I see. That makes sense, thanks!

PR #1295, opened 2025-07-03
projects/NodeBB: wait until service is ready for demo vm

### LINKED ISSUE #1262: NodeBB: Print clickable URL in demo VM
Currently the demo shell greeting says

> NodeBB needs some time to set up and will list on port 4567 when ready. 

It would be easier to figure out what to do if it said `http://localhost:4567` and only did so when the service is actually ready.

Also the demo config shows the prompt, this probably shouldn't happen:

Image

Prince213:
Resolves #1262.

### DIFF
--- a/projects/NodeBB/services/nodebb/examples/postgresql.nix
+++ b/projects/NodeBB/services/nodebb/examples/postgresql.nix
@@ -20,11 +20,4 @@
   services.postgresql.initialScript = pkgs.writeText "init-sql-script" ''
     CREATE ROLE nodebb LOGIN PASSWORD 'nodebb';
   '';
-
-  # demo-vm
-  programs.bash.interactiveShellInit = ''
-    echo "NodeBB is starting. Please wait ..."
-    until systemctl show nodebb.service | grep -q ActiveState=active; do sleep 1; done
-    echo "NodeBB is ready at http://localhost:4567"
-  '';
 }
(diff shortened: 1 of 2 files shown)


projects/NodeBB/services/nodebb/examples/postgresql.nix
```diff
@@ -22,8 +22,9 @@
 
   # demo-vm
   networking.firewall.allowedTCPPorts = [ 4567 ];
-  services.getty.helpLine = ''
-    NodeBB needs some time to set up and will list on port 4567 when ready.
-    View journal with `journalctl -efu nodebb.sevice` to see progress.
+  programs.bash.interactiveShellInit = ''
+    echo "NodeBB is starting..."
```
eljamm, line 26:
```suggestion
    echo "NodeBB is starting. Please wait ..."
```

projects/NodeBB/services/nodebb/examples/postgresql.nix
```diff
@@ -22,8 +22,9 @@
 
   # demo-vm
   networking.firewall.allowedTCPPorts = [ 4567 ];
-  services.getty.helpLine = ''
-    NodeBB needs some time to set up and will list on port 4567 when ready.
-    View journal with `journalctl -efu nodebb.sevice` to see progress.
+  programs.bash.interactiveShellInit = ''
```
eljamm, line 25: I think moving this inside the module configuration will be better. We can modify it to only be enabled inside a demo, later on.

projects/NodeBB/services/nodebb/examples/postgresql.nix
```diff
@@ -22,8 +22,9 @@
 
   # demo-vm
   networking.firewall.allowedTCPPorts = [ 4567 ];
```
eljamm, line 24: We can make an `openFirewall` option for this, or at least get the service port from `settings`.

PR #1296, opened 2025-07-03
Inventaire: open firewall; print ready message in demo 

eljamm:
## Changes

- Inventaire
  - open port in firewall so the demo service becomes accessible on host machines
  - print ready message when service starts, courtesy of @Prince213 (see https://github.com/ngi-nix/ngipkgs/pull/1295)
- Demo
  - add option to separate demo code from production

inventaire


Prince213 (approved):
Diff LGTM

Issue #782, opened 2025-04-11
Heads: Enable more boards


OPNA2608:
https://github.com/ngi-nix/ngipkgs/blob/b6d62473f51f5d51282554eed8a6daee2c22a700/pkgs/by-name/heads/package.nix#L422-L427

We should enable more boards, so we can eventually get rid of this allowlist.

Complications:

- CI seems to be very *optimistic* about how long builds may take,so it fires every build jobs with as many cores as the builder has available. This will quickly overwhelm CI and put it at 100% CPU load, make builds come to a crawl, and time out the builds due to slow output.
  Maybe the builder should adjust its `max-jobs` and `cores` parameters better?
- Currently, `fetchgit`'s `fetchSubmodules` doesn't handle submodules that define their `update` value as i.e. `none`, meaning that they're skipped during a normal `git submodules update --init --recursive`.
  Needs either https://github.com/NixOS/nixpkgs/pull/286228, or a workaround on our side

fricklerhandwerk:
The `cores = 0` is very likely the culprit, but it seems we can just override that per derivation, since `cores` merely passes an environment variable and it's the `builder` executable of the derivation that controls parallelism for itself: https://nix.dev/manual/nix/latest/command-ref/conf-file#conf-cores 

Specifically, it may be fixed using that `enableParallelBuilding` argument to `mkDerivation`.

OPNA2608:
@fricklerhandwerk Well, we could do `enableParallelBuilding = false` to *disable* parallel building, or test for `enableParallelBuilding` in `preConfigure` and reduce `NIX_BUILD_CORES` to a lower value... but both of those will also apply to anyone that's trying to build those targets *without* this configuration situation. I.e. my workstation has (or rather *had* 🙃) 48 threads, and I would often run builds with `--max-jobs 1 --cores 40` just to blast through them.

I think limiting parallelism inside the derivation makes sense if there are actual issues *within* the project when building it with a certain amount of cores (some `checkPhase`s, handwritten Makefiles, and the old from-source Pale Moon build in Nixpkgs come to mind). Limiting parallelism in a package because infrastructure is misconfigured seems like the wrong fix to me though.

fricklerhandwerk:
In that case we could indeed just reduce cores for infra... @dgrig wdyt?

OPNA2608:
[...]

PR #1298, opened 2025-07-03
makemake: Limit max-jobs and cores per nix build

Erethon:
Related to ngi-nix/ngipkgs#782.

Makemake has 16 logical cores, so 4 jobs times 4 cores will max it out. I tested this with ngi-nix/ngipkgs#1292, the [build took ~7 hours](https://buildbot.ngi.nixos.org/#/builders/5/builds/2629) and ultimately failed, but the host was responsive throughout the build. [Grafana dashboard with timestamp](https://grafana.nixos.org/d/rYdddlPWk/node-exporter-full?orgId=1&from=2025-07-02T21:43:16.804Z&to=2025-07-03T05:30:11.949Z&timezone=browser&var-datasource=default&var-job=node&var-node=makemake.ngi.nixos.org:9100&var-diskdevices=%5Ba-z%5D%2B%7Cnvme%5B0-9%5D%2Bn%5B0-9%5D%2B%7Cmmcblk%5B0-9%5D%2B) while makemake was trying to build heads.

### DIFF
--- a/infra/makemake/configuration.nix
+++ b/infra/makemake/configuration.nix
@@ -42,9 +42,9 @@
           GiB = 1024 * 1024 * 1024;
         in
         4 * GiB;
-      max-jobs = lib.mkDefault 16;
+      max-jobs = lib.mkDefault 4;
       allowed-uris = "https://github.com/ https://git.savannah.gnu.org/ github: gitlab: git+https:";
-      cores = 0;
+      cores = 4;
       experimental-features = [
         "nix-command"
         "flakes"


OPNA2608:
One of the issues in the builds was that the `linux` part of them is very quiet in terms of output. That runs into the "I haven't received any output in awhile" timeout. We can try to patch the build to be less quiet, and see if any succeed with that alone.

fricklerhandwerk (approved):

Erethon:
> Checking the Heads board PR again with this applied. max-jobs might still be 16 somehow?
> cores is definitely getting set to 4 though, so that helps with the system load.
> 
> > ```
> > heads-librem_11> Used **CPUS**: 4
> > ```

I see `max-jobs` as `4` on the host:
```
[root@makemake:~]# cat /etc/nix/nix.conf 
<snip>
cores = 4
max-jobs = 4
```

OPNA2608:
Build times still go up significantly when there's too many boards that need to be built, >2h per board vs 30 mins to 1h when there's less boards.

I think `max-jobs` only controls the amount of derivation parallelism that happens within a single `nix-build` / `nix build`, right? Looks like the buildbot `nix build`s every derivation individually, so that doesn't really work. From the very top of a build log:

```
nix build -L --option keep-going true --max-silent-time 1200 --accept-flake-config --out-link result-x86_64-linux.packages/heads.librem_mini_v2 '/nix/store/ylh9q8myqrk95zqiq6s7qpyqw629svv3-heads-librem_mini_v2-0.2.1-unstable-2025-04-03.drv^*'
```

So if we can't somehow control how many of those get launched in parallel, then reducing the workers is prolly the way to go?

(Also, it looks like the invocation overrides the max silent time from #1447, so that might need a different approach as well?)

Issue #1306, opened 2025-07-04
Implement NixOS module for owi program


ethancedwards8:
### Instructions

Use metadata about the application from its tracking issue:

- #1149 

and implement a NixOS module for the program.

> [!NOTE]
> For an example of how this looks like, please refer to the [project template](https://github.com/ngi-nix/ngipkgs/tree/main/maintainers/templates/project).

Issue #1305, opened 2025-07-04
Implement a demo example for owi


ethancedwards8:
### Instructions

Use metadata about the application from its tracking issue:

- #1149 

and implement a demo.

Do a demo of how Owi can prove two functions are equivalent, among other things

Issue #1151, opened 2025-06-13
Implement NixOS program module for OWI


eljamm:
### Instructions

Given #1149 for metadata about the application, implement a NixOS program module for OWI.

### Tasks

After or during the implementation of this module, create issues for the following deliverables using the appropriate GitHub template:

- [x] Example configuration
- [x] NixOS test
- [x] Shell demo

**Note:** Each example needs to be tested.

eljamm:
Derivation is being packaged in https://github.com/NixOS/nixpkgs/pull/414133

PR #1308, opened 2025-07-04
projects/owi: init

ethancedwards8:
Closes: https://github.com/ngi-nix/ngipkgs/issues/1151
Closes: https://github.com/ngi-nix/ngipkgs/issues/1305
Closes: https://github.com/ngi-nix/ngipkgs/issues/1306
Closes: https://github.com/ngi-nix/ngipkgs/issues/1306

Tentatively closes: https://github.com/ngi-nix/ngipkgs/issues/1149

~~**EDIT, IMPORTANT: this eval will not pass until owi makes its way into the nixpkgs-unstable channel as it was just merged recently.**~~

### DIFF
--- a/projects/owi/programs/owi/module.nix
+++ b/projects/owi/programs/owi/module.nix
@@ -1,0 +1,21 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.programs.owi;
+in
+{
+  options.programs.owi = {
+    enable = lib.mkEnableOption "owi";
+    package = lib.mkPackageOption pkgs "owi" { };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = with pkgs; [
+      cfg.package
+    ];
+  };
+}
--- a/projects/owi/default.nix
+++ b/projects/owi/default.nix
@@ -1,0 +1,34 @@
+{
+  lib,
+  pkgs,
+  sources,
+  ...
+}@args:
+
+{
+  metadata = {
+    summary = "Cross-language symbolic execution for C, C++, Rust, Zig, and Wasm";
+    subgrants = [
+      "OWI"
+      "Owi-2"
+    ];
+  };
+
+  nixos.modules.programs = {
+    owi = {
+      name = "owi";
+      module = ./programs/owi/module.nix;
+      examples.basic = {
+        module = ./programs/owi/examples/basic.nix;
+        description = "Enable the owi program";
+        tests.basic.module = import ./programs/owi/tests/shell.nix args;
+      };
+    };
+  };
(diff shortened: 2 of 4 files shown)


ethancedwards8:
See https://nixpk.gs/pr-tracker.html?pr=414133 for status on Owi reaching nixpkgs-unstable, etc

ethancedwards8:
Rebased on origin/main and force pushed.

ethancedwards8:
Fixed the formatting error

projects/owi/programs/owi/tests/basic.nix
```diff
@@ -0,0 +1,28 @@
...
+        ];
+      };
+  };
+
+  testScript =
+    { nodes, ... }:
+    ''
+      start_all()
+
+      machine.succeed("owi version")
```
fricklerhandwerk, line 26: Hm this will be a rather expensive test for something that can run in the derivation. Let's better keep that out.

projects/owi/programs/owi/module.nix
```diff
@@ -0,0 +1,21 @@
...
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.programs.owi;
+in
+{
+  options.programs.owi = {
```
fricklerhandwerk, line 11: Side note to @eljamm: this pattern is repeating itself so often, we could actually try applying the portable services idea to it. I.e. make (the equivalent of) `programs` an `attrsOf (submoduleWith { modules = [ program ]; })` and set instances as an attr with only the package specified in the simplest case. But that allows mixing in additional aspects or changing (in a central place!) what a program is made of and what API it exposes. And for consumers it allows having multiple instances of a program in a system and addressable by name, even referring to each other in their customization options.

fricklerhandwerk (approved):

projects/owi/programs/owi/tests/basic.nix
ethancedwards8: Sure thing. That's what I thought too. I included it because the majority (if not all) of the other projects in ngipkgs do this too. We might want to create an issue documenting and figuring out how to replace it. 

On that note, what should I replace it with? The shell already has some comprehensive tests.

fricklerhandwerk: @themadbit and @eljamm were discussing how to reuse samples across demos IIUC, so maybe we can align that somehow? And if not, let's just skip it for now (set to null) and figure it out later.

projects/owi/programs/owi/module.nix
eljamm: I'm not sure. Let's discuss this in more details later and we can perhaps create a User Story for it.

Issue #1327, opened 2025-07-07
Implement a demo example for jaq


ethancedwards8:
### Instructions

Use metadata about the application from its tracking issue:

- #929 

and implement a demo.

PR #1328, opened 2025-07-07
projects/jaq: add a demo

ethancedwards8:
Closes: https://github.com/ngi-nix/ngipkgs/issues/1327


fricklerhandwerk (dismissed):

projects/jaq/default.nix
```diff
@@ -26,5 +26,9 @@
     };
   };
 
-  # no service module as this is strictly a program
+  nixos.demo.shell = {
+    module = ./programs/jaq/examples/basic.nix;
+    description = "jaq example";
```
themadbit, line 31: If you have played with the demo, it would be better to have the description as usage instructions for using the demo, rather than a title description. See https://github.com/ngi-nix/ngipkgs/blob/main/projects/holo/default.nix#L43-L63

ethancedwards8: Good idea. I updated it. Thanks!

themadbit (approved):

PR #1338, opened 2025-07-09
overview: fix update script status for scopes

### LINKED ISSUE #1326: Bug: ngi.nixos.org does not correctly detect update scripts in python libraries
`buildPythonApplication` has an updateScript that is automatically included in every instance/python library in Nixpkgs.

See
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/interpreters/python/mk-python-derivation.nix#L399-L422
and
https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/up/update-python-libraries/update-python-libraries.py

However, ngi.nixos.org does not correctly detect this updateScript and instead requires that there be a `passthru.updateScript` set in the upstream derivation.

See related: https://github.com/NixOS/nixpkgs/pull/423104#pullrequestreview-2992331357

We need to edit our implementation of ngi.nixos.org to detect the default updateScript for python libraries.

eljamm:
Selecting nested attributes doesn't work with strings (for example: `pkgs."python3Packages.dmt-core"`), which results in update scripts wrongly marked as missing (see #1326).

This change corrects that by checking the derivation's attribute path using `lib.attrByPath`.

Closes #1326

image

### DIFF
--- a/overview/content-types/option.nix
+++ b/overview/content-types/option.nix
@@ -84,11 +84,15 @@ in
           alert-update-script =
             let
               isDrv = self.type == "package";
-              optionName = lib.removePrefix "pkgs." self.default.text;
+              attrPath = lib.splitString "." self.default.text ++ [
+                "passthru"
+                "updateScript"
+              ];
+              updateScript = lib.attrByPath attrPath null pkgs;
             in
             # TODO: plug a function that recurses into all dependencies,
             # so we'd know how much of the build graph is kept up to date automatically
-            optionalString (isDrv && !pkgs ? ${optionName}.passthru.updateScript) ''
+            optionalString (isDrv && updateScript == null) ''
               <dt>Notes:</dt>
               <dd><span class="option-alert">Missing update script</span> An update script is required for automatically tracking the latest release.</dd>
             '';


ethancedwards8 (approved):
Very nice sol. LGTM. Thanks!

overview/content-types/option.nix
```diff
@@ -84,11 +84,15 @@ in
           alert-update-script =
             let
               isDrv = self.type == "package";
```
fricklerhandwerk, line 86: Side note: shouldn't the type field be types.optionType? Is that information preserved in optionsDoc? Otherwise we're matching against a magic string essentially. Not urgent or important, just a little paper cut one can prevent by being more rigorous.

overview/content-types/option.nix
```diff
@@ -84,11 +84,15 @@ in
...
-              optionName = lib.removePrefix "pkgs." self.default.text;
+              attrPath = lib.splitString "." self.default.text ++ [
```
fricklerhandwerk, line 87: Just curious: Can we in principle get the value directly?

overview/content-types/option.nix
eljamm: I agree with your point, but I'm not sure we can use this to get the type definitions:

```
nix-repl> lib.types.optionType.getSubOptions {}
{ }
```

Maybe I haven't looked hard enough, but I've looked for a proper way to do this before and couldn't find any ...

overview/content-types/option.nix
eljamm: If there is a way to transform a string path into an attribute set, then yes, else all we can do is check the attribute path. I don't think the former is possible, though, last time I checked.

PR #1357, opened 2025-07-12
overview: fix demo description access

JuneStepp:
The descriptions aren't shown in the overview yet, but the not implemented warning now works correctly.


eljamm (approved):
Thanks!

Issue #1032, opened 2025-05-16
Link high and low-level deliverables


eljamm:
When clicking on a high-level deliverable (service, program, binary, ...), users should get a detailed view on which low-level deliverable (example, test, ...) are missing with the option of [editing their code on GitHub](https://github.com/ngi-nix/ngipkgs/issues/1022).

PR #1397, opened 2025-07-15
overview: link deliverables from project list to project page

eljamm:
Clicking the deliverable label from the project list also auto-opens the option details in the project page.

Closes https://github.com/ngi-nix/ngipkgs/issues/1032

image

image


overview/content-types/deliverable.nix
```diff
@@ -0,0 +1,52 @@
...
+          # programs or services
+          type = head (splitString "." self.name);
+        in
+        ''
+          <a
+            class="deliverable-tag ${optionalString self.hasProblem "deliverable-has-problem"}"
+            title="${self.name}${optionalString self.hasProblem " has a problem"}"
+            href="/project/${self.project-name}#${self.name}"
+          >
+            ${removeSuffix "s" type}
```
erictapen, line 47: This is quite the assumption, that the singular form is always the plural with the "s" removed…
I think we should stay explicit here and keep the `type` field.

I'd also question wether we need a separate file for the tag, as this means we have to set the project name twice this way. Imo these content-types can get a lot longer than they are currently, before they are getting messy to maintain.

overview/content-types/project-list-item.nix
```diff
@@ -17,26 +17,7 @@ in
...
-                "demo"
-              ];
-            };
-            hasProblem = mkOption {
-              type = nullOr bool;
-              default = null;
-            };
-          };
-        });
+      type = with types; listOf (submodule ./deliverable.nix);
```
erictapen, line 20:
```suggestion
      type = with types; listOf (submodule ./project-list-item-tag.nix);
```

The naming of the modules should be catered towards presentation, not our deliverable data model. Imagine a maintainer that looks at a certain UI element (the tags in the items of the project list) and tries to find the correct content-type module for that.

erictapen (commented):
Love that this works now!

overview/content-types/deliverable.nix
eljamm: > I'd also question wether we need a separate file for the tag, as this means we have to set the project name twice this way. Imo these content-types can get a lot longer than they are currently, before they'd get messy to maintain.

You're right that over-using content-types might get out of hand, but it improves the overall readability of the code IMO, and thus decreases the long-term maintainability, compared to not using it. In this instance, I suspect that the deliverable tags will grow even more complex in the future (colors, group by type, details, ...), which won't be as big of a deal thanks to the content-type.

If not, we can always re-integrate this into other modules, which should be cheaper to do, and is likely what we're gonna do anyways as we continue exercising this workflow.

overview/default.nix
```diff
@@ -301,9 +300,7 @@ let
             setTimeout(() => button.textContent = "Copy", 2000);
           }
 
-          ${
-            "" # TODO: this should be the exact same code for copying file content
-          }
+          // TODO: this should be the exact same code for copying file content
```
fricklerhandwerk, line 303: Why? This comment should not show up in the file output, and changing it will require a rebuild for no good reason.

PR #1398, opened 2025-07-16
shell: add ngipkgs-test to test nixpkgs pr's against ngipkgs

ethancedwards8:
Based off a discussion in Matrix.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -304,6 +304,20 @@ rec {
       (pkgs.devmode.override {
         buildArgs = "-A overview --show-trace -v";
       })
+
+      (pkgs.writeShellApplication {
+        # TODO: have the program list available tests
+        name = "ngipkgs-test";
+        text = ''
+          export pr="$1"
+          export proj="$2"
+          export test="$3"
+          # remove the first args and feed the rest (for example flags)
+          export args="''${*:4}"
+
+          nix build --override-input nixpkgs "github:NixOS/nixpkgs?ref=pull/$pr/merge" .#checks.x86_64-linux.projects/"$proj"/nixos/tests/"$test" "$args"
+        '';
+      })
     ];
   };
 


default.nix
```diff
@@ -305,6 +305,15 @@ rec {
         buildArgs = "-A overview --show-trace -v";
       })
     ];
+    shellHook = ''
```
fricklerhandwerk, line 308: why not a `writeShellApplication` in `packages`?

default.nix
```diff
@@ -305,6 +305,15 @@ rec {
...
+      ngipkgs-test() {
+        local pr="$1"
+        local proj="$2"
+        local test="$3"
```
fricklerhandwerk, line 312: would be cool if it could tell you which tests are available if you forget that one (we can leave this as an in-code comment, not needed now)

default.nix
```diff
@@ -305,6 +305,15 @@ rec {
...
+
+        nix build --override-input nixpkgs "github:NixOS/nixpkgs?ref=pull/$pr/merge" .#checks.x86_64-linux.projects/$proj/nixos/tests/$test -vL
```
fricklerhandwerk, line 314: maybe?

```suggestion
        nix build --override-input nixpkgs "github:NixOS/nixpkgs?ref=pull/$pr/merge" .#checks.x86_64-linux.projects/$proj/nixos/tests/$test "$@"
```

default.nix
ethancedwards8: Yeah I agree. 

One quirk I noticed was that when using `nix-build -A` you can just do `<project>.tests` and it'll build them all. If you do `nix build .checks.....<project>/tests` it errors. You have to specifcy the test.

default.nix
ethancedwards8: Yeah, makes sense.

fricklerhandwerk (approved):

PR #1403, opened 2025-07-16
WIP: extract demo shell specifics from nixos modules

fricklerhandwerk:
this puts things where they belong instead of polluting nixos-specifics

@eljamm


eljamm:
This works, but still needs cleanup, which I'll do later.

eljamm:
Cleaned up and dropped some unrelated refactors, so this should be more reviewable now.

default.nix
```diff
@@ -330,18 +328,32 @@ rec {
...
   };
 
   inherit (demo)
     demo-vm
     demo-shell
     ;
+
+  # bash $(nix-build -A demos.projectName)
+  demos = lib.mapAttrs (
+    project: project-demo: demo."demo-${project-demo.type}" project-demo.module
```
fricklerhandwerk, line 357: Huh, can't we just tuck the function application into the demo deliverable type itself? Such that the field we'd expose is essentially of type `package`?

projects/types.nix
```diff
@@ -253,25 +253,49 @@ let
...
+      { name, config, ... }:
+      {
+        options = {
+          inherit (types'.example.getSubOptions { })
+            module
+            tests
+            description
+            links
+            ;
+          # TODO: why doesn't this work with an enum?
```
fricklerhandwerk, line 266: That's very weird. What's the error? (likely something to do with lazyness??)

default.nix
eljamm: Technically, we can and I've explored this a bit, but I don't know if what I did is the sound approach:

<details><summary>Refactor demo activation</summary>
<p>

```diff
commit ade2578300ab8e6cf56b434b0ca7c9d2ac925dff
Author: eljamm <fedi.jamoussi@protonmail.ch>
Date:   Fri Jul 18 09:09:51 2025 +0200

    Refactor demo activation

diff --git a/default.nix b/default.nix
index 562cbd67..17bedb20 100644
--- a/default.nix
+++ b/default.nix
@@ -354,6 +354,6 @@ rec {
 
   # bash $(nix-build -A demos.projectName)
   demos = lib.mapAttrs (
-    project: project-demo: demo."demo-${project-demo.type}" project-demo.module
+    project: project-demo: project-demo.activate (demo.eval project-demo.module)
   ) all-the-demos;
 }
diff --git a/overview/demo/default.nix b/overview/demo/default.nix
index 11d70b38..8a557f2f 100644
--- a/overview/demo/default.nix
+++ b/overview/demo/default.nix
@@ -23,13 +23,14 @@ let
         inherit sources;
       };
     };
-in
-{
-  demo-vm =
-    module:
-    pkgs.writeShellScript "demo-vm" ''
-      exec ${(eval module).config.system.build.vm}/bin/run-nixos-vm "$@"
-    '';
 
+  demo-vm = module: (eval module).config.demo-vm.activate;
   demo-shell = module: (eval module).config.shells.bash.activate;
+in
+{
+  inherit
+    eval
+    demo-vm
+    demo-shell
+    ;
 }
diff --git a/overview/demo/vm/default.nix b/overview/demo/vm/default.nix
index 06e9c261..e3f39b7c 100644
--- a/overview/demo/vm/default.nix
+++ b/overview/demo/vm/default.nix
@@ -1,3 +1,15 @@
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  inherit (lib)
+    types
+    mkOption
+    ;
+in
 {
   imports = [
     ./services.nix
@@ -5,12 +17,26 @@
     ./virtualisation.nix
   ];
 
-  services.getty.helpLine = ''
+  options = {
+    demo-vm.activate = mkOption {
+      type = with types; nullOr package;
+      default = null;
+      apply =
+        self:
+        pkgs.writeShellScript "demo-vm" ''
+          exec ${config.system.build.vm}/bin/run-nixos-vm "$@"
+        '';
+    };
+  };
+
+  config = {
+    services.getty.helpLine = ''
 
-    Welcome to NGIpkgs!
+      Welcome to NGIpkgs!
 
-    - To exit the demo VM, run: `sudo poweroff`
-  '';
+      - To exit the demo VM, run: `sudo poweroff`
+    '';
 
-  system.stateVersion = "25.05";
+    system.stateVersion = "25.05";
+  };
 }
diff --git a/projects/types.nix b/projects/types.nix
index 60f12d47..68653f64 100644
--- a/projects/types.nix
+++ b/projects/types.nix
@@ -254,7 +254,7 @@ let
       );
 
     demo = types.submodule (
-      { name, config, ... }:
+      { name, ... }:
       {
         options = {
           inherit (types'.example.getSubOptions { })
@@ -263,10 +263,11 @@ let
             description
             links
             ;
-          # TODO: why doesn't this work with an enum?
-          type = mkOption {
-            type = types.str;
-            default = name;
+          activate = mkOption {
+            type = with types; functionTo package;
+            default =
+              system:
+              if name == "shell" then system.config.shells.bash.activate else system.config.demo-vm.activate;
           };
           demo-stuff = mkOption {
             description = ''
@@ -275,10 +276,10 @@ let
             type = types.deferredModuleWith {
               staticModules =
                 with lib;
-                optionals (config.type == "vm") [
+                optionals (name == "vm") [
                   ../overview/demo/vm
                 ]
-                ++ optionals (config.type == "shell") [
+                ++ optionals (name == "shell") [
                   ../overview/demo/shell.nix
                 ];
             };
```

</p>
</details>

default.nix
```diff
@@ -330,18 +328,32 @@ rec {
...
   demo = import ./overview/demo {
     inherit
       lib
       pkgs
       sources
-      extendedNixosModules
       system
       ;
+    all-the-demo-modules = lib.flatten (
+      lib.mapAttrsToList (name: value: value.demo-stuff.imports) all-the-demos
```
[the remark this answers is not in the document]
eljamm: > Does accessing `imports` always work? Is that an attribute of the `deferredModule` config value?

Yes, I think so.

projects/types.nix
fricklerhandwerk: `null` would produce a type error though, no?

default.nix
fricklerhandwerk: Let's look at it in sync later

projects/types.nix
eljamm: Yeah, not sure. I'm just speculating, because the type can't be anything else since we're using an attrTag.

eljamm:
Renamed the new option to `module-demo`, which IMO is a bit clearer on what its purpose is. Let's merge this for now for the sake of separating the demo config from production. We can always address any shortcomings in a followup.

PR #1406, opened 2025-07-17
projects/owi: add demo-shell to module.nix

ethancedwards8:
Missed this in the original PR.

### DIFF
--- a/projects/owi/programs/owi/module.nix
+++ b/projects/owi/programs/owi/module.nix
@@ -17,5 +17,9 @@ in
     environment.systemPackages = with pkgs; [
       cfg.package
     ];
+
+    demo-shell.owi.programs = {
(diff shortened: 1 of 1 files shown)


projects/owi/programs/owi/module.nix
```diff
@@ -17,5 +17,9 @@ in
     environment.systemPackages = with pkgs; [
       cfg.package
     ];
+
+    demo-shell.owi.programs = {
```
eljamm, line 21: We've reworked the demo in https://github.com/ngi-nix/ngipkgs/pull/1403 such that this is set in a separate config, now (also see https://github.com/ngi-nix/ngipkgs/pull/1407).

eljamm (approved):

PR #1416, opened 2025-07-17
projects/stalwart: init

ethancedwards8:


fricklerhandwerk:
We somehow need to make sure our examples aren't diverging from the test. @ethancedwards8 would it be an option for you to extract the config-under-test upstream into a file, so we can access it directly?

fricklerhandwerk:
Make an upstream change. So far we've been copying stuff here, and this is also valid, but increases the amount of code under our exclusive maintenance instead of reducing it.

ethancedwards8:
Definitely. Makes sense. I'll do it up!

ethancedwards8:
Done here: https://github.com/NixOS/nixpkgs/pull/427050

ethancedwards8:
Config is split. Will have to wait for the next flake update to merge.

ethancedwards8:
Failures are due to rate limiting...

PR #1419, opened 2025-07-18
mox: add demo vm

### LINKED ISSUE #1319: Implement a demo example for Mox
### Instructions

Use metadata about the application from its tracking issue:

- ngi-nix/projects#23

and implement a demo.

themadbit:
heavily WIP

closes ngi-nix/projects#1791


projects/Mox/services/mox/module.nix
```diff
@@ -64,6 +64,7 @@ in
     ];
 
     networking.firewall = lib.mkIf cfg.openFirewall {
+      enable = true;
```
Prince213, line 67: This is bad practice.

https://github.com/NixOS/nixpkgs/blob/master/nixos/README.md?plain=1#L96-L97
> Ensure that the module respect other modules functionality.
> - For example, enabling a module should not open firewall ports by default.

overview/demo/vm/services.nix
```diff
@@ -11,5 +12,5 @@
     };
   };
 
-  networking.firewall.enable = false;
+  networking.firewall.enable = lib.mkDefault false;
```
Prince213, line 15: And this shouldn't be necessary.

Prince213 (changes requested):

fricklerhandwerk:
Blocked on:
- There's no tooling to generate the config file, which is in a custom ini-like format
- We can't disable TLS without accessing the config file
- Alternatively running a CA (pebble) would require two machines for the demo (https://github.com/ngi-nix/ngipkgs/issues/1434)
- Also for some reason the service doesn't start up in the demo VM (but does in the test VM, to be investigated)

@imincik for when you're back, we should assign this to someone else so @themadbit can take on something different, otherwise this will be just frustrating.

PR #1424, opened 2025-07-18
projects/Agorakit: add summary

ethancedwards8:


projects/Agorakit/default.nix
```diff
@@ -6,6 +6,7 @@
 }@args:
 {
   metadata = {
+    summary = "Agorakit is a web-based, open source organization tool for collectives";
```
themadbit, line 9:
```suggestion
    summary = "A web-based, open source organization tool for collectives";
```
Small nit. I've noticed that pretty much the standard. Just saying what it is
Also, it's better to track these small changes too with their issue template. i.e implementing metadata and adding it to the parent project issue. Helps in tracking progress and reporting.

ethancedwards8: Sure thing.

Good point with the issues. I'll rectify that.

fricklerhandwerk (approved):

themadbit (approved):

PR #1428, opened 2025-07-18
projects/OpenWebCalendar: add summary, additional subgrant

ethancedwards8:


projects/OpenWebCalendar/default.nix
```diff
@@ -5,7 +5,13 @@
   ...
 }@args:
 {
-  metadata.subgrants = [ "OpenWebCalendar" ];
+  metadata = {
+    summary = "The Open Web Calendar stack is an open-source set of Python libraries and programs which read and write calendars based on the iCalendar standard";
```
fricklerhandwerk, line 9:
```suggestion
    summary = "Python programs and libraries which read and write calendars based on the iCalendar standard";
```

fricklerhandwerk (approved):

PR #1432, opened 2025-07-21
ethersync: remove in favor of nixpkgs ethersync

ethancedwards8:
As of https://github.com/NixOS/nixpkgs/pull/425651, ethersync is in Nixpkgs.

We will have to wait for the channel to update and include it, but we should be safe to drop it from ngipkgs eventually.


ethancedwards8:
This is ready for merging I think

PR #1444, opened 2025-07-21
verso: init at 0-unstable-2025-06-15

ethancedwards8:
I've been trying to get this in Nixpkgs for several weeks now with no movement. Get it in Ngipkgs to get the ball rolling finally.

PR #1447, opened 2025-07-21
makemake: Set Nix max-silent-time to one hour

Erethon:
This will hopefully allow some big project compilations to finish before being killed by Nix. This is because of servo in #1444


fricklerhandwerk (approved):

fricklerhandwerk:
Side note, the failing builds fail because nix is found. Huh?

Issue #1401, opened 2025-07-16
US: Binary deliverables


fricklerhandwerk:
### Description

As a NGIpkgs user, I want to know how to obtain binaries exposed by projects so that I can use them.

### Acceptance criteria

- Given a project that exposes at least one binary, when I view the project details, there is an example for how to obtain one of the binaries
- Given a project that exposes at least one binary, when I view the project details, all of the available binaries are listed

ethancedwards8:
Do we want to add something copy-and-pastable like?

```
nix-store -r /nix/store/zympn1h6a6awh5x2fifbsm1q4yp4c9sn-goa-3.21.5 \
  --option binary-caches 'https://cache.nixos.org/ https://nixpkgs-update-cache.nix-community.org/' \
  --option trusted-public-keys '
  nixpkgs-update-cache.nix-community.org-1:U8d6wiQecHUPJFSqHN9GSSmNkmdiFW7GW7WNAnHW0SM=
  cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
  '
```
Where you literally just download the binary?

Or maybe using `nix-env` or `nix profile install`? Does it need to be declarative? How nix'y do we want to make it?

fricklerhandwerk:
There should definitely be a way to build it from source. The slickest way would be `nix build <ngipkgs-flake>#something` but that's experimental and we don't want people to jump through hoops. Therefore something like `nix-build https://github.com/ngi-nix/ngipkgs/tarball/main -A something` with an optional `export NIX_CONFIG ...` as in the current instructions is probably the way to go.

fricklerhandwerk:
To be clear, we're talking about raw binaries such as firmware images (not executables from programs) for projects such as Heads or LibreSoC. The attribute should definitely live in the project, because this is where we have full control.

ethancedwards8:
Oh I see. My apologies. I misunderstood the objective. That elucidates things for me. Thank you.

PR #1448, opened 2025-07-22
overview: add instructions to download a binary

ethancedwards8:
Still a WIP, but the basic function is there.

Closes: https://github.com/ngi-nix/ngipkgs/issues/1401

Screenshot 2025-07-21 at 20 57 46


themadbit:
>... how I can access the package of a project?

You might want to check https://github.com/ngi-nix/ngipkgs/blob/4040be4360392eb01acca65038f6088c59e7ccfc/overview/content-types/option.nix#L86 or the entirety of the expression. Accesses the package to determine if it has an updateScript

PR #1470, opened 2025-07-29
heads: Rework board enablement, add more boards

OPNA2608:
- Don't allow firmware blobs by default. Assume that when they're enabled, they taint the resulting package license.
- Explicitly comment that the allowedBoards list only includes boards for which we are sure that we can redistribute the results for.
- Allow multiple `qemu-coreboot*` boards that have managed to build fine on CI so far.
- Make a list of known-broken boards: Depend on coreboot 4.11, which wants a known-insecure expat version.

### DIFF
--- a/pkgs/by-name/heads/package.nix
+++ b/pkgs/by-name/heads/package.nix
@@ -13,6 +13,7 @@
   _experimental-update-script-combinators,
 
   autoconf,
+  automake,
   bashNonInteractive,
   bc,
   binwalk,
@@ -33,6 +34,7 @@
   imagemagick,
   innoextract,
   jq,
+  libtool,
   ncurses,
   nix,
   nixfmt-rfc-style,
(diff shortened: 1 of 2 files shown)


fricklerhandwerk (dismissed):

OPNA2608:
I *think* we might have all the Qemu board targets building now (I only tested one of the TPM2 ones, but I assume the other TPM2 ones will work too).

There are some more targets with (likely?) no need for blobs that we could look into enabling and would cover some actual hardware, but they currently don't build:
- Some of the Librem ones ask for the coreboot version that we refuse to offer due to it trying to fetch & build-from-source a known-vulnerable package
- The rest of the Librem targets fail due to missing libcrypto pkg-config files. I'm assuming they would rather build against their own openssl package than our `openssl` package?
- The Talos II target, besides being marked untested and thus potentially not being of interest to us, currently fails due to it trying to pull in a tarball that we explicitly don't provide due to there being no 100% known-good copy of it: https://github.com/ngi-nix/ngipkgs/pull/1293

fricklerhandwerk (approved):

PR #1471, opened 2025-07-30
overview: add homepage using breadcrumbs

themadbit:


ethancedwards8 (dismissed):
diff lgtm. Good idea. This had been annoying me.

JuneStepp (changes requested):
Breadcrumbs have certain accessibility considerations. See https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb.

JuneStepp (commented):
I have a few suggestions:

- Don't make the final breadcrumb component a link. Instead of trying to CSS and ARIA it back to text, it can just be text.
- Fix typo. `<li">` to `<li>` .
- Add empty alt text to separator, so that it isn't read by screen readers: `content: "/" / ""`
- Remove `li` css. It's not needed.
- Don't set color on separator. It makes little difference and worsens theming.
- Don't change last component font weight. I don't see why it should be emphasized but if it is supposed to be emphasized, that should be done semantically with HTML.
- Sort the render let-in by where stuff is used. Just a nice touch.

Here's what the diff looks like with my suggestions:
```diff
diff --git a/overview/default.nix b/overview/default.nix
index 46cd2cb..3f1b427 100644
--- a/overview/default.nix
+++ b/overview/default.nix
@@ -102,17 +102,26 @@ let
     projects.one =
       name: project:
       let
-        examples = eval {
-          imports = [ ./content-types/example-list.nix ];
-          examples = map (value: {
-            inherit (value)
-              description
-              module
-              name
-              tests
-              ;
-          }) (pick.examples project);
-        };
+        breadcrumbs = ''
+          <nav aria-label="Breadcrumb" class="breadcrumb">
+            <ol>
+              <li><a href="/">Projects</a></li>
+              <li>${name}</li>
+            </ol>
+          </nav>
+        '';
+
+        metadata-summary = optionalString (project.metadata != null && project.metadata.summary != null) ''
+          <p>
+            ${project.metadata.summary}
+          </p>
+        '';
+
+        demo-instructions = optionalString (project.nixos.demo != null) (
+          lib.concatMapAttrsStringSep "\n" (
+            type: demo: toString (render.serviceDemo.one type demo)
+          ) project.nixos.demo
+        );
 
         # TODO: clean up
         optionsRender =
@@ -135,17 +144,22 @@ let
               "services"
             ];
 
-        metadata-summary = optionalString (project.metadata != null && project.metadata.summary != null) ''
-          <p>
-            ${project.metadata.summary}
-          </p>
-        '';
+        examples = eval {
+          imports = [ ./content-types/example-list.nix ];
+          examples = map (value: {
+            inherit (value)
+              description
+              module
+              name
+              tests
+              ;
+          }) (pick.examples project);
+        };
 
         metadata-subgrants = eval {
           imports = [ ./content-types/metadata-subgrants.nix ];
           subgrants = project.metadata.subgrants or null;
         };
-
         metadata-links = eval {
           imports = [ ./content-types/metadata-links.nix ];
           links = project.metadata.links or null;
@@ -153,13 +167,10 @@ let
       in
       ''
         <article class="page-width">
+          ${breadcrumbs}
           ${heading 1 null name}
           ${metadata-summary}
-          ${optionalString (project.nixos.demo != null) (
-            lib.concatMapAttrsStringSep "\n" (
-              type: demo: toString (render.serviceDemo.one type demo)
-            ) project.nixos.demo
-          )}
+          ${demo-instructions}
           ${optionalString (lib.trim optionsRender != "") "${heading 2 "service" "Options"}"}
           ${optionsRender}
           ${examples}
diff --git a/overview/style.css b/overview/style.css
index 4c81799..47a018c 100644
--- a/overview/style.css
+++ b/overview/style.css
@@ -452,6 +452,18 @@ dd {
   margin-bottom: 1.5em;
 }
 
+nav.breadcrumb ol {
+  display: flex;
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
+nav.breadcrumb li:not(:last-child):after {
+  content: "/" / "";
+  margin: 0 0.5em;
+}
+
 @media (prefers-color-scheme: dark) {
   body {
     background: #3c3434;

```

overview/style.css
eljamm, line 1: Since we don't currently have any formatting convention for css, I think it would be better to not include style changes until we do so.

themadbit: For this diff in particular, what's the style change to avoid?

themadbit: I see, you mean the formatting of the css file?

overview/default.nix
```diff
@@ -102,17 +102,28 @@ let
...
-              tests
-              ;
-          }) (pick.examples project);
-        };
+        breadcrumbs = ''
+          <nav aria-label="Breadcrumb" class="breadcrumb">
+            <ol>
+              <li>
+                <a href="/">Projects</a>
+              </li>
```
JuneStepp, line 110:
```suggestion
              <li><a href="/">Projects</a></li>
```

This gets rid of the little extra whitespace after "Projects" that makes the space to the left of the slash larger than on the right.

eljamm (approved):
Very cool, thanks!

Issue #1434, opened 2025-07-21
Spike: Multi-machine demos


fricklerhandwerk:
### Objective

With Kaidan (and possibly Ethersync) we have an instance where the demo would only really shine if run from two different machines simultaneously. Find out how to do that (including a test, of course) and display it appropriately.

**Note:** The desired outcome of the spike are refined user stories.

- https://github.com/ngi-nix/ngipkgs/issues/1500
- https://github.com/ngi-nix/ngipkgs/issues/1501

PR #1474, opened 2025-07-31
projects/PeerTube: add peertube-runner service

### LINKED ISSUE #1148: Implement NixOS service module for peertube-runer
### Instructions

Given #821 for metadata about the application, implement a NixOS service module for PeerTube.

### Tasks

After or during the implementation of this module, create issues for the following deliverables using the appropriate GitHub template:

- [ ] Example configuration
- [ ] NixOS test

**Note:** Each example needs to be tested.

JuneStepp:
I did most of this on the nixpkgs side. That's also where a more comprehensive test is.

Closes #1148

### DIFF
--- a/projects/PeerTube/services/peertube-runner/examples/basic.nix
+++ b/projects/PeerTube/services/peertube-runner/examples/basic.nix
@@ -1,0 +1,14 @@
+{
+  services.peertube-runner = {
+    enable = true;
+    instancesToRegister = {
+      # All instances must be registered here. They can't be registered using the CLI.
+      # personal = {
+      #   url = "https://mypeertubeinstance.com";
+      #   # See how to generate registration tokens at https://docs.joinpeertube.org/admin/remote-runners#manage-remote-runners.
+      #   registrationTokenFile = "/run/secrets/my-peertube-instance-registration-token";
+      #   runnerName = "Main";
+      # };
+    };
+  };
+}
--- a/projects/PeerTube/default.nix
+++ b/projects/PeerTube/default.nix
@@ -52,11 +52,25 @@
       services = {
         peertube = {
           module = ./services/peertube/module.nix;
-          examples.basic = {
+          examples.basic-server = {
             module = ./services/peertube/examples/basic.nix;
             description = "Basic server configuration";
             tests.peertube-plugins.module = import ./services/peertube/tests/peertube-plugins.nix args;
             tests.peertube-plugin-livechat.module = import ./services/peertube/tests/peertube-plugin-livechat.nix args;
+          };
+        };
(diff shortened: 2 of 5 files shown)


projects/PeerTube/services/peertube-runner/examples/basic.nix
```diff
@@ -0,0 +1,13 @@
+{
+  services.peertube-runner = {
+    enable = true;
+    instancesToRegister = {
+      # All instances must be registered here. They can't be registered using the CLI.
```
fricklerhandwerk, line 5: Is there any way we could have an actual sample here that works? After skimming the upstream docs I don't have an intuition for how to do it here, or how to even set up a runner.

JuneStepp: I've added a comment and switched the linked documentation page to make things a bit clearer.

The thought process for doing it this way was something like:
- Having an example set up out of the box with a PeerTube instance would be complicated. https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/tests/web-apps/peertube.nix
- Running peertube-runner on the same system as the PeerTube isn't something anyone would actually do. (it provides no benefits)
- peertube-runner doesn't have the same audience as PeerTube. People looking to use peertube-runner are administrators that want to offload specific processing tasks from their existing PeerTube server to separate ones. These servers may or may not be even configured with NixOS.
- There's no real benefit to peertube-runner in the demo either, because it changes no features of PeerTube when on a single system.  
- An example with a pre-setup registration token wouldn't be realistic. Even when using Nix, the expected way to use `services.peertube-runner` module is to manually generate a runner token in the web UI. Of course, it would still be helpful if there was reason to demo it.

There are certainly counterpoints to those thoughts, but it was enough for me to think at least the example part should be done like this.

JuneStepp: There are also publicly available test instances for PeerTube, but they didn't have a pre-setup registration token or anything.

OPNA2608: Spike issue for multi-machine demos: https://github.com/ngi-nix/ngipkgs/issues/1434

OPNA2608 (approved):
LGTM. Let's open an issue about enhancing this demo with a multi-machine setup after the merge.

projects/PeerTube/services/peertube-runner/examples/basic.nix
OPNA2608: Issue for the peertube-runner demo idea: https://github.com/ngi-nix/ngipkgs/issues/1489

PR #1476, opened 2025-07-31
heads: Add Librem targets

OPNA2608:
Except `librem_l1um`, since that one still needs a Coreboot version we refuse to offer.

---

First builds for actual hardware platforms, and they don't need any externally-supplied blobs! Some of [these products](https://puri.sm/products/) even ship with Heads pre-installed apparently (11 & Mini mention it in their marketing blurbs).

<sup><sub>(#NotAnAd)</sub></sup>

I don't own any of these products, so the resulting ROMs are all untested. I successfully built two of them on my end (`librem_14` and `librem_l1um_v2`), I'll let CI tell me if any of the other ones have unique issues.

### DIFF
--- a/pkgs/by-name/heads/package.nix
+++ b/pkgs/by-name/heads/package.nix
@@ -39,6 +39,8 @@
   nix,
   nixfmt-rfc-style,
   nix-prefetch-scripts,
+  nss,
+  openssl,
   perl,
   pkg-config,
   python3,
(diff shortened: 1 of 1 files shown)


OPNA2608:
I noticed that some board names in the `allowedBoards` list are typo'd and silently ignored. I have a change ready to fix those typos and assert that all board names everywhere are valid, but it'll result in new builds, so I'll submit that separately tonight.

All boards added here generated a corresponding derivation, so they should be typo-free.

fricklerhandwerk (approved):

Issue #1362, opened 2025-07-14
Bug: overview does not detect update script of programs packaged in NGIpkgs


ethancedwards8:
The overview page does not detect the update scripts of packages packged in NGIpkgs rather than Nixpkgs.

For example, `inventaire-client` and `blink-qt` both indicate that they are missing update scripts despite them being present in their derivations.

I'm guessing it has something to do with now the packages are exposed in NGIpkgs.

fricklerhandwerk:
Hm, I thought we're adding our local derivations as an overlay?

PR #1481, opened 2025-08-04
overview: extend pkgs with ngipkgs overlay

eljamm:
Closes https://github.com/ngi-nix/ngipkgs/issues/1362


ethancedwards8 (dismissed):
neat. Thanks!

PR #1515, opened 2025-08-10
projects/owasp: switch to upstream tests

ethancedwards8:
I upstreamed these recently.


ethancedwards8:
This will have to wait until the flake update tonight.

PR #1529, opened 2025-08-12
Kaidan: fix deprecated option; separate demo config

eljamm:
## Changes

- Replace `services.prosody.uploadHttp' with `services.prosody.httpFileShare`
- Split demo config into a separate file


projects/Kaidan/default.nix
```diff
@@ -64,17 +65,17 @@
       NOTE: You need an XMPP user account to use Kaidan.
       If you have an account, use that to login and you should be able to send/receive a message.
 
-      If you don't have one, two demo user accounts have been created (`alice` and `john`).
+      If you don't have one, two demo-user accounts will be available: `alice` and `john`
```
themadbit, line 68:
```suggestion
      If you don't have one, two demo-user accounts, `alice` and `john`, are available.
```

themadbit (dismissed):

PR #1530, opened 2025-08-12
docs: document project types

eljamm:


themadbit (dismissed):

projects/types.nix
```diff
@@ -183,6 +289,30 @@ let
...
+            description = "Basic configuration example for foobar";
+            tests.basic.module = import ./path/to/tests/basic.nix args;
+          };
+        };
+        ```
+
+        :::
+
+        :::{.note}
+        Each service must include at least one example, so users get an idea of what to do with it.
```
imincik, line 311:
```suggestion
        Each service must include at least one example (each example must include test), so users get an idea of what to do with it.
```

projects/types.nix
```diff
@@ -220,6 +384,40 @@ let
...
+    /**
+      Configuration of an application module that illustrates how to use it.
+
+      :::{.example}
+
+      ```nix
+      nixos.modules.services.some-service.examples = {
+        "Basic mail server setup with default ports" = {
+          module = ./services/some-service/examples/basic.nix;
+          description = "Send email via SMTP to port 587 to check that it works";
```
imincik, line 396: Add `test` :)

PR #1400, opened 2025-07-16
Usage instructions with screenshots in project overview page

Prince213:
Closes https://github.com/ngi-nix/ngipkgs/issues/1331.

PR #1533, opened 2025-08-13
overview: add usage-instructions

Prince213:
See also https://github.com/ngi-nix/ngipkgs/pull/1400.


eljamm (commented):
Nice job! But I think we can improve it further. This is how it looks like with the suggestions, below:

image

eljamm (approved):

PR #1537, opened 2025-08-15
projects/Kaidan: disable demo module by default

JuneStepp:
It was somewhat confusing when random services showed up in my demo.


eljamm (approved):
Oops, thanks!

PR #1539, opened 2025-08-16
eris-go: init package with modular-service

ghost:
There is a new release of eris-go that fixes bugs and adds a new configuration mechanism.

The eris-go package in Nixpkgs is already a few versions behind and I cannot do anything to update it.

This PR packages the latest version with a modular service so that eris-go can be purged from Nixpkgs.

The point of technical contention here is that there are no [modular services](https://github.com/NixOS/nixpkgs/pull/372170) in Ngipkgs yet and it makes the example more confusing to read.

I'm the upstream maintainer of eris-go and this modular service is what I'm using on my non-NixOS machines, so I would prefer to have it published with the package.

### DIFF
--- a/pkgs/by-name/eris-go/package.nix
+++ b/pkgs/by-name/eris-go/package.nix
@@ -1,0 +1,58 @@
+{
+  lib,
+  buildGoModule,
+  fetchFromGitea,
+  installShellFiles,
+  execline,
+  formats,
+}:
+
+buildGoModule (finalAttrs: {
+  pname = "eris-go";
+  version = "20250820";
+  outputs = [
+    "out"
+    "man"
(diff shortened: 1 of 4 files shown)


eljamm (commented):
Hey, @ehmry. Apologies for taking this long to reply, and thanks for your contribution. We've considered using modular services in NGIpkgs before and it's really interesting to see it in action. I think it would be nice to explore using them, but it would be better to do it in a way that keeps UX consistent with what we already have.

PS: I'm not sure why you're unable to update `eris-go`, but we eventually want to migrate everything from NGIpkgs to Nixpkgs, which would still be possible with modular services since they're merged, upstream.

PR #1589, opened 2025-08-29
projects/ReOxide: implement program module

themadbit:


projects/ReOxide/services/reoxided/tests/basic.nix
```diff
@@ -0,0 +1,40 @@
...
+      simple_plugin = "${pkgs.reoxide-plugin-simple}";
+    in
+    ''
+      start_all()
+
+      machine.wait_for_unit("reoxided.service")
+
+      # copy simple plugin to plugins directory
+      machine.succeed("mkdir -p ~/.local/share/reoxide/plugins")
+      machine.succeed("cp ${simple_plugin}/lib/libsimple.so ~/.local/share/reoxide/plugins/")
```
eljamm, line 35:
```suggestion
      machine.succeed("cp ${simple_plugin} ~/.local/share/reoxide/plugins/")
```

projects/ReOxide/services/reoxided/tests/basic.nix
```diff
@@ -0,0 +1,40 @@
...
+          sources.examples.ReOxide."Enable reoxided"
+          (sources.inputs.nixpkgs + "/nixos/tests/common/x11.nix")
+        ];
+      };
+  };
+
+  testScript =
+    { nodes, ... }:
+    let
+      simple_plugin = "${pkgs.reoxide-plugin-simple}";
```
eljamm, line 26: I feel like this is better, since it points exactly to the plugin, as opposed to the plugin's directory:

```suggestion
      simple_plugin = "${pkgs.reoxide-plugin-simple}/lib/libsimple.so";
```

projects/ReOxide/services/reoxided/tests/basic.nix
```diff
@@ -0,0 +1,40 @@
...
+      start_all()
+
+      machine.wait_for_unit("reoxided.service")
+
+      # copy simple plugin to plugins directory
+      machine.succeed("mkdir -p ~/.local/share/reoxide/plugins")
+      machine.succeed("cp ${simple_plugin}/lib/libsimple.so ~/.local/share/reoxide/plugins/")
+
+      machine.succeed("systemctl restart reoxided.service")
+      machine.succeed("journalctl -u reoxided.service | grep 'libsimple.so'")
```
eljamm, line 38:
```suggestion
      machine.systemctl("restart reoxided.service")
      machine.wait_for_console_text("libsimple.so")
```

projects/ReOxide/programs/reoxide/module.nix
```diff
@@ -0,0 +1,22 @@
...
+{
+  options.programs.reoxide = {
+    enable = lib.mkEnableOption "enable reoxide";
+    package = lib.mkPackageOption pkgs "reoxide" { };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = with pkgs; [
+      cfg.package
+      reoxide-plugin-simple
```
eljamm, line 19: Installing the plugin to `systemPackages` is probably not that useful, since users won't be able to consume it.

Do you think it would be possible to wrap reoxide with the plugins it needs?

projects/ReOxide/services/reoxided/tests/basic.nix
```diff
@@ -0,0 +1,40 @@
...
+  nodes = {
+    machine =
+      { ... }:
+      {
+        imports = [
+          sources.modules.ngipkgs
+          sources.modules.programs.reoxide
+          sources.modules.services.reoxided
+          sources.examples.ReOxide."Enable reoxided"
+          (sources.inputs.nixpkgs + "/nixos/tests/common/x11.nix")
```
eljamm, line 18: Why do we need a graphical session if the test only checks the console output?

```suggestion
```

projects/ReOxide/services/reoxided/module.nix
```diff
@@ -0,0 +1,73 @@
...
+  inherit (lib)
+    mkEnableOption
+    mkPackageOption
+    mkOption
+    mkIf
+    types
+    ;
+
+  cfg = config.services.reoxided;
+  toToml = pkgs.formats.toml { };
```
eljamm, line 17: Just a nit, but generally you'd see this called `settingsFormat` in Nixpkgs.


```suggestion
  settingsFormat = pkgs.formats.toml { };
```

projects/ReOxide/services/reoxided/module.nix
```diff
@@ -0,0 +1,73 @@
...
+  configFile = toToml.generate "reoxide.toml" {
+    ghidra-install = cfg.ghidraInstall;
+  };
+
```
eljamm, line 21: Extra empty line

```suggestion
```

projects/ReOxide/services/reoxided/module.nix
```diff
@@ -0,0 +1,73 @@
...
+in
+{
+  options.services.reoxided = {
+    enable = mkEnableOption "enable reoxided";
+    package = mkPackageOption pkgs "reoxide" { };
+
+    ghidraInstall = mkOption {
+      type = types.listOf (
```
eljamm, line 29: Why do we need a list of ghidra installs? what happens when we add multiple ones?

projects/ReOxide/services/reoxided/module.nix
```diff
@@ -0,0 +1,73 @@
...
+    ];
+
+    environment.etc."reoxide.toml".source = configFile;
+
+    systemd.services.reoxided = {
+      description = "ReOxide daemon";
+      wantedBy = [ "multi-user.target" ];
+      after = [ "network.target" ];
+      serviceConfig = {
+        ExecStart = "${cfg.package}/bin/reoxided -c /etc/reoxide.toml";
```
eljamm, line 68: You can do this, directly:


```suggestion
        ExecStart = "${cfg.package}/bin/reoxided -c ${configFile}";
```

projects/ReOxide/services/reoxided/module.nix
```diff
@@ -0,0 +1,73 @@
...
+        Restart = "always";
```
eljamm, line 69: Would be nice to add some time between restart attempts:


```suggestion
        Restart = "always";
        RestartSec = 5;
```

projects/ReOxide/services/reoxided/module.nix
```diff
@@ -0,0 +1,73 @@
...
+      };
```
eljamm, line 70: It's good to cap the max number of attempts before the service fails, else it will keep on restarting indefinitely.

Here, we say that the service should fail if it's restarted 5 times within 100 seconds (see [systemd docs](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#StartLimitIntervalSec=interval)).

```suggestion
      };
      unitConfig = {
        StartLimitBurst = 5;
        StartLimitInterval = 100;
      };
```

projects/ReOxide/services/reoxided/module.nix
```diff
@@ -0,0 +1,73 @@
...
+    mkOption
+    mkIf
+    types
+    ;
+
+  cfg = config.services.reoxided;
+  toToml = pkgs.formats.toml { };
+  configFile = toToml.generate "reoxide.toml" {
+    ghidra-install = cfg.ghidraInstall;
+  };
```
eljamm, line 20: We can put this in a `services.reoxided.settings` option and just pass that, here. No need for repetition.


```suggestion
  configFile = toToml.generate "reoxide.toml" cfg.settings;
```

projects/ReOxide/programs/reoxide/module.nix
eljamm: > I think, however, that we can have a module option in the service, to specify plugins that will be loaded.

That would probably be the best course of action.

> This would still need packaging of the plugins, though.

Yeah, that's fine. We can make a package set that contains all the plugins (for example, see the [bigbluebutton scope](https://github.com/ngi-nix/ngipkgs/blob/45f16613865ba9cc9e4578f518f1a33051f54069/pkgs/by-name/bigbluebutton/sbt-derivation.nix)).

projects/ReOxide/services/reoxided/module.nix
themadbit: Reoxide can symlink multiple Ghidra instances. So, for example, you can have various Ghidra installs on your machine, and you want to use the reoxide decompiler in multiple instances. The root-dir specifies where Ghidra is installed, and that's where it will find and symlink the decompiler.

themadbit: Reoxide can symlink multiple Ghidra instances. So, for example, you can have various Ghidra installs on your machine, and you want to use the reoxide decompiler in multiple instances. The root-dir specifies where Ghidra is installed, and that's where it will find and symlink the decompiler.

projects/ReOxide/services/reoxided/examples/basic.nix
```diff
@@ -0,0 +1,12 @@
+{ lib, ... }:
+{
+  programs.reoxide.enable = true;
+  services.reoxided = {
+    enable = true;
+    ghidraInstall = [
+      {
+        enabled = lib.mkForce true;
+      }
+    ];
```
eljamm, line 10:
```suggestion
    settings.ghidra-install = [
      # default instance
      {
        enabled = true;
      }
      # NOTE: you can supply additional ghidra instances
      # {
      #   enabled = true;
      #   root-dir = "/path/to/other/ghidra/root/dir";
      # }
    ];
```

projects/ReOxide/services/reoxided/tests/basic.nix
eljamm: This is still not removed? but the comment was marked as resolved.

projects/ReOxide/services/reoxided/module.nix
themadbit: yes. See https://reoxide.eu/guide/getting-started. I still think that we should not rule out the possibility of having multiple instances. For example, one can use the shipped reoxide-ghidra, but also install another one using the ghidra installer.

projects/ReOxide/services/reoxided/module.nix
eljamm: I think you've mistakenly removed these changes.

eljamm (approved):
Very nice. Well done, Mark.

PR #1620, opened 2025-09-02
Projects/Wax

themadbit:
Still WIP.

The test needs some more work.
- The wax-server fails to start due to an issue with graphql
- ~~The wax-client currently breaks as it drops into a different shell, maybe because of `{stdenv.shell}`. I will investigate~~ (fixed https://github.com/ngi-nix/ngipkgs/pull/1620/commits/29b01c8d092ad1af1c1ebdc0ab7a333e6ffd9dd2)
- Uses minio for file/object storage over S3 buckets. Seems to be working but will be confirmed once other bugs are fixed.
- Needs the wax-server package (#1544) merged


wamirez:
Related to https://github.com/ngi-nix/projects/issues/111.

PR #1625, opened 2025-09-02
projects/hockeypuck: init

ethancedwards8:


projects/hockeypuck/default.nix
```diff
@@ -0,0 +1,32 @@
...
+  lib,
+  pkgs,
+  sources,
+  ...
+}@args:
+
+{
+  metadata = {
+    summary = "Next generation OpenPGP keyserver";
+    subgrants = [
```
eljamm, line 11:
```suggestion
    subgrants.Core = [
```

projects/hockeypuck/default.nix
```diff
@@ -0,0 +1,32 @@
...
+      name = "hockeypuck";
+      module = lib.moduleLocFromOptionString "services.hockeypuck";
+      examples."Enable hockeypuck" = {
+        module = ./services/hockeypuck/examples/basic.nix;
+        tests.basic.module = pkgs.nixosTests.hockeypuck;
+      };
+    };
+  };
+
+  nixos.demo.vm = {
```
eljamm, line 27: Usage instructions? At least how to access the service.

projects/hockeypuck/services/hockeypuck/examples/basic.nix
```diff
@@ -0,0 +1,16 @@
+{ ... }:
+
+{
+  services.hockeypuck.enable = true;
```
eljamm, line 4: The service port needs to be open for it to be forwarded to the host. You can either do that directly here or [extend the NixOS module](https://github.com/ngi-nix/ngipkgs/blob/1de6b05dfd632cda7f39436c69cfb31b59d2a04f/projects/types.nix#L119-L140) and add an `openFirewall` option, then enable it.

PR #1631, opened 2025-09-15
ci: improve security

eljamm:
- https://docs.zizmor.sh/audits/#unpinned-uses
- https://docs.zizmor.sh/audits/#excessive-permissions
- https://docs.zizmor.sh/audits/#artipacked

### DIFF
--- a/.github/workflows/makemake.yaml
+++ b/.github/workflows/makemake.yaml
@@ -1,5 +1,7 @@
 name: makemake
 
+permissions: {}
+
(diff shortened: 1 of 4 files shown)


Erethon (approved):
LGTM, thank you.

As a not, there are updated versions of the cachix/cachix-action and DeterminateSystems/update-flake-lock, but I don't think we need to update them in this PR (or yet if they're not fixing something we care about).

PR #1632, opened 2025-09-16
refactor: project composition

eljamm:
- move custom `lib` to a separate file
- refactor and simplify project composition
  - project eval time should be less now that we're not evaluating the entirety of Nixpkgs


imincik (approved):
Discussed during review meeting.

PR #1647, opened 2025-09-19
overview: display binary files 

eljamm:
Closes https://github.com/ngi-nix/ngipkgs/issues/1401

image

### DIFF
--- a/projects/Heads/default.nix
+++ b/projects/Heads/default.nix
@@ -43,7 +43,8 @@
       };
     };
   };
-  binary = lib.attrsets.mapAttrs' (
-    board: pkg: lib.attrsets.nameValuePair "${board}.rom" { data = "${pkg}/${pkg.passthru.romName}"; }
-  ) (lib.attrsets.filterAttrs (_: lib.attrsets.isDerivation) pkgs.heads);
+  binary = lib.attrsets.mapAttrs (board: pkg: {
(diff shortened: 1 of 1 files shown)


overview/content-types/binary-list.nix
```diff
@@ -0,0 +1,95 @@
...
+            imports = [ ./code-snippet.nix ];
+            snippet-text = ''
+              # test.nix
+              {
+                ngipkgs ? import (fetchTarball "https://github.com/ngi-nix/ngipkgs/tarball/main") { },
+                # Extend Nixpkgs package set with NGIpkgs
+                pkgs ? ngipkgs.pkgs.extend ngipkgs.overlays.default,
+              }:
+              {
+                example-binary = pkgs.${first-available-binary};
```
imincik:
```suggestion
                ${first-available-binary} = pkgs.${first-available-binary};
```

imincik: Actually, why we don't list all available binaries instead of example ?

imincik (changes requested):

overview/content-types/binary-list.nix
eljamm: We are listing them under `#available-binaries`. Here we're only using one for illustrative purposes.

PR #1661, opened 2025-10-02
overview: use tabs when shell-instructions is a list i.e. platform tabs

### LINKED ISSUE #1084: US: Show one platform for running a demo
### Description

As an NGIpkgs user, I want to see the demo instructions just for my platform (e.g. Debian) so that I won't get distracted with irrelevant detail.

### Acceptance criteria

Given some demoable service,
When I view the demo instructions,
I can select the platform for which they are valid.

## Implementation notes

- Probably tabbed view for desktop, not sure about mobile
- Ideally clicking one tab switch all platform-specific tabs
- Default to the one with the simplest way of doing things

JuneStepp:
See commits.

image

image

Closes #1084.

### DIFF
--- a/overview/content-types/demo-instructions.nix
+++ b/overview/content-types/demo-instructions.nix
@@ -16,16 +16,6 @@ in
       type = types.submodule ./shell-instructions.nix;
       default.instructions = [
         {
-          platform = "Arch Linux";
-          shell-session.bash = [
-            {
-              input = ''
-                pacman --sync --refresh --noconfirm curl git jq nix
-              '';
-            }
-          ];
-        }
-        {
           platform = "Debian";
           shell-session.bash = [
             {
@@ -41,6 +31,26 @@ in
             {
               input = ''
                 apt install --yes curl git jq nix
+              '';
+            }
+          ];
+        }
+        {
+          platform = "Arch";
+          shell-session.bash = [
+            {
+              input = ''
+                pacman --sync --refresh --noconfirm curl git jq nix
+              '';
+            }
+          ];
+        }
+        {
+          platform = "Other";
+          shell-session.bash = [
+            {
+              input = ''
+                sh <(curl --proto '=https' --tlsv1.2 -L https://nixos.org/nix/install)
               '';
             }
           ];
@@ -87,7 +97,7 @@ in
           ];
         }
         {
-          platform = "Ubuntu 24.04";
+          platform = "Ubuntu 24.04/24.10";
           shell-session.bash = [
             {
               input = ''
--- a/overview/content-types/shell-instructions.nix
+++ b/overview/content-types/shell-instructions.nix
@@ -36,14 +36,21 @@ in
         self:
         if lib.isList self.instructions then
           ''
-            <dl>
-            ${lib.concatMapStringsSep "\n" (i: ''
-              <dt>${i.platform}</dt>
-              <dd>
-                ${toString i.shell-session}
-              </dd>
-            '') self.instructions}
-            </dl>
+            <div class=tabs>
+              {% set input_group_name = unique_id() %}
+              ${lib.concatStringsSep "\n" (
+                lib.zipListsWith (i: instruction: ''
+                  {% set input_id = unique_id() %}
+                  <input id="{{ input_id }}" type="radio" name="{{ input_group_name }}" ${
+                    lib.optionalString (i == 1) "checked"
+                  }>
+                  <label for="{{ input_id }}" >${instruction.platform}</label>
+                  <div class="tab-content">
+                    ${toString instruction.shell-session}
+                  </div>
+                '') (lib.range 1 (lib.length self.instructions)) self.instructions
+              )}
+            </div>
           ''
         else
           toString self.instructions;
--- a/overview/render-template.py
+++ b/overview/render-template.py
@@ -10,26 +10,28 @@ from pygments.formatters import HtmlFormatter
 jinja2_template_file = sys.argv[1]
 output_file = sys.argv[2]
 
+
 # Include code from a file and highlight it in HTML
 def include_code(language, file_path, relative_path=False):
-  if relative_path:
-    file_path = os.path.dirname(output_file) + "/" + file_path
-  with open(file_path) as file:
-    code = file.read()
-    try:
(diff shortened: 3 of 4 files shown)


overview/content-types/demo-instructions.nix
eljamm, line 1: In the `build-instructions`, do you think it would be better to split these up?

Image

So we'd have `Arch`, `Debian Sid/Trixie`, `Ubuntu 25.04` and `Ubuntu 24.04`.

overview/content-types/demo-instructions.nix
```diff
@@ -36,11 +36,21 @@ in
...
+            }
+          ];
+        }
+        {
+          platform = "Other";
+          shell-session.bash = [
+            {
+              input = ''
+                sh <(curl --proto '=https' --tlsv1.2 -L https://nixos.org/nix/install)
               '';
```
eljamm, line 54: We discussed using the installer before and decided that using the packaged Nix would be a better alternative, so I'm not sure if we'd like to add this. How about we keep the current platforms and discuss this again, later?

overview/content-types/demo-instructions.nix
```diff
@@ -87,7 +97,7 @@ in
           ];
         }
         {
-          platform = "Ubuntu 24.04";
+          platform = "Ubuntu 24.04/24.10";
```
eljamm, line 100: `Ubuntu 24.10` is EOL and as thus we can't test it, that's why [we dropped it](https://github.com/eljamm/ngipkgs/commit/8f5b5d52721407278e4e24e3382f6e35e6c4bfa2).

overview/content-types/shell-instructions.nix
```diff
@@ -36,14 +36,21 @@ in
...
+                    lib.optionalString (i == 1) "checked"
+                  }>
+                  <label for="{{ input_id }}" >${instruction.platform}</label>
+                  <div class="tab-content">
+                    ${toString instruction.shell-session}
+                  </div>
+                '') (lib.range 1 (lib.length self.instructions)) self.instructions
+              )}
+            </div>
           ''
```
eljamm, line 54: This is a bit more readable for me. What do you think?

```suggestion
          let
            instruction-ids = lib.range 1 (lib.length self.instructions);

            /**
              Return the ith tab input for the ith instruction.
            */
            make-tab-input = i: instruction: ''
              {% set input_id = unique_id() %}
              <input id="{{ input_id }}" type="radio" name="{{ input_group_name }}" ${
                lib.optionalString (i == 1) "checked"
              }>
              <label for="{{ input_id }}" >${instruction.platform}</label>
              <div class="tab-content">
              ${toString instruction.shell-session}
              </div>
            '';
          in
          ''
            <div class=tabs>
              {% set input_group_name = unique_id() %}
              ${lib.concatStringsSep "\n" (lib.zipListsWith make-tab-input instruction-ids self.instructions)}
            </div>
          ''
```

JuneStepp: Honestly, no.

eljamm: > `i` traditionally stands for index not id. Now it's confused with the IDs for HTML.

Well, it does correspond to the tab `id`, so how about renaming it to that?

> The separate variable makes it seem like the "ids" are all important when they aren't.

To me, it's rather nice describing what it is. When I looked at it the first time, I had to kinda decrypt what it's trying to do: take the length of the instructions and generate a range from 1 to that length, which still doesn't tell you what that range represents.

This said, this isn't a blocker for me.

> input_group_name isn't defined where it's used.

That's a fair point.

> The comment seems superfluous/giving the wrong detail.

> make-tab-input isn't an accurate name. Maybe just make-tab.

Yes, naming things is hard so I just put whatever came to mind ;)

> The most confusing part to me previously was using ziplistswith to get indexes, but now it feels even more unclear why its there.

So, instead of changing the code, how about we add a comment that describes what's happening?

Something in the lines of:
- we generate a unique id for each table
- we do the same for each tab and group them in the same table
- ...

overview/content-types/demo-instructions.nix
eljamm: If we include this method of installation, we'd have to support it to some degree, which might prove to be too much for us at this moment. We should also consider users, who'd not only have to trust this method, but who won't have a clue of how to uninstall either.

We can definitely discuss whether or not we'd like to add or support the installer and other platforms, but I feel like it would be better to do it outside this PR, which is kinda out of its scope.

overview/content-types/shell-instructions.nix
JuneStepp: > Well, it does correspond to the tab id, so how about renaming it to that?

It has no relation. It is only there so that the first tab can be marked as `checked`.

overview/content-types/demo-instructions.nix
JuneStepp: How about instead, something like "See https://nixos.org/download/". The "Other" tab only affects people who aren't supported by the first tabs, so it would just make things easier for them by linking to the page they would get to if they feel bothered to try and figure out how to install Nix on their (not covered by the first tabs) system.

overview/content-types/shell-instructions.nix
eljamm: I still don't like that we're making things too complex than they should (with zipping), especially since we'd only ever need the first id. If you're ok with this, we can simply use jinja:

```nix
''
  <div class=tabs>
    {% set input_group_name = unique_id() %}
    ${lib.concatMapStringsSep "\n" (instruction: ''
      {% set input_id = unique_id() %}
      <input id="{{ input_id }}" type="radio" name="{{ input_group_name }}" {% if input_id == "id-1" %} checked {% endif %}>
      <label for="{{ input_id }}" >${instruction.platform}</label>
      <div class="tab-content">
        ${toString instruction.shell-session}
      </div>
    '') self.instructions}
  </div>
''
```

overview/content-types/demo-instructions.nix
eljamm: Sure, please open a new PR and we can discuss this properly, there.

eljamm (approved):
Looks great, thanks! and also thank you @erictapen for your work on this as well.

PR #1684, opened 2025-10-14
Add update script & bump packages

eljamm:
Bumps were produced by the update script in https://github.com/ngi-nix/ngipkgs/pull/1680

**PS**: this PR is non-exhaustive and we probably want to figure out how to mass-trigger all updates, but it's a good start, especially since we need to first add the update scripts and make sure they work.

### DIFF
--- a/pkgs/by-name/nodebb/package-lock.json
+++ b/pkgs/by-name/nodebb/package-lock.json
@@ -1,66 +1,64 @@
 {
     "name": "nodebb",
-    "version": "4.4.3",
+    "version": "4.6.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "nodebb",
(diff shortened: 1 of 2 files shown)


pkgs/by-name/taldir/package.nix
```diff
@@ -39,6 +43,33 @@ buildGoModule (finalAttrs: {
   # dial error (dial tcp [::1]:5432: connect: connection refused)
   doCheck = false;
 
+  passthru = {
+    updateScriptSrc = gitUpdater { rev-prefix = "v"; };
+    updateScriptVendor = writeShellApplication {
+      name = "update-taldir-vendorHash";
+      runtimeInputs = [ nix ];
+      text = ''
```
imincik: Is this script way too hacky or not ? I am not sure.

Issue #1128, opened 2025-06-11
Implement project metadata for Funkwhale


eljamm:
### Instructions

Given ngi-nix/projects#105 for metadata about the application, follow the instructions in the [contributor documentation](https://github.com/ngi-nix/ngipkgs/blob/main/CONTRIBUTING.md#addingexposing-an-ngi-project) and implement project metadata for Funkwhale.

PR #1707, opened 2025-10-20
projects/Funkwhale: init

JuneStepp:
Ignore any funkiness in the code. That's just the whale. This PR requires https://github.com/NixOS/nixpkgs/pull/453976. (and thus is a draft right now)

I think the search tests in `tests/music/test_views` are little flaky. Could be an upstream issue.

I didn't add anything for plugins, since I couldn't find any not pre-included ones.
Same deal with frontend [settings.json](https://docs.funkwhale.audio/administrator/configuration/frontend.html). Didn't seem to be much use for it. Both of these should be easy to add later if needed.

Closes ngi-nix/ngipkgs#1128 
Closes ngi-nix/ngipkgs#1649 

Will resolve https://github.com/NixOS/nixpkgs/issues/114161 once upstreamed.


projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,538 @@
...
+        isSystemUser = true;
+        group = cfg.group;
+      };
+    };
+    users.groups = lib.mkIf (cfg.group == "funkwhale") {
+      ${cfg.group} = { };
+    };
+
+    environment.systemPackages = [
+      (pkgs.writeShellScriptBin "funkwhale-manage" ''
```
imincik: I would move this script creation to let binding to make configuration of `environment.systemPackages` cleaner

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,538 @@
...
+
+            before = [
+              "funkwhale-server.service"
+              "funkwhale-worker.service"
+              "funkwhale-beat.service"
+            ];
+            requiredBy = [ "funkwhale.target" ];
+
+            script = ''
+              if [[ ! -v DJANGO_SECRET_KEY ]]; then
```
imincik: I would move this script to let binding

imincik (changes requested):

projects/Funkwhale/demo/tests/basic.nix
```diff
@@ -0,0 +1,27 @@
...
+{
+  name = "funkwhale-demo-basic";
+
+  nodes = {
+    machine = {
+      imports = [
+        sources.modules.ngipkgs
+        sources.modules.services.funkwhale
+        sources.examples.Funkwhale."Basic local configuration"
+        ../module-demo.nix
```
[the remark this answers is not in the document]
JuneStepp: `../module-demo.nix`, isn't exposed in `sources`.

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,538 @@
...
+        ''}
+        set +o allexport
+
+        exec ${lib.getExe cfg.package} "$@"
+      '')
+    ];
+
+    systemd.tmpfiles.rules = [
+      # Where generated secrets will be.
+      "d /var/lib/funkwhale/config 0700 ${cfg.user} ${cfg.group} - -"
```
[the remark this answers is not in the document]
JuneStepp: That's just where auto-generated internally used secrets go. The user doesn't place anything there. I've updated the comment to say `auto-generated` instead of `generated` to make this clearer.

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,558 @@
...
+        CACHE_URL = "unix://${config.services.redis.servers.funkwhale.unixSocket}?db=0";
+        CELERY_BROKER_URL = "redis+socket://${config.services.redis.servers.funkwhale.unixSocket}?virtual_host=0";
+      })
+      (lib.mkIf cfg.configureNginx {
+        REVERSE_PROXY_TYPE = "nginx";
+        NGINX_MAX_BODY_SIZE = lib.mkDefault "100M";
+      })
+    ];
+
+    users.users = lib.mkIf (cfg.user == "funkwhale") {
```
imincik: Why is the user and group created only when it equals `funkwhale` ?

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,558 @@
...
+        group = cfg.group;
+        mode = "0755";
+      };
+      # Default `MUSIC_DIRECTORY_PATH`.
+      "/var/lib/funkwhale/data/music".d = {
+        user = cfg.user;
+        group = cfg.group;
+        mode = "0755";
+      };
+    };
```
imincik: Use systemd `StateDirectory` to create `/var/lib/funkwhale` directory with correct ownership and permissions.

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,558 @@
...
+        # Set `cfg.settings` environment variables.
+        ${lib.concatMapAttrsStringSep "\n" (
+          name: value: ''export ${name}=${lib.escapeShellArg value}''
+        ) cfg.settings}
+
+        # Set secret environment variables.
+        set -o allexport
+        if [ -f "/var/lib/funkwhale/config/django_secret_key.env" ]; then
+          source "/var/lib/funkwhale/config/django_secret_key.env"
+        fi
```
imincik: `/var/lib/funkwhale/config/django_secret_key.env` shouldn't be sourced when `cfg.secretEnvironmentFile` is set.

imincik (changes requested):

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,558 @@
...
+
+    systemd.tmpfiles.settings.funkwhale = {
+      # Where auto-generated secrets will be.
+      "/var/lib/funkwhale/config".d = {
+        user = cfg.user;
+        group = cfg.group;
+        mode = "0700";
+      };
+
+      # Niscellaneous API state files.
```
imincik:
```suggestion
      # Miscellaneous API state files.
```

projects/Funkwhale/services/funkwhale/module.nix
imincik: Still sounds strange to me and unfortunately this explanation doesn't help me.

pkgs/by-name/funkwhale/package.nix
```diff
@@ -0,0 +1,231 @@
...
+  meta = {
+    description = "Federated platform for audio streaming, exploration, and publishing";
+    homepage = "https://www.funkwhale.audio/";
+    changelog = "https://docs.funkwhale.audio/changelog.html";
+    license = lib.licenses.agpl3Only;
+    teams = [ lib.teams.ngi ];
+    mainProgram = "funkwhale-manage";
+  };
+in
+python3.pkgs.buildPythonApplication rec {
```
imincik: Nixpkgs manual suggests to use `python3Packages` (https://nixos.org/manual/nixpkgs/stable/#buildpythonapplication-function) but I am not sure if this makes any difference. 

```suggestion
python3Packages.buildPythonApplication rec {
```

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,559 @@
...
+      funkwhale = {
+        description = "Funkwhale";
+        wants = [
+          "funkwhale-server.service"
+          "funkwhale-worker.service"
+          "funkwhale-beat.service"
+        ];
+        wantedBy = [ "multi-user.target" ];
+      };
+    };
```
imincik: Can be also done using `systemd.slices`

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,559 @@
...
+            ++ lib.optionals cfg.createDatabaseLocally [ "postgresql.service" ]
+            ++ lib.optionals cfg.createRedisLocally [ "redis-funkwhale.service" ];
+
+            partOf = [ "funkwhale.target" ];
+
+            environment = cfg.settings;
+
+            serviceConfig = {
+              User = cfg.user;
+              Group = cfg.group;
```
imincik: User must be created otherwise service will fail. See comment above.

projects/Funkwhale/services/funkwhale/module.nix
JuneStepp: Can you expand on why slices might be used instead? I assume the main purpose of the target as Funkwhale uses it upstream is to be a single thing to put in After=, Wants=, ect.

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,559 @@
...
+          "funkwhale-server.service"
+          "funkwhale-worker.service"
+          "funkwhale-beat.service"
+        ];
+        wantedBy = [ "multi-user.target" ];
+      };
+    };
+
+    systemd.services =
+      lib.mapAttrs
```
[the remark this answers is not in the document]
JuneStepp: Ya. Would have to use `mapAttrs` either way.

projects/Funkwhale/services/funkwhale/module.nix
JuneStepp: I asked Kerstin and she doesn't get why this stuff is done either, so I've just removed the `.user` and `.group` options. If someone needs it later, it won't be hard to add.

projects/Funkwhale/services/funkwhale/module.nix
```diff
@@ -0,0 +1,559 @@
...
+        group = cfg.group;
+        mode = "0755";
+      };
+      # Default `MUSIC_DIRECTORY_PATH`.
+      "/var/lib/funkwhale/data/music".d = {
+        user = cfg.user;
+        group = cfg.group;
+        mode = "0755";
+      };
+    };
```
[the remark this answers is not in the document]
eljamm: Using tmpfiles isn't wrong, but this has been discussed lately and the conclusion was that it's better to let systemd create and manage its state directories to avoid issues, but also allow users to modify where they point to using [bind paths](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#BindPaths=) (see [example](https://github.com/ngi-nix/ngipkgs/pull/1913/changes)). This means that it's up to users to create and manage the bound dirs' permissions for existing directories, but for new ones [UMask](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#UMask=) can still be used (or [`StateDirectoryMode`](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectoryMode=)?).

**PS:** I personally have limited experience with this, so this is just my current understanding, but perhaps @ju1m can chime in and provide a clearer picture.

projects/Funkwhale/services/funkwhale/module.nix
eljamm: Forgot what we discussed last time about this, but it appears that [slices](https://www.freedesktop.org/software/systemd/man/latest/systemd.slice.html) are used to manage resources whereas [targets](https://www.freedesktop.org/software/systemd/man/latest/systemd.target.html) are used to group units, so it seems to me that using the latter is more correct in this case.

ju1m: Well, I was wrong again, AFAICS `system-funkwhale.slice` can replace `funkwhale.target`:
```
  systemd.slices.system-foo = {
    description = "Foo Slice";
    documentation = [ "https://foo.example.com/" ];
    wantedBy = [ "multi-user.target" ];
  };
  systemd.services.foo-service-1 = {
    wantedBy = [ "system-foo.slice" ];
    partOf = [ "system-foo.slice" ];
    serviceConfig = {
      ExecStart = "${lib.getExe' pkgs.coreutils "sleep"} infinity";
      Slice = "system-foo.slice";
    };
  };
  systemd.services.foo-service-2 = {
    wantedBy = [ "system-foo.slice" ];
    partOf = [ "system-foo.slice" ];
    serviceConfig = {
      ExecStart = "${lib.getExe' pkgs.coreutils "sleep"} infinity";
      Slice = "system-foo.slice";
    };
  };
```

```console
$ sudo systemctl status system-foo.slice
○ system-foo.slice - Foo Slice
     Loaded: loaded (/etc/systemd/system/system-foo.slice; enabled; preset: ignored)
     Active: inactive (dead) since Mon 2025-12-29 20:39:59 CET; 1s ago
   Duration: 1min 27.810s
 Invocation: ecbe7bb88de2419fb7acae3a57922975
       Docs: https://foo.example.com/
         IP: 0B in, 0B out
         IO: 0B read, 0B written
   Mem peak: 3.3M
        CPU: 20ms

déc. 29 20:38:32 nan2gua1 systemd[1]: Created slice Foo Slice.
déc. 29 20:39:59 nan2gua1 systemd[1]: Removed slice Foo Slice.

$ sudo systemctl start system-foo.slice
$ sudo systemctl status system-foo.slice
● system-foo.slice - Foo Slice
     Loaded: loaded (/etc/systemd/system/system-foo.slice; enabled; preset: ignored)
     Active: active since Mon 2025-12-29 20:40:08 CET; 1s ago
 Invocation: 7861d2113aa94e0eb47398efebbbea4b
       Docs: https://foo.example.com/
         IP: 0B in, 0B out
         IO: 0B read, 0B written
      Tasks: 2
     Memory: 560K (peak: 3.4M)
        CPU: 18ms
     CGroup: /system.slice/system-foo.slice
             ├─foo-service-1.service
             │ └─1274321 /nix/store/rry6qingvsrqmc7ll7jgaqpybcbdgf5v-coreutils-9.7/bin/sleep infinity
             └─foo-service-2.service
               └─1274322 /nix/store/rry6qingvsrqmc7ll7jgaqpybcbdgf5v-coreutils-9.7/bin/sleep infinity

déc. 29 20:40:08 nan2gua1 systemd[1]: Created slice Foo Slice.
```

PR #1725, opened 2025-10-22
feat: count maintained derivations in Nixpkgs

eljamm:
Add script that returns maintained derivations in Nixpkgs:

  ```shellSession
  # release tag
  nix eval --json -f maintainers/metrics-maintainer.nix packages --argstr rev 25.02 | jq '[.. | objects | select(has("name")) | .name]'

  # latest Nixpkgs
  nix eval --json -f maintainers/metrics-maintainer.nix packages --argstr rev nixpkgs | jq '[.. | objects | select(has("name")) | .name]'
  ```

And added derivation and update scripts count to metrics:

```shellSession
$ nix eval --json -f default.nix metrics.summary.nixos
evaluation warning: # ...
{
  "demos": 30,
  "derivations": 134,
  "examples": 95,
  "programs": 54,
  "services": 46,
  "tests": 98,
  "update-script": 108
}
```

### DIFF
--- a/maintainers/metrics.nix
+++ b/maintainers/metrics.nix
@@ -112,6 +112,67 @@ let
           > "$out/count.json"
       '';
 
+  /*
+    Summary of NGIpkgs and NixOS metrics.
+
+    # Outputs
+
+    ## NGIpkgs
+
+    Metrics related to the NGIpkgs monorepo.
+
+    - `derivations`
+
+      : Number of derivations maintained in NGIpkgs
+
+    - `update-scripts`
+
+      : Number of derivations maintained in NGIpkgs that have an explicit update script
+
+    - `metadata`
+
+      : Information about NGI-funded projects
+
+    - `metadata.subgrants`
+
+      : Attribute set that contains the number of projects for each [fund](maintainers/docs/project.md#libsubgrant)
+
+    - `projects`
+
+      : Number of NGI-funded projects in NGIpkgs
+
+    ## NixOS
+
+    Metrics related to NixOS.
+
+    - `derivations`
+
+      : Number of derivations maintained in Nixpkgs
+
+    - `update-scripts`
+
+      : Number of derivations maintained in Nixpkgs that have an explicit update script
+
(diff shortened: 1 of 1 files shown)


imincik:
I wrote a similar [script](https://github.com/ngi-nix/ngipkgs/pull/1725) to get list of packages maintained by person. Maybe you can take some inspiration for looping over package sets (like `python3Packages`, ...)

maintainers/metrics-maintainer.nix
```diff
@@ -0,0 +1,101 @@
...
+
+  # branch
+  nix eval -f ./metrics-maintainer.nix metrics.maintained --argstr team ngi --argstr rev main
+
+  # latest Nixpkgs
+  nix eval -f ./metrics-maintainer.nix metrics.maintained --argstr team ngi --arg nixpkgs 'fetchTarball "https://github.com/NixOS/nixpkgs/tarball/master"'
+*/
+{
+  rev ? "main",
+  ngipkgs ? import (fetchTarball "https://github.com/ngi-nix/ngipkgs/tarball/${rev}") { },
```
ethancedwards8: wouldn't it make more sense for this to default to the repo path but then we can override it manually to a rev/tag if we need to instead of it being the default?

ethancedwards8 (changes requested):

imincik:
> The script works beautifully, but I've noticed that it sometimes counts duplicates. For example, it takes into account both `python312` and `python313` packages for the same derivation.

Yes, this is intended. We want this granularity - for example package can fail with one python version but not with other.

maintainers/metrics.nix
imincik: Would you please document the output values of this script and the repository to which this value is related.

For example:
  
* "demos": number of demos maintained in NGIpkgs repo
* "derivations": number of derivations maintained in nixpkgs repo

....

Thanks.

imincik (changes requested):

imincik (approved):
Thank you Fedi

PR #1734, opened 2025-10-30
docs: fix links

phanirithvij:
I ran https://github.com/lycheeverse/lychee once locally in the repo to see how many links are dead.

### DIFF
--- a/overview/content-types/demo-instructions.nix
+++ b/overview/content-types/demo-instructions.nix
@@ -154,7 +154,9 @@ in
                 ${config.set-nix-config}
               </li>
               <li>
-                <strong>Build and run a virtual machine</strong>
+                <strong>Build and run ${
+                  if config.demo.type == "vm" then "a virtual machine" else "the shell"
+                }</strong>
                 ${config.build-instructions}
               </li>
               ${


maintainers/nixbuild.md
```diff
@@ -73,7 +73,7 @@ usually exactly the contents of the file
 
     ~/.ssh/id_ed25519.pub
 
-Navigate to <https://nixbuild.net/register>, enter your e-mail
+Navigate to <https://nixbuild.net/#register>, enter your e-mail
 address and public key.
```
phanirithvij: I don't see the site asking for public key, there is only email and it sends a code.

projects/owasp/programs/owasp/tests/depscan.nix
```diff
@@ -38,16 +38,16 @@
         # the program will try to download them at runtime.
         # https://github.com/owasp-dep-scan/dep-scan/issues/443
         environment.etc."vdb/data.index.vdb6".source = pkgs.fetchurl {
-          url = "https://huggingface.co/datasets/AppThreat/vdb/resolve/72377024f9742c6e700a113fc7059b18f738081c/app-2y/data.index.vdb6";
```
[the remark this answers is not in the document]
phanirithvij: I wonder if the previous commit was removed because it was added to nixpkgs? maybe caused a ton of traffic to huggingface so they removed it?

Otherwise can't think of a reason why that is missing. https://huggingface.co/datasets/AppThreat/vdb/tree/72377024f9742c6e700a113fc7059b18f738081c (dir https://huggingface.co/datasets/AppThreat/vdb/commits/main/app-2y)

projects/Inventaire/default.nix
```diff
@@ -27,7 +27,7 @@
         };
         settings-defaults = {
           text = "Default settings";
-          url = "https://codeberg.org/inventaire/inventaire/src/branch/main/config/default.cjs";
+          url = "https://codeberg.org/inventaire/inventaire/src/branch/main/config/default.ts";
```
eljamm: Perhaps a permalink would be better?


```suggestion
          url = "https://codeberg.org/inventaire/inventaire/src/commit/fd2083a2e19c690c198351d81ed34a8857e0ead6/config/default.ts";
```

projects/Inventaire/module.nix
```diff
@@ -20,7 +20,7 @@ let
   settingsLinks = ''
     Links for settings (make sure to look at the version that's actually packaged):
     - <a href="https://codeberg.org/inventaire/inventaire/src/branch/main/server/types/config.ts">Types</a>
-    - <a href="https://codeberg.org/inventaire/inventaire/src/branch/main/config/default.cjs">Defaults</a>
+    - <a href="https://codeberg.org/inventaire/inventaire/src/branch/main/config/default.ts">Defaults</a>
```
eljamm: Same suggestion as https://github.com/ngi-nix/ngipkgs/pull/1734/files#r2480377771

projects/owasp/programs/owasp/tests/depscan.nix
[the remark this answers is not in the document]
eljamm: > I think this file can be removed as we rely on `nixosTest.dep-scan` which is the same file, upstreamed to nixpkgs.

Yes, could you please do that? also remove the one for `blint` and mark them unbroken by removing [these lines](https://github.com/ngi-nix/ngipkgs/blob/ad9a77a45aee7b6b80e4d50a8aba12952ff6eb58/projects/owasp/default.nix#L41-L46)

CONTRIBUTING.md
```diff
@@ -64,7 +64,7 @@ Instead, write one sentence per line, as this makes it easier to review changes.
      nix fmt pkgs/by-name/some-package
      ```
 
-   An existing example is [libgnunetchat](https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/by-name/libgnunetchat/package.nix).
+   An existing example is [wireguard-rs](https://github.com/ngi-nix/ngipkgs/blob/main/pkgs/by-name/wireguard-rs/package.nix).
```
eljamm: Not sure if this is the best example, but let's use a permalink


```suggestion
   An existing example is [wireguard-rs](https://github.com/ngi-nix/ngipkgs/blob/ad9a77a45aee7b6b80e4d50a8aba12952ff6eb58/pkgs/by-name/wireguard-rs/package.nix).
```

projects/Inventaire/module.nix
eljamm: If that's the case, we shouldn't change this, since the [current inventaire version's default config file](https://codeberg.org/inventaire/inventaire/src/tag/v4.0.1/config/default.cjs) has the `.cjs` extension.

phanirithvij: Reverted to cjs for this file. I will squash all commits once you review.

projects/Inventaire/default.nix
```diff
@@ -23,11 +23,11 @@
       links = {
         settings-types = {
           text = "TypeScript file listing the intended types of every setting";
-          url = "https://codeberg.org/inventaire/inventaire/src/branch/main/server/types/config.ts";
+          url = "https://codeberg.org/inventaire/inventaire/src/commit/fd2083a2e19c690c198351d81ed34a8857e0ead6/server/types/config.ts";
         };
         settings-defaults = {
           text = "Default settings";
-          url = "https://codeberg.org/inventaire/inventaire/src/branch/main/config/default.cjs";
+          url = "https://codeberg.org/inventaire/inventaire/src/commit/fd2083a2e19c690c198351d81ed34a8857e0ead6/config/default.ts";
```
eljamm: I think these should be reverted since the current inventaire version has the `.cjs` extension

eljamm: I'd rather we don't do that, because we'd have to remember to keep this in sync. Let's keep the URLs how they were, for now.

PR #1784, opened 2025-11-11
demo: refactor and expose in the toplevel

eljamm:
Demos can now be triggered from the command line:

- `$(nix-build -A demos.PROJECT_NAME)`
- `nix build .#demos.PROJECT_NAME && ./result`

Supersedes https://github.com/ngi-nix/ngipkgs/pull/1528

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -252,26 +252,23 @@ rec {
     inherit lib pkgs metrics;
   };
 
-  project-demos = lib.filterAttrs (name: value: value != null) (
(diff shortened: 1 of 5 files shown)


imincik:
> I agree that it makes more sense and with flakes we can surely do it, but in default.nix we can't modify projects

Ok then. Would you mind to document `nix-build -A demos.PROJECT_NAME` somewhere ?

PR #1821, opened 2025-11-19
pkgs/pdfding: init at 1.4.1

### LINKED ISSUE #1820: Package PdfDing derivation for NGIpkgs/Nixpkgs
### Instructions

- For NGIpkgs:

  Follow the steps in the [contribution documentation](https://github.com/ngi-nix/ngipkgs/blob/main/CONTRIBUTING.md#how-to-create-pull-requests-to-ngipkgs) and package the derivation.

- For Nixpkgs:

  Follow the [contribution guide in the Nixpkgs repository](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#quick-start-to-adding-a-package) and package the derivation.

phanirithvij:
init frontend
init update script

closes #1820

### DIFF
--- a/pkgs/by-name/pdfding/frontend.nix
+++ b/pkgs/by-name/pdfding/frontend.nix
@@ -1,0 +1,86 @@
+{
+  stdenv,
+  nodejs,
+  npmHooks,
+  fetchpatch2,
+  fetchNpmDeps,
+  fetchzip,
+  fetchFromGitHub,
+  tailwindcss_4,
+}:
+stdenv.mkDerivation (
+  finalAttrs:
+  let
+    pdfjsVersion = "5.4.296"; # see update script
+    pdfjsHash = "sha256-b4W7wETq2CIZm2rJCmXEYvPhQtCbXS76L7GDvng6wn4=";
+    pdfjs = fetchzip {
+      url = "https://github.com/mozilla/pdf.js/releases/download/v${pdfjsVersion}/pdfjs-${pdfjsVersion}-dist.zip";
+      hash = pdfjsHash;
+      stripRoot = false;
+      postFetch = ''
+        rm -rf $out/web/locale \
+        $out/web/standard_fonts \
+        $out/web/compressed.tracemonkey-pldi-09.pdf
+
+        # remove source maps
+        find "$out" -name '*.map' -exec rm -f '{}' \;
+      '';
+    };
+  in
+  {
+    pname = "pdfding-frontend";
+    version = "1.4.1";
+    src = fetchFromGitHub {
+      owner = "mrmn2";
+      repo = "PdfDing";
+      tag = "v${finalAttrs.version}";
+      hash = "sha256-8e80gMdB6U3977dIU7bIAAEguYmi+AWQgUgYPDLCYLI=";
+    };
+
+    npmDeps = fetchNpmDeps {
+      inherit (finalAttrs) src;
+      name = "pdfding-frontend-${finalAttrs.version}-npm-deps";
+      hash = "sha256-v1NFqDnFcRK8sd0bV3ck+LLMYQ90Dl1R1OnBTwWUVUg=";
+    };
+
+    patches = [
+      # remove patch in 1.4.2
+      # package.json has missing version and name
+      (fetchpatch2 {
+        url = "https://github.com/mrmn2/PdfDing/pull/203.patch?full_index=1";
+        hash = "sha256-lKtpqKdyoGZdU4fTegto+YUIduIWbM82RQU9459NpC0=";
+      })
+    ];
+
+    nativeBuildInputs = [
+      nodejs
+      npmHooks.npmConfigHook
+      # it is in package.json and thus node_modules but no cli executable
+      tailwindcss_4
+    ];
+
+    # keeping the file structure same as upstream to minimise confusion
+    buildPhase = ''
+      runHook preBuild
+      mkdir -p $out/pdfding
+      cp -r --no-preserve=mode pdfding/static $out/pdfding/static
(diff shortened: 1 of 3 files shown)


pkgs/by-name/pdfding/package.nix
```diff
@@ -0,0 +1,194 @@
...
+    # remove in 1.4.2 (next version after 1.4.1)
+    # patch to add data_dir
+    # https://github.com/mrmn2/PdfDing/pull/202
+    (fetchpatch2 {
+      url = "https://github.com/mrmn2/PdfDing/commit/387ca2079f74844203e2e91fac00e0d0e0e5fdb9.patch?full_index=1";
+      hash = "sha256-VGjyIAVi+qd2WZ8FVKKC2ijLinoflO7RmPwIW1/oGcY=";
+    })
+    # pyproject.toml still has 0.1.1 very old version
+    # follow https://github.com/mrmn2/PdfDing/pull/203
+    ./add-version.patch
```
eljamm: No need to vendor the patch since it's been merged.

pkgs/by-name/pdfding/package.nix
```diff
@@ -0,0 +1,194 @@
...
+    # make an empty dir to supress the warning
+    mkdir -p $pdfdingDir/static
+
+    makeWrapper "$pdfdingDir/manage.py" $out/bin/pdfding-manage \
+      --set-default DATA_DIR "/var/lib/pdfding" \
+      --prefix PYTHONPATH : "${pythonPath}"
+
+    makeWrapper ${lib.getExe python.pkgs.gunicorn} $out/bin/pdfding-start \
+      --set-default DATA_DIR "/var/lib/pdfding" \
+      --prefix PYTHONPATH : "${pythonPath}:$pdfdingDir" \
```
eljamm:
```suggestion
      --prefix PYTHONPATH : "$out/${python.sitePackages}" \
```

pkgs/by-name/pdfding/package.nix
```diff
@@ -0,0 +1,194 @@
...
+
+  # dev.py is required for tests and MUST be removed from the final output
+  # this could be done in postCheck, but doing it here will allow doCheck to be toggleable
+  postPhases = [ "finalPhase" ];
+
+  finalPhase = ''
+    # dev.py should be removed on production build (source Dockerfile)
+    # can't be removed earlier, required for checkPhase
+    rm $out/${python.sitePackages}/pdfding/core/settings/dev.py
+  '';
```
eljamm: [Fixup phases](https://github.com/NixOS/nixpkgs/blob/master/doc/stdenv/stdenv.chapter.md#the-fixup-phase-ssec-fixup-phase) are precisely made for this purpose.

> The fixup phase performs (Nix-specific) post-processing actions on the files installed under `$out` by the install phase.

```suggestion
  postFixup = ''
    # dev.py should be removed on production build (source Dockerfile)
    rm $out/${python.sitePackages}/pdfding/core/settings/dev.py
  '';
```

pkgs/by-name/pdfding/package.nix
```diff
@@ -0,0 +1,194 @@
...
+
+  passthru = {
+    updateScript = ./update.sh;
+    inherit frontend;
+  };
+
+  meta = {
+    description = "Selfhosted PDF manager, viewer and editor offering a seamless user experience on multiple devices";
+    homepage = "https://github.com/mrmn2/PdfDing";
+    changelog = "https://github.com/mrmn2/PdfDing/blob/${src.tag}/CHANGELOG.md";
```
eljamm: This way, if we switch to unstable releases, the changelog will still work.

```suggestion
    changelog = "https://github.com/mrmn2/PdfDing/blob/${src.rev}/CHANGELOG.md";
```

pkgs/by-name/pdfding/package.nix
```diff
@@ -0,0 +1,194 @@
...
+
+    echo "VERSION = '${version}'" > pdfding/core/settings/version.py;
+  '';
+
+  postInstall = ''
+    mkdir -p $out/{bin,share}
+    pdfdingDir=$out/${python.sitePackages}/pdfding
+
+    # make an empty dir to supress the warning
+    mkdir -p $pdfdingDir/static
```
eljamm: I don't see a warning when this is missing.

phanirithvij: maybe it was happening in 1.3.3 when I first started packaging it, I can take a look at it, but its fine to leave it as is, I think.

pkgs/by-name/pdfding/package.nix
eljamm: We can rename the license in `postFetch` (so the patch could apply) and remove it after 1.4.2.

phanirithvij: done

pkgs/by-name/pdfding/frontend.nix
```diff
@@ -0,0 +1,80 @@
...
+      stripRoot = false;
+      postFetch = ''
+        rm -rf $out/web/locale \
+        $out/web/standard_fonts \
+        $out/web/compressed.tracemonkey-pldi-09.pdf
+
+        # remove source maps
+        find "$out" -name '*.map' -exec rm -f '{}' \;
+      '';
+    };
```
imincik: What about moving `pdfjs` to separate package ?

phanirithvij: I don't know if we would see much use of it in other packages. Also we have pdfjs's version tied to what pdfding specifies it the dockerfile.

eljamm: I agree with moving `passthru` down, but I think we keep `pdfjs` there because it's easier to override with `finalAttrs`.

eljamm (approved):

PR #1865, opened 2025-12-02
projects(SSTorytime): use module option to set db ownership

imincik:


projects/SSTorytime/services/sstorytime/module.nix
```diff
@@ -167,11 +168,5 @@ in
         host ${cfg.database.dbname} ${cfg.database.user} localhost trust
```
phanirithvij: @eljamm I think this is meant for that?

Edit: Maybe not.

Not many services are considering that possibilty of having a different name, as ensureDBOwnership description is this.

 > Grants the user ownership to a database with the same name.
 This database must be defined manually in
 `[](#opt-services.postgresql.ensureDatabases).`

eljamm: Yes, most services assume `dbname` and `user` will be the same, which doesn't seem right to me.

imincik:
Done in https://github.com/ngi-nix/ngipkgs/pull/1895 .

Issue #1867, opened 2025-12-04
Kaidan: package doesn't work


imincik:
From Michiel:

I'm trying to test Kaidan. When I try to find the right derivation to use this package in *ngipkgs* I get:

nix flake show --all-systems
git+file:///home/michiel/repo/ngipkgs?ref=refs/heads/main&rev=49573cacf9248262ff15a5d71b31ec45ca27aa26
├───checks
│ ├───aarch64-linux
│ │ ├───"infra/makemake": derivation 'nixos-system-makemake-26.05.20251130.23258e0'
│ │ ├───"infra/overview": derivation 'overview'
│ │ ├───"infra/pre-commit": derivation 'pre-commit-run'
│ │ ├───"packages/_0wm-ap-mock": derivation '0wm-ap-mock-0-unstable-2025-10-27'
│ │ ├───"packages/_0wm-client": derivation '0wm-client-0-unstable-2025-10-27'
│ │ ├───"packages/_0wm-opmode": derivation '0wm-opmode-0-unstable-2025-10-27'
│ │ ├───"packages/_0wm-server": derivation 'ocaml5.2.1-0wm-server-0-unstable-2025-11-24'
│ │ ├───"packages/anastasis": derivation 'anastasis-0.6.4'
│ │ ├───"packages/anastasis-gtk": derivation 'anastasis-gtk-0.6.3'
│ │ ├───"packages/anastasis-vmTest": derivation 'vm-test-run-anastasis-httpd'
│ │ ├───"packages/atomic-browser": derivation 'atomic-browser-0.40.0'
│ │ ├───"packages/atomic-cli": derivation 'atomic-cli'
│ │ ├───"packages/atomic-server": derivation 'atomic-server-0.40.0'
error: expected a derivation

When I use regular nixpkgs-unstable (which has the same version of Kaidan I believe), I get:

nix shell .#kaidan
error: builder for '/nix/store/88klqpf14dyndw4vk74ny8wavb8m89ag-kaidan-0.13.0.drv' failed with exit code 1;
last 25 log lines:
>
> * org.kde.kquickimageeditor-QMLModule, QML module 'org.kde.kquickimageeditor' is a runtime dependency.
>
> -- Configuring done (6.6s)
> CMake Error at src/app/CMakeLists.txt:27 (target_link_libraries):
> Target "kaidan" links to:
>
> Qt::GuiPrivate
>
> but the target was not found. Possible reasons include:
>
> * There is a typo in the target name.
[...]

PR #1868, opened 2025-12-04
fix: kaidan build with qt 6.10; switch to unstable

eljamm:
Closes https://github.com/ngi-nix/ngipkgs/issues/1867


imincik (approved):
Great !

PR #1913, opened 2025-12-16
projects(sylk): use bind paths instead of tmpfiles

eljamm:
This simplifies the configuration of service directories, which users can still modify using their respective settings options. This said, if users choose to change them, it's still up to them to make sure that those directories exist and have the correct permissions.


wamirez:
Related to https://github.com/ngi-nix/projects/issues/109.

Issue #1921, opened 2025-12-17
Reaction: Improve the demo example


phanirithvij:
### Instructions

The current demo for Reaction is a ssh authentication failure demo.
The demo works fine and does do its job of demonstrating reaction's banning capability but there is a minor issue with it.

```console
# cd ngipkgs and run the vm
$ nix-build -A demos.Reaction && ./result

# run inside the vm
$ journalctl -fu reaction -u sshd
```

```console
# run in a new terminal on the host, Don't enter the password
$ ssh -p 10022 nixos@localhost -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null
```

This happens immediately without even the user entering any password.

```console
$ journalctl -fu reaction -u sshd
Dec 17 17:50:23 nixos unix_chkpwd[905]: password check failed for user (nixos)
Dec 17 17:50:23 nixos sshd-session[903]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=10.0.2.2  user=nixos
Dec 17 17:50:24 nixos reaction[762]:  INFO ssh.failedlogin: match ["10.0.2.2"]
```

The issue is with the first line in the log. This causes reaction to register that a failure has occured. (demo is not impacted because it is okay for it to fail for the demo)

The above issue wasn't reproducible in a real nixos bare metal deployment, but can be reproduced in the demo and nixostest vms. So it's an issue with how `unix_chkpwd` works inside the vm.

We have two options:
- [ ] investigate and fix this for vms.
- [ ] use another example as a demo other than ssh for reaction, eg. https://reaction.ppom.me/filters/web-crawlers.html nginx config.

eljamm:
> There is a vendored `example-ssh.jsonnet` file in ngipkgs tree, but it's not there in the user's current directory so `settingsFiles = [ ./example-ssh.jsonnet ];` would fail for sure. (only way as of now to run reaction's demo is to use `nix-build -A demos.Reaction` after cloning ngipkgs)

We're already [instructing users to download a config file](https://github.com/ngi-nix/ngipkgs/blob/dfab738d4a1d00f6c1b958be29163d672badf05f/projects/Reaction/default.nix#L66). Just change this link to the vendored one in NGIpkgs.

phanirithvij:
I was under the assumption that a demo should work with a copy paste and build. If that's an option then great

eljamm:
[...]

PR #1922, opened 2025-12-17
demo: fix trusted-public-keys, case sensitive

phanirithvij:
We have had wrong instructions for a while. Noticed that these should be case sensitive, copied them from https://app.cachix.org/cache/ngi and https://github.com/NixOS/nixpkgs/blob/233f4672b905e51077ecd1dccf6dde02e5f3c3c6/nixos/modules/config/nix.nix#L371 respectively.

I had a call with @ppom0 to try to debug https://github.com/ngi-nix/ngipkgs/issues/1921 and we noticed this when trying to run the demo. We faced several issues (#1923) this being one of them.

### DIFF
--- a/overview/content-types/demo-instructions.nix
+++ b/overview/content-types/demo-instructions.nix
@@ -59,8 +59,8 @@ in
           "https://ngi.cachix.org/"
         ];
         trusted-public-keys = [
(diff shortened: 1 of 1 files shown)


eljamm (approved):
Thanks!

PR #1927, opened 2025-12-27
reuse(devLib): expose extended `lib` by default and use an overlay to extend it

ju1m:
+reuse: exposing the extended `lib` by default enables to reuse it in modules
evaluated by `pkgs.testers.runNixOSTest` in `projects/tests.nix`,
otherwise `lib` given to modules when testing is Nixpkgs' `pkgs.lib`.

+clarity: being able to factorize code in a `lib`rary
is much easier to check, maintain and develop
than having to manage multiple copies of the same code.

+standardization: having some code in a `lib`rary
nudges more to use it than if it would require to copy,
since it's already an explicitly shared pattern.

+customizability: using an overlay enables to customize `lib`'s
sub attrs (eg. adding types to `lib.types`).

-portability: using a custom `lib` specific to NGIpkgs
means the code is less portable, eg. pushing it into Nixpkgs or NixOS
means pushing the part of `lib` it depends on with it,
or to copy it in the package or module.

-installability: using a custom `lib` specific to NGIpkgs
means the modules must be imported with that `lib` by users,
which may not be obvious.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -14,12 +14,13 @@
   lib ? import "${sources.nixpkgs}/lib",
 }:
 let
-  devLib = import ./pkgs/lib.nix { inherit lib sources system; };
+  devLibOverlay = import ./pkgs/lib.nix { inherit sources system; };
+  devLib = lib.extend devLibOverlay;
 
   flakeAttrs = default.import ./maintainers/flake { };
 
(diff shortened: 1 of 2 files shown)


wamirez:

PR #1931, opened 2025-12-28
libervia-backend: clean up dependencies to rely on upstream packages

ethancedwards8:
Many dependencies in `libervia-backend` are upstreamed in Nixpkgs, use those instead.


ethancedwards8:
I tried removing the memo packages and using the upstream versions but I was having trouble with the build when I did...

PR #1939, opened 2026-01-03
projects(Teamtype): use demo module in tests without examples

### LINKED ISSUE #1491: Use demo module inside demo tests without needing examples
Currently, we can only use example configs inside tests (see [example](https://github.com/ngi-nix/ngipkgs/blob/c965815880b8d17ec84a00152abe7512c74036bb/projects/Ethersync/programs/ethersync/tests/basic.nix#L22)), which means that we we'll need to expose one, even if all we need is a demo (see https://github.com/ngi-nix/ngipkgs/pull/1490).

Aman16102000:
Fixes #1491

### DIFF
--- a/projects/Teamtype/default.nix
+++ b/projects/Teamtype/default.nix
@@ -34,11 +34,6 @@
       module = ./programs/teamtype/module.nix;
       examples."Enable Teamtype" = {
         module = ./programs/teamtype/examples/basic.nix;
-        tests.basic.module = import ./programs/teamtype/tests/basic.nix args;
-        tests.basic.problem.broken.reason = ''
-          Needs a self-hosted relay server to work non-interactively (without internet).
-          Requires: https://github.com/teamtype/teamtype/issues/344
-        '';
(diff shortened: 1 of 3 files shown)


projects/Teamtype/default.nix
```diff
@@ -34,11 +34,6 @@
       module = ./programs/teamtype/module.nix;
       examples."Enable Teamtype" = {
         module = ./programs/teamtype/examples/basic.nix;
-        tests.basic.module = import ./programs/teamtype/tests/basic.nix args;
-        tests.basic.problem.broken.reason = ''
-          Needs a self-hosted relay server to work non-interactively (without internet).
-          Requires: https://github.com/teamtype/teamtype/issues/344
-        '';
```
eljamm: Self-hosted relays still don't exist AFAIK, so the test shouldn't be unbroken.

wamirez:
Related to https://github.com/ngi-nix/projects/issues/25.

PR #1907, opened 2025-12-15
dnsvizor: init at 0-unstable-2025-12-15

jian-lin:
To build this locally, run command like this:

```sh
nix build -f . dnsvizor.unix
```

```sh
nix build -f . dnsvizor.hvt
```

### Progress

- [X] build `unix` target
- [X] build `hvt` target
- [X] build `spt` target
- [X] build `xen` target
- [X] build `qubes` target
- [X] build `virtio` target
- [X] build `muen` target
- [x] try to get a valid `version` for nix, currently `version = "dev"`
- [ ] move [my][3] [patches][4] to the right™ places
  - https://github.com/robur-coop/dnsvizor/pull/114
  - https://github.com/RyanGibb/hillingar/pull/16
  - https://github.com/RyanGibb/hillingar/pull/15
- [x] make buildbot-nix CI pass with IFD
- [ ] fix runtime error of the hvt target (discovered when working on https://github.com/ngi-nix/ngipkgs/pull/1944), [more info](https://github.com/ngi-nix/ngipkgs/pull/1907#issuecomment-3728667791)

The `macosx` and `genode` targets fail to build.  They seem to be niche targets so I do not plan to fix them.  So they are not exposed to avoid CI errors.

### Optional TODO list
- [ ] enable tests: tests fail to build if target is not `unix`
- [ ] avoid IFD (optional because CI passes now even with IFD)

[1]: https://github.com/RyanGibb/hillingar
[2]: https://github.com/robur-coop/dnsvizor
[3]: https://github.com/linj-fork/hillingar/tree/local/debug-clean
[4]: https://github.com/linj-fork/dnsvizor/tree/local/debug-clean
[5]: https://github.com/linj-fork/hillingar/commit/750327155eb31a49e276cafb33c733a3b88c171c

Closes #1906

PR #1762, opened 2025-11-07
projects(MirageOS): init

eljamm:
Closes https://github.com/ngi-nix/ngipkgs/issues/1761

PR #1944, opened 2026-01-09
projects/DNSvizor: init

### LINKED ISSUE #1956: DNSvizor: Implement a demo example
### Instructions

Use metadata about the application from its tracking issue and implement a demo.

### LINKED ISSUE #1957: DNSvizor: Implement a `recursive DNS resolver` example for service module
### Instructions

Use metadata about the project from its tracking issue and implement an example for the module.

> [!NOTE]
> For an example of how this looks like, please refer to the [project template](https://github.com/ngi-nix/ngipkgs/tree/main/maintainers/templates/project).

### LINKED ISSUE #1959: DNSvizor: Implement NixOS VM test for example `recursive DNS resolver`
### Instructions

Use metadata about the application from its tracking issue and implement a NixOS VM test for the example.

> [!NOTE]
>
> - The test needs to validate the basic functionality of the example configuration.
> - For an example of how this looks like, please refer to the [project template](https://github.com/ngi-nix/ngipkgs/tree/main/maintainers/templates/project).

### LINKED ISSUE #1960: DNSvizor: Implement a `stub DNS resolver` example for service module
### Instructions

Use metadata about the project from its tracking issue and implement an example for the module.

> [!NOTE]
> For an example of how this looks like, please refer to the [project template](https://github.com/ngi-nix/ngipkgs/tree/main/maintainers/templates/project).

### LINKED ISSUE #1961: DNSvizor: Implement NixOS VM test for example `stub DNS resolver`
### Instructions

Use metadata about the application from its tracking issue and implement a NixOS VM test for the example.

> [!NOTE]
>
> - The test needs to validate the basic functionality of the example configuration.
> - For an example of how this looks like, please refer to the [project template](https://github.com/ngi-nix/ngipkgs/tree/main/maintainers/templates/project).

jian-lin:
This PR is self-contained and is **not** blocked by the DNSvizor packaging PR

I create project DNSvizor instead of working under [project MirageOS][1] because:
1. I think it is more reasonable to have one project for each MirageOS unikernel
2. Each project can only has one demo.  I do not think it makes much sense to demo all these unikernels together.  They are not tightly connected.

I am open to moving this into project MirageOS if that is indeed a better organization.

This PR depends on:
- https://github.com/NixOS/nixpkgs/pull/479944
  - thus https://github.com/ngi-nix/ngipkgs/pull/1983

## TODO / Future work

- Things marked with `TODO(linj)` are nice-to-have, but adding them needs more time.
- `TODO` comments are blocked by other things such as [upstream][3] or the [packaging][2].  When their blockers are fixed, they can be added in future PRs.

## Run tests

```
nix build -f . projects.DNSvizor.tests
```

[1]: https://github.com/ngi-nix/ngipkgs/pull/1762
[2]: https://github.com/ngi-nix/ngipkgs/pull/1907
[3]: https://github.com/robur-coop/dnsvizor/issues/117

---

almost `fix` #1958
fix #1957
fix #1959
fix #1960
fix #1961
fix #1956

### DIFF
--- a/projects/DNSvizor/default.nix
+++ b/projects/DNSvizor/default.nix
@@ -1,0 +1,315 @@
+{
+  lib,
+  pkgs,
+  sources,
+  ...
+}@args:
+
+let
+  problem =
+    if lib.meta.availableOn pkgs.stdenv.hostPlatform pkgs.syslinux then
+      null
+    else
+      {
+        broken.reason = ''
+          dependency pkgs.syslinux is not available on this platform
+        '';
+      };
+  webInterfaceManual = {
+    text = "Web interface manual";
+    url = "https://robur-coop.github.io/dnsvizor-handbook/dnsvizor_web_interface.html";
+  };
+in
+{
+  metadata = {
+    summary = "Privacy-enhanced, secure and robust DNS resolver and DHCP server with a small resource footprint as a MirageOS unikernel";
+    subgrants = {
+      Entrust = [ "DNSvizor" ];
+    };
+    links = {
+      homepage = null;
+      repo = {
+        text = "Source repository";
+        url = "https://github.com/robur-coop/dnsvizor";
+      };
+      docs = {
+        text = "Handbook";
+        url = "https://robur-coop.github.io/dnsvizor-handbook/";
+      };
+      blog = {
+        text = "Blog";
+        url = "https://blog.robur.coop/tags.html#tag-DNSvizor";
+      };
+      upstreamBuilds = {
+        text = "Reproducible unikernel binaries built by upstream";
+        url = "https://builds.robur.coop/job/dnsvizor";
+      };
+      mirageOS = {
+        text = "MirageOS";
+        url = "https://mirage.io/";
+      };
+    };
+  };
+
+  nixos.modules.services.dnsvizor = {
+    name = "DNSvizor";
+    module = ./services/dnsvizor/module.nix;
+    examples =
+      let
+        mkExample =
+          {
+            exampleName,
+            exampleModule,
+            exampleDescription,
+            testName,
+            testModule,
+            testCfg,
+          }:
+          lib.nameValuePair exampleName {
+            module = exampleModule;
+            description = exampleDescription;
+            tests =
+              let
+                testArgToString =
+                  testArgName: testArgValue: seperator:
+                  if testArgValue == true then
+                    seperator + testArgName
+                  else if testArgValue == false then
+                    ""
+                  else if lib.isString testArgValue then
+                    seperator + testArgValue
+                  else
+                    throw "testArgToString: not implemented for ${testArgName}=${toString testArgValue}";
+                testArgsToString =
+                  testArgs:
+                  lib.pipe testArgs [
+                    (lib.mapAttrs (testArgName: testArgValue: "${testArgToString testArgName testArgValue "-"}"))
+                    lib.attrValues
+                    lib.concatStrings
+                  ];
+                mkTest =
+                  testArgs:
+                  lib.nameValuePair ("${testName}${testArgsToString testArgs}") {
+                    module = import testModule (args // testArgs // { inherit exampleName; });
+                    inherit problem;
+                  };
+              in
+              lib.listToAttrs (map mkTest (lib.cartesianProduct testCfg));
+            links = { inherit webInterfaceManual; };
+          };
+        dnsResolverExampleDescription = ''
+          Usage instructions
+
+          1. Query DNS with UDP/TCP, DNS-over-TLS(DoT) and DNS-over-HTTPS(DoH)
+          2. Open the web interface for management
+        '';
+      in
+      lib.listToAttrs (
+        map mkExample [
+          {
+            exampleName = "Enable DNSvizor as a IPv4-only stub DNS resolver";
+            exampleModule = ./services/dnsvizor/examples/stub-dns-resolver.nix;
+            exampleDescription = dnsResolverExampleDescription;
+            testName = "dns-ipv4";
+            testModule = ./services/dnsvizor/tests/dns.nix;
+            testCfg = {
+              resolverKind = [ "stub" ];
+              useNetworkd = [
+                true
+                false
+              ];
+              useNftables = [
+                true
+                false
+              ];
+            };
+          }
+          {
+            exampleName = "Enable DNSvizor as a dual-stack recursive DNS resolver";
+            exampleModule = ./services/dnsvizor/examples/recursive-dns-resolver.nix;
+            exampleDescription = dnsResolverExampleDescription;
+            testName = "dns-dualstack";
+            testModule = ./services/dnsvizor/tests/dns.nix;
+            testCfg = {
+              resolverKind = [ "recursive" ];
+              useNetworkd = [
+                true
+                false
+              ];
+              useNftables = [
+                true
+                false
+              ];
+            };
+          }
+        ]
+      );
+  };
+
+  nixos.demo.vm = {
+    module = ./services/dnsvizor/examples/recursive-dns-resolver.nix;
+    module-demo = ./demo/module-demo.nix;
+    usage-instructions = [
+      {
+        instruction = ''
+          Visit <https://127.0.0.1:4443> in your browser.  This is a web interface of DNSvizor.
+
+          You'll see a warning of potential security risk.
+          Rest assured.
+          This is because DNSvizor uses a self-signed certification.
+
+          Accept the risk and continue.
+        '';
+      }
+      {
+        instruction = ''
+          In the demo VM terminal, run this to send DNS queries to DNSvizor (`10.0.0.2`):
+
+          ```shellSession
+          $ q --verbose www.example.com A @10.0.0.2
+          ```
+
+          If the query timeouts, re-run the command to query again.
+        '';
+      }
+      {
+        instruction = ''
+          You can also use an IPv6 address for DNSvizor (`fdc9:281f:4d7:9ee9::2`):
+
+          ```shellSession
+          $ q --verbose www.example.com A @fdc9:281f:4d7:9ee9::2
+          ```
+        '';
+      }
+      {
+        instruction = ''
+          You can also use a domain name for DNSvizor (`dnsvizor.mydomain.example`):
+
+          ```shellSession
+          $ q --verbose www.example.com A @dnsvizor.mydomain.example
+          ```
+        '';
+      }
+      {
+        instruction = ''
+          Send encrypted DNS queries using `DNS-over-TLS`:
+
+          ```shellSession
+          $ q --verbose www.example.com A @tls://dnsvizor.mydomain.example
+          ```
+
+          You'll see an error telling you that certification verification failed.
+          This is because DNSvizor uses a self-signed certification, which is not trusted by default.
+          You can ignore that error by adding `--tls-insecure-skip-verify`:
+
+          ```shellSession
+          $ q --verbose --tls-insecure-skip-verify www.example.com A @tls://dnsvizor.mydomain.example
+          ```
+        '';
+      }
+      {
+        instruction = ''
+          Let's show you how to trust that self-signed certification of DNSvizor.
+
+          First, we extract that certification using `curl`.
+
+          ```shellSession
+          $ curl --write-out %{certs} https://dnsvizor.mydomain.example > /tmp/self-signed-cert.pem
+          ```
+
+          Then we setting environment variable `SSL_CERT_FILE` before running `q`.
+
+          ```shellSession
+          $ SSL_CERT_FILE=/tmp/self-signed-cert.pem q --verbose www.example.com A @tls://dnsvizor.mydomain.example
+          ```
+        '';
+      }
+      {
+        instruction = ''
+          Send encrypted DNS queries using `DNS-over-HTTPS`:
+
+          ```shellSession
+          $ SSL_CERT_FILE=/tmp/self-signed-cert.pem q --verbose www.example.com A @https://dnsvizor.mydomain.example
+          ```
+        '';
+      }
+      {
+        instruction = ''
+          Go back to the [Dashboard](https://127.0.0.1:4443/dashboard) page, you should see those numbers, such as `Total queries`, have changed.
+        '';
+      }
+      {
+        instruction = ''
+          Go to the [Query log](https://127.0.0.1:4443/querylog) page, you should see domains you just queried.
+        '';
+      }
+      {
+        instruction = ''
+          DNSvizor supports blocking DNS resolution for some domains.
+          You can specify them as boot parameters.
+
+          Go to the [Blocklist](https://127.0.0.1:4443/blocklist) page.
+          Enter the password `password`.
+          User can be anything you like.
+          You can see some blocked domains we already specified.
+
+          Query one of them:
+
+          ```shellSession
+          $ q --verbose --format raw block1.cli.example.com A @dnsvizor.mydomain.example
+          ```
+
+          You should see `status: NXDOMAIN` and `ANSWER: 0` in the output.
+          This means there is no answer to your DNS query.
+          You should also see `appears in blocklist boot-parameter` in the output.
+        '';
+      }
+      {
+        instruction = ''
+          Blocked domains can also be specified using URLs.
+          DNSvizor will fetch them using those URLs.
+
+          Go to the [Blocklist](https://127.0.0.1:4443/blocklist) page, you can see some blocked domain lists we already specified.
+
+          Query one of them:
+
+          ```shellSession
+          $ q --verbose --format raw block1.url.example.com A @dnsvizor.mydomain.example
+          ```
+
+          You should see `status: NXDOMAIN` and `ANSWER: 0` in the output.
+          You should also see `appears in blocklist http://10.0.0.1/block-list-4` in the output.
+        '';
+      }
+      {
+        instruction = ''
+          In the [Blocklist](https://127.0.0.1:4443/blocklist) page, you can also add or delete blocked domains.
+
+          Add one, query it and check the result.
+
+          You should see a similar output as before.  But this time, it shows `appears in blocklist web-ui`.
+
+          Delete the domain you just added, query it and check the result.
+
+          You should see the normal result again.
+        '';
+      }
+      {
+        instruction = ''
+          Now go back to the [Dashboard](https://127.0.0.1:4443/dashboard) page, you should see the number of "Queries blocked" has changed.
+        '';
+      }
+    ];
+    tests.dns-dualstack-recursive.module = import ./services/dnsvizor/tests/dns.nix (
+      args
+      // {
+        exampleName = "Enable DNSvizor as a dual-stack recursive DNS resolver";
+        resolverKind = "recursive";
+        useNetworkd = false;
+        useNftables = false;
+      }
+    );
+    inherit problem;
+    links = { inherit webInterfaceManual; };
+  };
+}
--- a/projects/DNSvizor/services/dnsvizor/examples/recursive-dns-resolver.nix
+++ b/projects/DNSvizor/services/dnsvizor/examples/recursive-dns-resolver.nix
@@ -1,0 +1,30 @@
+{ ... }:
+
+{
+  services.dnsvizor = {
+    enable = true;
+    memory = 128;
+    mainInterface = "enp1s0";
+    settings = {
+      hostname = "dnsvizor.mydomain.example";
+      ipv4 = "10.0.0.2/24";
+      ipv4-gateway = "10.0.0.1";
+      ipv6 = "fdc9:281f:4d7:9ee9::2/64";
+      ipv6-gateway = "fdc9:281f:4d7:9ee9::1";
+      ca-seed = "Te9ffyY3Clcaz/4P7eFLyZQfLWIz/fSSK4NDb8THMDc=";
+      password = "password";
+      dns-block = [
+        "block1.cli.example.com"
+        "block2.cli.example.com"
+      ];
+      dns-blocklist-url = [
+        "http://10.0.0.1/block-list-4"
+        "http://[fdc9:281f:4d7:9ee9::1]:80/block-list-6"
+        "https://example.com/non-existent-block-list"
+      ];
+      qname-minimisation = true;
+      opportunistic-tls-authoritative = true;
+    };
+    openFirewall = true;
+  };
+}
--- a/projects/DNSvizor/services/dnsvizor/module.nix
+++ b/projects/DNSvizor/services/dnsvizor/module.nix
@@ -1,0 +1,457 @@
+{
+  config,
+  lib,
+  pkgs,
+  utils,
+  ...
+}:
+
+# TODO(linj) implement and test DHCP
+#   - run dnsvizor as a DHCP server
+#   - update DNS record in the authoritative DNS server when DHCP ip changes
+#   - update config for tlstunnel mirageos unikernel when DHCP ip changes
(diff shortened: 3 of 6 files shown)


projects/DNSvizor/default.nix
```diff
@@ -0,0 +1,108 @@
...
+        text = "MirageOS";
+        url = "https://mirage.io/";
+      };
+    };
+  };
+
+  nixos.modules.services.dnsvizor = {
+    name = "DNSvizor";
+    module = ./services/dnsvizor/module.nix;
+    examples = {
```
jian-lin: Changing some config options of existing examples leads to similar new examples.  Testing those new examples increase test coverage.  However, I do not think it is a good idea to list a lot of very similar examples in the `overview` page.  WDYT?

projects/DNSvizor/services/dnsvizor/module.nix
```diff
@@ -0,0 +1,475 @@
...
+        assertion = !(onlyOneIsNull cfg.settings.ipv6 cfg.settings.ipv6-gateway);
+        message = "services.dnsvizor.settings.{ipv6,ipv6-gateway} must be set together.";
+      }
+      {
+        assertion = cfg.settings.ipv6-only != "true";
+        message = "services.dnsvizor.settings.ipv6-only has not been implemented.";
+      }
+    ];
+
+    # TODO remove this after the dnsvizor package is merged into ngipkgs
```
Prince213: Have you opened a PR?  If so, add a link here.

projects/DNSvizor/services/dnsvizor/module.nix
```diff
@@ -0,0 +1,475 @@
...
+          "@system-service"
+          "~@privileged"
+          "~@resources"
+        ];
+        SystemCallErrorNumber = "EPERM";
+        UMask = "0077";
+      };
+    };
+
+    networking.useNetworkd = true;
```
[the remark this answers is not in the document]
Prince213: I use it on a daily basis, so I think it's mostly stable.  Or at least as stable as systemd-networkd itself.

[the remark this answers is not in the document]
jian-lin: It seems there is a misunderstanding here.

Enabling `networking.useNetworkd` translates `networking.*` into systemd-networkd config files.  I am asking the stability of this translation.

`systemd.network.*` options natively translating into systemd-networkd config files.  I am not asking the stability of these options.

projects/DNSvizor/services/dnsvizor/module.nix
```diff
@@ -0,0 +1,473 @@
...
+        RestrictAddressFamilies = [ "AF_NETLINK" ];
+        RestrictNamespaces = true;
+        RestrictRealtime = true;
+        SystemCallArchitectures = [ "native" ];
+        SystemCallFilter = [
+          "@system-service"
+          "~@privileged"
+          "~@resources"
+        ];
+        SystemCallErrorNumber = "EPERM";
```
ju1m:
```suggestion
```

I know [`SystemCallErrorNumber=EPERM` appears in systemd's manual](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#SystemCallFilter=), but I've had a bad experience with it because the software was catching `EPERM` and just changing its behavior instead of raising an error, not letting me know I was constraining it too much.
I suggest to keep the default (`SIGSYS`) that is more likely to let us know we're constraining the software too much.

projects/DNSvizor/services/dnsvizor/tests/dns.nix
```diff
@@ -0,0 +1,609 @@
+# To test dnsvizor running as a recursive DNS resolver, we setup a
+# root DNS server, a TLD DNS server and an authoritative DNS server.
+# To let dnsvizor query our root DNS server instead of real root DNS
+# servers, we setup a firewall to deceive dnsvizor.
```
ju1m: AFAIU the DNAT can be replaced by static addresses:
- on `nodes.rootDnsServer`: adding root server addresses to the relevant network interface with: `networking.interfaces.eth1.ipv{4,6}.addresses`.
- on `nodes.dnsResolver` adding static routes to with `networking.interfaces.eth0.ipv{4,6}.routes`.

This is much simpler to understand and write than DNAT.
This removes the need for half of the tests by removing `useNftables`.

See my implementation in https://github.com/ju1m/ngipkgs/commits/dnsvizor-tests-routing/

projects/DNSvizor/services/dnsvizor/tests/dns.nix
```diff
@@ -0,0 +1,609 @@
...
+        if resolverKind == "stub" then "forwarding to" else "listening on"
+      }"')
+      # we assume the DNS block list is loaded after it is accessed on the web server
+      dns_resolver.wait_for_unit("caddy.service")
+      dns_resolver.wait_until_succeeds("journalctl -u caddy -g http.log.access")
+
+      dns_client.log("I am the DNS client")
+
+      with subtest("Web interface can be accessed"):
+          web_interface_url = "https://${webInterfaceDomainOrIp}"
```
ju1m: Not really important, so only answer me if you already know, but I am unable to reach that webpage from the host using `virtualisation.forwardPorts` in `interactive` mode, are you?
AFAIU, the module setups a NAT to `tap-unikernel`:

```console
[root@dnsResolver:~]# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:54:00:12:34:56 brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    altname ens3
    altname enx525400123456
    inet 10.0.2.15/24 brd 10.0.2.255 scope global dynamic noprefixroute eth0
       valid_lft 86226sec preferred_lft 75426sec
    inet6 fec0::de2b:2df8:56bf:352/64 scope site temporary dynamic
       valid_lft 86223sec preferred_lft 14223sec
    inet6 fec0::5054:ff:fe12:3456/64 scope site dynamic mngtmpaddr noprefixroute
       valid_lft 86223sec preferred_lft 14223sec
    inet6 fe80::5054:ff:fe12:3456/64 scope link
       valid_lft forever preferred_lft forever
3: enp1s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:54:00:12:01:03 brd ff:ff:ff:ff:ff:ff
    altname enp0s10
    altname ens10
    altname enx525400120103
    inet 192.168.1.3/24 scope global enp1s0
       valid_lft forever preferred_lft forever
    inet6 2001:db8:1::3/64 scope global nodad
       valid_lft forever preferred_lft forever
    inet6 fe80::5054:ff:fe12:103/64 scope link proto kernel_ll
       valid_lft forever preferred_lft forever
4: tap-unikernel: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether ee:ad:f0:71:c1:63 brd ff:ff:ff:ff:ff:ff
    inet 10.0.0.1/24 scope global tap-unikernel
       valid_lft forever preferred_lft forever
    inet6 fdc9:281f:4d7:9ee9::1/64 scope global nodad
       valid_lft forever preferred_lft forever
    inet6 fe80::ecad:f0ff:fe71:c163/64 scope link proto kernel_ll
       valid_lft forever preferred_lft forever
```

```console
[root@dnsResolver:~]# iptables -L -v -t nat
[…]
Chain nixos-nat-post (1 references)
 pkts bytes target     prot opt in     out     source               destination
    8   480 MASQUERADE  all  --  any    enp1s0  anywhere             anywhere             mark match 0x1

Chain nixos-nat-pre (1 references)
 pkts bytes target     prot opt in     out     source               destination
   42  2640 MARK       all  --  tap-unikernel any     anywhere             anywhere             MARK set 0x1
   11   846 DNAT       udp  --  enp1s0 any     anywhere             anywhere             udp dpt:domain to:10.0.0.2:53
    0     0 DNAT       tcp  --  enp1s0 any     anywhere             anywhere             tcp dpt:domain to:10.0.0.2:53
   10   600 DNAT       tcp  --  enp1s0 any     anywhere             anywhere             tcp dpt:https to:10.0.0.2:443
   10   600 DNAT       tcp  --  enp1s0 any     anywhere             anywhere             tcp dpt:domain-s to:10.0.0.2:853
```

```console
[root@dnsResolver:~]# sysctl net.ipv4.ip_forward
net.ipv4.ip_forward = 1
```

But none of this seem to work for me:
```nix
        virtualisation.forwardPorts = [
          {
            from = "host";
            host.port = 6000;
            guest.address = "10.0.0.2";
            guest.port = 443;
          }
          {
            from = "host";
            host.port = 6001;
            guest.address = "127.0.0.1";
            guest.port = 443;
          }
          {
            from = "host";
            host.port = 6002;
            guest.address = "10.0.2.15";
            guest.port = 443;
          }
          {
            from = "host";
            host.port = 6003;
            guest.address = "192.168.1.3";
            guest.port = 443;
          }
        ];
```
Passing down to `qemu`:
`-net nic,netdev=user.0,model=virtio -netdev user,id=user.0,hostfwd=tcp::6000-10.0.0.2:443,hostfwd=tcp::6001-127.0.0.1:443,hostfwd=tcp::6002-10.0.2.15:443,hostfwd=tcp::6003-192.168.1.3:443`

But then none of those ports work:
```console
$ for p in {6000..6003}; do curl -kLvs --connect-timeout=1 https://localhost:$p; done
* Host localhost:6000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:6000...
* connect to ::1 port 6000 from ::1 port 39242 failed: Connection refused
*   Trying 127.0.0.1:6000...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* SSL Trust: peer verification disabled
* Connection timed out after 1002 milliseconds
* closing connection #0
* Host localhost:6001 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:6001...
* connect to ::1 port 6001 from ::1 port 44010 failed: Connection refused
*   Trying 127.0.0.1:6001...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* SSL Trust: peer verification disabled
* Connection timed out after 1001 milliseconds
* closing connection #0
* Host localhost:6002 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:6002...
* connect to ::1 port 6002 from ::1 port 48576 failed: Connection refused
*   Trying 127.0.0.1:6002...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* SSL Trust: peer verification disabled
* Send failure: Broken pipe
* TLSv1.3 (OUT), TLS alert, decode error (562):
* TLS connect error: error:0A000126:SSL routines::unexpected eof while reading
* closing connection #0
* Host localhost:6003 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:6003...
* connect to ::1 port 6003 from ::1 port 39080 failed: Connection refused
*   Trying 127.0.0.1:6003...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* SSL Trust: peer verification disabled
* Send failure: Broken pipe
* TLSv1.3 (OUT), TLS alert, decode error (562):
* TLS connect error: error:0A000126:SSL routines::unexpected eof while reading
* closing connection #0
```
Maybe the `vlans`

projects/DNSvizor/services/dnsvizor/tests/dns.nix
```diff
@@ -0,0 +1,609 @@
...
+          for dns_resolver_ip_or_domain in ${dnsResolverIpOrDomainsPython}:
+              dns_resolver_url = f"{protocol}://{dns_resolver_ip_or_domain}:{port}"
+              with subtest(f"DNS query results from {dns_resolver_url} are correct"):
+                  for query, query_type, expected_answer in ${dnsQueryAndExpectedAnswersPython}:
+                      test_dns(dns_resolver_url, query, query_type, expected_answer)
+
+      with subtest("Systemd hardening works, exposure level is low"):
+          output = dns_resolver.succeed("systemd-analyze security dnsvizor.service | grep -v ✓")
+          dns_resolver.log(output)
+          assert " OK " in output, "overall exposure level too high"
```
ju1m:
```suggestion
          print(dns_resolver.succeed("systemd-analyze security dnsvizor.service --threshold=1 --no-pager"))
```
Note that I'm not adding ` | grep -v ✓` since it would not preserve the `errno` set by a failing `--threshold=1`

projects/DNSvizor/services/dnsvizor/tests/dns.nix
```diff
@@ -0,0 +1,609 @@
...
+
+      with subtest("Web interface can be accessed"):
+          web_interface_url = "https://${webInterfaceDomainOrIp}"
+          if ${if dnsResolverCfg.settings.hostname == null then "True" else "False"}:
+              command = f"curl --insecure {web_interface_url}"
+          else:
+              self_signed_cert = "/tmp/self-signed-cert.pem"
+              dns_client.fail(f"curl --write-out %{{certs}} {web_interface_url} >{self_signed_cert}")
+              dns_client.succeed(f'grep "BEGIN CERTIFICATE" {self_signed_cert}')
+              command = f"curl --cacert {self_signed_cert} {web_interface_url}"
```
ju1m: What is this testing? Details of the TLS implementation of `dnsvizor` ? If so, I would personally not bother that much and always use `--insecure` :>

projects/DNSvizor/services/dnsvizor/tests/dns.nix
jian-lin: @ju1m 

> The rest is mainly me taking notes and recording questions I had.
If I may, I would have appreciated a bit more explanation and/or description comments around unusual blocks.
I think it's better to have them inlined than on some easily forgotten Github PR.

Good idea!  Thanks for pointing out unusual blocks.  I'll add comments for them.  (It is a bit hard for authors themselves to find which blocks need some comment.)

To be clear, what are your "unusual blocks"?  I guess the ones with your comment, right?

projects/DNSvizor/services/dnsvizor/module.nix
jian-lin: `SystemCallErrorNumber` does not just appear in systemd manual, it is **recommended**:

> It is recommended to enforce system call allow lists for all long-running system services. Specifically, the following lines are a relatively safe basic choice for the majority of system services:
[Service]
SystemCallFilter=@system-service
SystemCallErrorNumber=EPERM


> I've had a bad experience with it because the software was catching EPERM and just changing its behavior instead of raising an error, not letting me know I was constraining it too much

> I suggest to keep the default (SIGSYS) that is more likely to let us know we're constraining the software too much.

Systemd hardening is very service-specific.  You need to tune hardening options for each service.

I tend to blame the software for the experience your mentioned.  Luckily, it is not the case for DNSvizor.  So there is little harm to set `SystemCallErrorNumber` here.

I am open for this change.  WDYT?

projects/DNSvizor/services/dnsvizor/tests/dns.nix
jian-lin: > I am unable to reach that webpage from the host using virtualisation.forwardPorts in interactive mode

Why do you want to do that in an interactive **test**?  It isn't a supported use case for tests.  It is supported by our demo though.

Since it is not a supported use case and I have spent more time than I expected for the review, I won't figure out the correct config for you.  Instead, I'll just provide a few debug points in case you still want to go down the rabbit hole:

1. make sure `services.dnsvizor.openFirewall` is enabled
2. We assume `services.dnsvizor.mainInterface` is the default interface to use for connecting to the outside world.  This holds true in **non-interactive** tests but not in **interactive** tests.  In interactive tests, `eth0` is the default interface.  So you probably have to set more firewall rules and/or routes.

projects/DNSvizor/services/dnsvizor/tests/dns.nix
ju1m: > I guess we can get both of this by running one command with `--threshold` followed by another command with `grep`ing?
> 
> Maybe that's too much and just one command with `--threshold` is enough?
> 
> WDYT?

Well:
1. the `print` is here to let anyone who cares `grep ✗` in the nix build log, so I think that's covered with a single command.
2. I very rarely do such grepping as the check marks (`✓`) are also important to have when reviewing, to understand what's happening.
3. If you really want to do that just do it in python, that would preserve the `errno` without using the tricky `set -o pipefail`.

> BTW, the equivalent of `grep`ing `OK` is `--threshold 49` instead of `10`. See https://github.com/systemd/systemd/blob/95a7b7d474c6a8c8c03ae708ce405e1d3541897e/src/analyze/analyze-security.c#L1739

Right, `--threshold` enables to expect a narrower interval.

projects/DNSvizor/default.nix
```diff
@@ -0,0 +1,297 @@
...
+      };
+    };
+  };
+
+  nixos.modules.services.dnsvizor = {
+    name = "DNSvizor";
+    module = ./services/dnsvizor/module.nix;
+    examples =
+      let
+        mkExample =
```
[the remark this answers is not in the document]
ju1m: Sure not now, but if we want to write more tests using a `cartesianProduct` of options for other services, then that may come handy to have a helper.

projects/DNSvizor/services/dnsvizor/tests/dns.nix
```diff
@@ -0,0 +1,609 @@
...
+        {
+          ${cfg.ipv4Prefix} = [ cfg.settings.hostname ];
+        }
+        // lib.optionalAttrs cfg.ipv6Enabled {
+          ${cfg.ipv6Prefix} = [ cfg.settings.hostname ];
+        }
+      );
+
+      networking = {
+        inherit useNetworkd;
```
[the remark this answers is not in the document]
ju1m: Right, more tests have many benefits, kudos for coming up with a way to multiply them.

projects/DNSvizor/services/dnsvizor/tests/dns.nix
```diff
@@ -0,0 +1,609 @@
...
+}:
+
+assert builtins.elem resolverKind [
+  "stub"
+  "recursive"
+];
+assert builtins.isBool useNetworkd;
+
+let
+  vlan = 1;
```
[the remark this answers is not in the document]
jian-lin: > Inlined comments endure more, and are more discoverable than Git commit descriptions, so my general rule of thumb is that every thing that gets out of the ordinary benefits from an explanation comment, sometimes also a description comment, as why some piece of code is there and what it does may be obvious to you (or at least to your 2026-01 you), but may not be for others

In general, I agree!

However, it is arguable whether it is ordinary or not in this case.  One may argue that it is ordinary because it is used a lot in nixos tests in Nixpkgs.

I think it is a documentation issue of Nixpkgs in this case and should be fixed there.  Any volunteers?  I myself do not have enough bandwidth recently.

projects/DNSvizor/services/dnsvizor/tests/dns.nix
jian-lin: In production, the cert can be obtained more securely.  For example, it can be done by `ssh`ing into where dnsvizor is deployed.

projects/DNSvizor/services/dnsvizor/examples/recursive-dns-resolver.nix
```diff
@@ -0,0 +1,30 @@
+{ ... }:
+
+{
+  services.dnsvizor = {
+    enable = true;
+    memory = 128;
+    mainInterface = "enp0s3";
+    settings = {
+      hostname = "dnsvizor.mydomain.com";
```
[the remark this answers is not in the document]
ju1m: TIL `.example`. Perfect.

projects/DNSvizor/services/dnsvizor/tests/dns.nix
ju1m: Hopefully the _Release notes_ section of the _Contributor Manual_ will give us those kind of heads-up at some point.

jian-lin:
New changes:
- assert type of useNftables
- rebase against `main` to include #2011
  - this allows us to remove [mkForce in virtualisation.forwardPorts of demo][1]
- bump binary dnsvizor to include fixes for query log web page
- add a demo instruction about query log

[1]: https://github.com/ngi-nix/ngipkgs/pull/1944#discussion_r2703841520

jian-lin:
New changes:
- replace firewall rules with routes

All comments of @ju1m have been addressed.

projects/DNSvizor/services/dnsvizor/tests/dns.nix
jian-lin: OK, the implementation is pushed.

> it's weird that tests become flaky after that

Turns out it is an Nixpkgs bug.  I created https://www.github.com/NixOS/nixpkgs/issues/487520 to track it and worked it around in this PR.

jian-lin:
@imincik @eljamm Is there anything I can do to get this PR merged?

Let me reiterate that this PR
- is ready
- is self-contained and **does not depend on any packaging PR**

jian-lin:
rebased against `main`

projects/DNSvizor/default.nix
```diff
@@ -0,0 +1,302 @@
...
+                    (lib.concatStringsSep "-")
+                  ];
+                mkTest =
+                  testArgs:
+                  lib.nameValuePair ("${exampleName}:${testArgsToString testArgs}") {
+                    module = import testModule (args // testArgs // { inherit exampleName; });
+                    inherit problem;
+                  };
+              in
+              lib.listToAttrs (map mkTest (lib.cartesianProduct testCfg));
```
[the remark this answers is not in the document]
jian-lin: > makes tab-completion impossible

Only for some shells.  For example, fish works but bash does not.



> To something like:
> 
> ```nix
> [
>   "stub-ipv4"
>   "stub-ipv4-nft"
>   "stub-ipv4-networkd"
>   "stub-ipv4-networkd-nft"
>   "recursive-dual"
>   "recursive-dual-nft"
>   "recursive-dual-networkd"
>   "recursive-dual-networkd-nft"
> ]
> ```

Happy to do so.  Just to be clear, do the names have to be exactly the same as your suggested or it is enough if the names are not "too complicated" and do not break tab-completion in shells such as bash?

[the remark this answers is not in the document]
jian-lin: When playing with shell completion, I find that `projects` being system-agnostic is a bit strange.  Commented here: https://github.com/ngi-nix/ngipkgs/pull/1964/changes#r2781713153

projects/DNSvizor/default.nix
```diff
@@ -0,0 +1,302 @@
...
+                false
+              ];
+            };
+          }
+        ]
+      );
+  };
+
+  nixos.demo.vm = {
+    # no demo test since it is the same as the example test
```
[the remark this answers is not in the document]
eljamm: > About the naming, is it better to use `demo`? Among existing projects, some use `demo`, others use custom test names.

The naming is not an issue as we could have multiple demo tests, so I'd say it's more of a convention for now.

eljamm (approved):
Demo works great and the packaging is pretty comprehensive. Awesome job, @jian-lin!

Issue #1870, opened 2025-12-04
Broken `nix flake show` and `nix flake check`


imincik:
With current main version `nix flake show` and `nix flake check` are broken.

```
nix flake show

│   └───x86_64-linux
│       ├───"infra/makemake": derivation 'nixos-system-makemake-26.05.20251130.23258e0'
│       ├───"infra/overview": derivation 'overview'
│       ├───"infra/pre-commit": derivation 'pre-commit-run'
│       ├───"packages/_0wm-ap-mock": derivation '0wm-ap-mock-0-unstable-2025-10-27'
│       ├───"packages/_0wm-client": derivation '0wm-client-0-unstable-2025-10-27'
│       ├───"packages/_0wm-opmode": derivation '0wm-opmode-0-unstable-2025-10-27'
│       ├───"packages/_0wm-server": derivation 'ocaml5.2.1-0wm-server-0-unstable-2025-11-24'
│       ├───"packages/anastasis": derivation 'anastasis-0.6.4'
│       ├───"packages/anastasis-gtk": derivation 'anastasis-gtk-0.6.3'
│       ├───"packages/anastasis-vmTest": derivation 'vm-test-run-anastasis-httpd'
│       ├───"packages/atomic-browser": derivation 'atomic-browser-0.40.0'
│       ├───"packages/atomic-cli": derivation 'atomic-cli'
│       ├───"packages/atomic-server": derivation 'atomic-server-0.40.0'
error: expected a derivation
```

```
nix flake check

error:
       … while checking flake output 'nixosModules'
         at /nix/store/ikl8f5n4mhycfn41r2crddy3rw0wydh6-source/flake.nix:56:9:
           55|         # WARN: this is currently unstable and subject to change in the future
           56|         nixosModules = classic'.nixos-modules;
             |         ^
           57|

       … while checking the NixOS module 'nixosModules.services'

       … while evaluating a branch condition
         at /nix/store/j68m788zvrjj69vd4w3kpqq6qw9gxlfs-source/lib/attrsets.nix:1695:11:
         1694|           in
         1695|           if length values == 1 || pred here (elemAt values 1) (head values) then
             |           ^
         1696|             head values

       … while evaluating the module argument `pkgs' in "/nix/store/j68m788zvrjj69vd4w3kpqq6qw9gxlfs-source/nixos/modules/services/hardware/bluetooth.nix":

       … noting that argument `pkgs` is not externally provided, so querying `_module.args` instead, requiring `config`

[...]

Issue #1839, opened 2025-11-21
US: Improved design of top level ngipkgs attributes


imincik:
### Description

As an NGIpkgs contributor, I want a consistent interface to all top-level deliverable attributes (flake and non-flake) so that it's intuitive to use and access them.

### Acceptance criteria

- Given the new project interface, when I access an attribute (demo, derivation, test, ...), then usage is consistent between flakes and non-flakes.
- Given the new project interface, when I read the documentation, the attribute structure is clear for flakes and non-flakes.
- Given a new deliverable for a project is implemented, when I try to build it from my local source code, it works.

### Implementation details

- Pick the most important things to expose under toplevel project attributes
  - derivations
  - tests
  - demos

Issue #1872, opened 2025-12-05
Forgejo: Eval check fails in nix repl


phanirithvij:
It works via the cli `nix build -L .#checks.x86_64-linux.projects/Forgejo/nixos/check` and the upstream nixos test is working as well.

```console
$ git pull upstream main
From https://github.com/ngi-nix/ngipkgs
 * branch            main       -> FETCH_HEAD
Already up to date.

$ nix repl .
Nix 2.33pre20251107_479b6b73
Type :? for help.
Loading installable 'git+file:///shed/Projects/nixer/community-projects/ngipkgs/ngipkgs#'...
Added 8 variables.
checks, devShells, formatter, lib, nixosConfigurations, nixosModules, overlays, packages
nix-repl> checks.x86_64-linux."projects/Forgejo/nixos/check"
error:
       … while evaluating the attribute 'x86_64-linux."projects/Forgejo/nixos/check"'
         at /nix/store/ikl8f5n4mhycfn41r2crddy3rw0wydh6-source/flake.nix:105:25:
          104|                       checksForNixosTypes = {
          105|                         "projects/${projectName}/nixos/check" = classic.checks.${projectName};
             |                         ^
          106|                       };

       … in the condition of the assert statement
         at /nix/store/j68m788zvrjj69vd4w3kpqq6qw9gxlfs-source/pkgs/build-support/trivial-builders/default.nix:172:5:
          171|     # TODO: To fully deprecate, replace the assertion with `lib.isString` and remove the warning
          172|     assert lib.assertMsg (lib.strings.isConvertibleWithToString text)
             |     ^
          173|       ''pkgs.writeText ${lib.strings.escapeNixString name}: The second argument should be a string, but it's a ${builtins.typeOf text} instead.'';

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: The option `projects.Forgejo.nixos.tests.mysql.config' does not exist. Definition values:
       - In `<unknown-file>'

$ nom build -L .#checks.x86_64-linux.projects/Forgejo/nixos/check --rebuild
Forgejo-eval-check> checking outputs of '/nix/store/bj02f0ln87xpc4x8hgdjqyswp2plbyfx-Forgejo-eval-check.drv'
┏━ Dependency Graph:
┃ ✔ Forgejo-eval-check
┣━━━ Builds
┗━ ∑ ⏵ 0 │ ✔ 1 │ ⏸ 0 │ Finished at 11:43:20 after 0s

[...]

PR #1964, opened 2026-01-13
Make top-level deliverable attributes consistent

eljamm:
Make top-level deliverable attributes consistent between flake and non-flake attributes by exposing a top-level `projects` attribute in both, containing all important things like demos and tests.

```shellSession
$ nix build .#projects.Openfire/
.../demo  .../module-check  .../tests/basic  .../tests/demo

$ nix build .#projects.Openfire/tests/basic -L
$ nix run .#projects.Openfire/demo
```

```shellSession
$ nix-build -A projects.Openfire.tests.demo
$ nix-build -A projects.Openfire.demo
```

Closes https://github.com/ngi-nix/ngipkgs/issues/1839
Fixes https://github.com/ngi-nix/ngipkgs/issues/1870

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -57,6 +57,7 @@ let
       self = flake;
       pkgs = pkgs.extend self.overlays.default;
       options = self.optionsDoc.optionsNix;
+      projects = self.project-utils.projects;
     };
 
     nixos-modules =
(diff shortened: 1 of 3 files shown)


eljamm:
> I guess there is a good reason, but why is the attribute path in the new CLI joined by `/` instead of `.` as with old CLI ?

I tried to have them joined by `.`, but that didn't work with flakes:

```shellSession
$ nix build .#projects.Openfire.demo -L
error: flake 'git+file:///home/user/ngipkgs/.worktree/features/toplevel-design' does not provide attribute 'packages.x86_64-linux.projects.Openfire.demo', 'legacyPackages.x86_64-linux.projects.Openfire.demo' or 'projects.Openfire.demo'
```

In checks, we're joining by `/`, so I did the same here. Also, there actually exists a `flake-util` function that flattens attributes called [`flattenTree`](https://github.com/numtide/flake-utils?tab=readme-ov-file#flattentree--attrs---attrs) in the same way. The reason why we're not using it is that our own flattening function recurses indefinitely, as opposed to respecting `recurseIntoAttrs`, which I think works nicely here.

eljamm:
> When I try to explore `projects` by accessing them in the nix repl via the flakes interface
> 
> either `nix repl .` or `nix repl` and `:lf .`
> 
> I get this #1872
> 
> Can anyone reproduce it?
> 
> Edit: `nix-build -A projects.Forgejo.module-check` also fails now.

I can [reproduce](https://github.com/ngi-nix/ngipkgs/issues/1872#issuecomment-3748557637) it on the latest branch, so this PR is probably not the issue. It's also weird because the check seems to be good in CI. I'll take a look.

PR #1965, opened 2026-01-14
refactor: separate formatter from git hooks

### LINKED ISSUE #1880: US: Investigate using treefmt-nix for formatting like nixpkgs
nixpkgs has had treewide formatting checks enabled for a while (see nixpkgs ci/default.nix)

In ngipkgs we have a pre-commit hook along with a `nix fmt`-able `formatter` defined which runs these hooks.

There is a possibility to use treefmt-nix, git-hooks.nix has treefmt integration as well.

eljamm:
This should also fix inconsistencies with formatter versions. Test with:

```shellSession
$ nix fmt
$ nix-shell --run treefmt
```

**Note:** before this change, `nix fmt` wouldn't fail, even though some files weren't correctly formatted.

Closes: https://github.com/ngi-nix/ngipkgs/issues/1880
Closes: https://github.com/ngi-nix/ngipkgs/pull/864

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -47,6 +47,8 @@ let
       fixups = self.call ./pkgs/overlays.nix { };
     };
 
+    formatter = self.call ./maintainers/formatter.nix { };
+
     optionsDoc = pkgs.nixosOptionsDoc {
       inherit (self.project-utils.eval-projects) options;
     };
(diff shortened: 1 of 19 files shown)


imincik (approved):
Looks good to me. Thank you.

PR #1981, opened 2026-01-17
fix: wrong optionsDoc for modules

eljamm:
This also fixes the missing options regression in the overview.

image


phanirithvij:
I think it's best we setup CI to catch if this regression were to happen.

PR #1987, opened 2026-01-19
refactor: compose project modules as paths 

eljamm:
Currently, modules can be a `deferredModule`, which not only makes composing them more difficult, but also prevents us from showing them in the documentation.

This commit changes those types to `path` and fixes the issues listed above. This will also help us migrate away from `hydrated-projects` in the future.

Note that we still want to keep the tests' module type as `deferredModule`, otherwise the overview build will slow down considerably (see note in `projects/types.nix`).

Related: https://github.com/ngi-nix/ngipkgs/issues/1990

### DIFF
--- a/pkgs/by-name/anastasis/package.nix
+++ b/pkgs/by-name/anastasis/package.nix
@@ -19,7 +19,6 @@
   jq,
   texinfo,
   which,
-  callPackage,
   gitUpdater,
 }:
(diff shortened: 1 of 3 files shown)


imincik (dismissed):
Looks to me as very nice refactoring. Personally, I don't see any issue and tests are passing. Looks like a good work Fedi.

PR #1992, opened 2026-01-19
refactor(overview): pass args to (sub)modules; expose types as an option

eljamm:
With these changes:
- arguments are passed to modules and submodules, making it easier to change them without making treewide changes
- project types are exposed as an option inside the module system, which allows us to easily re-use them with the proper arguments

### DIFF
--- a/overview/content-types/nix-config.nix
+++ b/overview/content-types/nix-config.nix
@@ -1,25 +1,33 @@
 {
   lib,
   config,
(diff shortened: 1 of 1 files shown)


imincik (dismissed):
Looks good. I left one small comment.

imincik (approved):
Great. Looks good. Thanks.

PR #1999, opened 2026-01-20
refactor: coerce leaf test modules into derivations 

eljamm:
which is much simpler than collecting all the modules and converting them later on. The result is also easier to work with since it's either null or a derivation, as opposed to it potentially being a path or a module.

For now, tests are still passed to [`projects/tests.nix`](https://github.com/ngi-nix/ngipkgs/blob/912fdeb04b02fb9c4560ce5d32d597e9992195c3/projects/tests.nix), which is fine since they should already be derivations. We will clean this up when we migrate away from `hydrated-projects`.

### DIFF
--- a/maintainers/types/default.nix
+++ b/maintainers/types/default.nix
@@ -5,6 +5,8 @@
 */
 {
   lib,
+  pkgs,
+  sources,
   ...
 }@args:
 let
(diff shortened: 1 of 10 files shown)


eljamm:
I think this works nicely, but one clear disadvantage is that this increases evaluation times for flake checks. Interestingly, the overview doesn't seem to be affected, which I think is because we're not accessing any attribute that would make it evaluate.

PR #2008, opened 2026-01-22
fix: docs/project.md generation via nixdoc-to-github

phanirithvij:
After the recent restructuring of projects/types.nix, the nixdoc-to-github uitlity cannot create the project.md anymore

this fixes that in a way where `nix-shell --run nixdoc-to-github` still works, but it has to be run that exact way.

I've tried a naive `cat *.nix` into a single types.nix and tried to use nixdoc-to-github on that but did not produce a valid doc. So I had to do this ugly hack instead.

### DIFF
--- a/maintainers/shells/nixdoc-to-github.nix
+++ b/maintainers/shells/nixdoc-to-github.nix
@@ -1,0 +1,76 @@
+# nix-shell --run nixdoc-to-github
+{
+  lib,
+  gnused,
+  gitMinimal,
+  runCommand,
+  writeShellScriptBin,
+
+  nixdoc-to-github,
+}:
+let
+  mkDocPart =
+    file:
+    nixdoc-to-github.lib.nixdoc-to-github.run {
+      description = "\\\`lib.${file.name}\\\`";
+      category = "";
+      file = file.path; # copied to store
+      output = "\${out:-}";
+    };
+
+  docPart =
+    file:
+    runCommand "docpart-${file.name}"
+      {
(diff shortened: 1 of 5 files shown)


maintainers/shells/nixdoc-to-github.nix
```diff
@@ -0,0 +1,69 @@
...
+        "project.nix"
+        "metadata.nix"
+        "subgrant.nix"
+        "link.nix"
+        "binary.nix"
+        "module.nix"
+        "example.nix"
+        "demo.nix"
+        "test.nix"
+      ];
```
phanirithvij: I could iterate the files in bash instead and then it can work inside the nix-shell

maintainers/shells/nixdoc-to-github.nix
```diff
@@ -3,66 +3,58 @@
...
+    description = "PLACEHOLDER_DESC";
+    category = "";
+    file = "PLACEHOLDER_FILE";
+    output = "PLACEHOLDER_OUTFILE";
+  };
+
+  template = lib.getExe genericRunner;
+
+  cmd = writeShellScriptBin "nixdoc-to-github" ''
+    export PATH="${lib.makeBinPath (if stdenv.isDarwin then [ gnused ] else [ busybox ])}:$PATH"
```
eljamm: `gnused` is available in linux and darwin, right? why not just use that?

eljamm: But this aside, darwin isn't supposed by us, so it's not a priority.

PR #2061, opened 2026-02-09
project/DNSvizor: relax hardening option RestrictAddressFamilies

jian-lin:
This is needed by solo5-0.10.0

fixes https://github.com/ngi-nix/ngipkgs/pull/2056#issuecomment-3872704439


eljamm (approved):
Thanks! The missing option error was a bit misleading for me, how did you figure out the issue?

PR #2174, opened 2026-02-13
mirage: fix updateScript

jian-lin:
MirageOS unikernels are a package set of many targets, which share the
same `src`.  To avoid race condition, instead of attaching the same
updateScript to each target and running the same updateScript multiple
times, we create a dummy package called `update` in the package set
and only attach the updateScript to it.  This way, the updateScript is
only run once.

The result of the updateScript is two independent update commits:
1. pname: old-version -> new-version
2. pname: update deps

The 1st commit only exists if `src` is updated.  The 2nd commit only
exists if ocaml dependencies are updated.  The existence of one does
not affect the existence of the other.

Some detailed changes:
- fix meta.position of `configure`d unikernel
- support updating src and related flake inputs
- update src, related flake inputs and deps of all targets in lockstep
- add a no-op updateScript to each target to prevent
  .github/workflows/update-packages.yaml from running a default
  updateScript

In addition, we provide a script for each target to generate/update
its own deps:

```shellSession
$ nix run -f . dnsvizor.hvt.depsUpdateScriptForThisTarget
```

This PR needs:
- #2154
- https://github.com/NixOS/nixpkgs/pull/490004
  - https://github.com/ngi-nix/ngipkgs/pull/2227

### DIFF
--- a/pkgs/by-name/dnsvizor/mirage.nix
+++ b/pkgs/by-name/dnsvizor/mirage.nix
@@ -6,6 +6,9 @@
   opam-nix,
   stdenv,
   writeShellApplication,
+  unstableGitUpdater,
+  git,
+  emptyDirectory,
 }:
 
 rec {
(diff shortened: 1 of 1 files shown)


wamirez:
Related to https://github.com/ngi-nix/projects/issues/164 and https://github.com/ngi-nix/projects/issues/148.

PR #2251, opened 2026-03-31
Wip update flake lock action

eljamm:


phanirithvij:
ok that's unfortunate. I did the eval on #2252  using nix-eval-jobs with 12 workers and 2048 max memory on a checkout of this pr.

PR #2252, opened 2026-04-01
infra(makemake): reduce nix-eval-jobs workers count

phanirithvij:
It increases the evaluation time on makemake by 2.5mins

Before:
real    2m48.862s
After:
real    5m25.314s

This is not ideal and hopefully a temporary workaround until we think of a better solution, but required for unblocking #2251 #2227

### DIFF
--- a/infra/makemake/buildbot.nix
+++ b/infra/makemake/buildbot.nix
@@ -38,6 +38,9 @@ in
         auth.authToken.file = config.sops.secrets."cachix".path;
       };
       showTrace = true;
+
+      # reduce nix-eval-jobs workers count to avoid OOM
+      evalWorkerCount = 12;
     };
 
     buildbot-nix.worker = {


eljamm (approved):

eljamm:
This [doesn't solve the issue](https://buildbot.ngi.nixos.org/#/builders/5/builds/4662/steps/2/logs/stdio), unfortunately. Although from [the metrics](https://grafana.nixos.org/d/rYdddlPWk/node-exporter-full?orgId=1&from=now-5m&to=now&timezone=browser&var-datasource=default&var-job=node&var-node=makemake.ngi.nixos.org:9100&var-diskdevices=%5Ba-z%5D%2B%7Cnvme%5B0-9%5D%2Bn%5B0-9%5D%2B%7Cmmcblk%5B0-9%5D%2B&refresh=1m) it doesn't seem that we're OOM, the eval still fails when it reaches bonfire.

PR #2285, opened 2026-05-03
Mobilizon: add metadata links

alikhere:
## Summary

`projects/Mobilizon/default.nix` was missing `metadata.links` — the
homepage and source URLs shown on the overview site. Added both
following the same pattern used in `projects/Funkwhale/default.nix`.

## How to test

- `nix fmt -- --fail-on-change` — should pass
- `nix build .#checks.x86_64-linux.projects/Mobilizon/nixos/tests/basic`
  — existing test should still pass

Part of https://github.com/ngi-nix/projects/issues/161

Closes #2283 

## Notes

The upstream module reference, example, and test were already present
from a previous commit. This PR only adds the missing metadata links.

### DIFF
--- a/projects/Mobilizon/default.nix
+++ b/projects/Mobilizon/default.nix
@@ -17,6 +17,20 @@
         "Mobilizon"
       ];
     };
+    links = {
+      homepage = {
+        text = "Homepage";
+        url = "https://mobilizon.org/";
+      };
+      docs = {
+        text = "Documentation";
+        url = "https://docs.mobilizon.org/";
+      };
+      source = {
+        text = "Source Code";
+        url = "https://framagit.org/framasoft/mobilizon";
+      };
+    };
   };
 
   nixos.modules.services = {


projects/Mobilizon/default.nix
```diff
@@ -17,6 +17,16 @@
         "Mobilizon"
       ];
     };
+    links = {
+      homepage = {
+        text = "Homepage";
+        url = "https://joinmobilizon.org/";
```
phanirithvij: Homepage redirects please use the final redirect as the url here.

projects/Mobilizon/default.nix
```diff
@@ -17,6 +17,16 @@
...
+      };
+      source = {
+        text = "Source Code";
+        url = "https://framagit.org/framasoft/mobilizon";
+      };
+    };
```
phanirithvij, line 33: It has a docs link, so add it too

projects/Mobilizon/default.nix
phanirithvij:
```suggestion
        url = "https://mobilizon.org/";
```

projects/Mobilizon/default.nix
```diff
@@ -17,6 +17,20 @@
...
       ];
     };
+    links = {
+      homepage = {
+        text = "Homepage";
+        url = "https://joinmobilizon.org/";
+      };
+      docs = {
+        text = "Documentation";
+        url = "https://docs.joinmobilizon.org/";
```
phanirithvij:
```suggestion
        url = "https://docs.mobilizon.org/";
```

PR #2301, opened 2026-06-30
ci(makemake): use locked version of nixos-rebuild

jfly:
While investigating <https://github.com/ngi-nix/infra/issues/5>, I realized we're using an unlocked reference to `nixpkgs`, which gives us some non-repeatable behavior in CI.

`--inputs-from` is a nice trick here: we now benefit from whatever locked version of nixpkgs we have in our `flake.lock`. I checked our other github workflows and didn't find any other `nix run` incantations that would benefit from this.

A side effect of this is that it works around
<https://github.com/ngi-nix/infra/issues/5>, as we'll revert to a version of nixpkgs that predates
<https://github.com/NixOS/nixpkgs/commit/f23bab46664a3a511fa10d3dcdfb8a9322b14502>.

### DIFF
--- a/.github/workflows/makemake.yaml
+++ b/.github/workflows/makemake.yaml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       SSH_KEY: ${{ secrets.SSH_KEY }}
+      FLAKE_REF_ROOT: github:${{ github.repository }}/${{ github.sha }}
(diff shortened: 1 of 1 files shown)


phanirithvij (approved):
Nice, I wonder though if this pr is merged and makemake gets deployed from the latest committed declarative config, could you lose root access @jfly? Or did @imincik add your key manually?

jfly:
@imincik added my key to `~/.ssh/authorized_keys`. I deployed a few times while working on this, and didn't lose access.

